Issue #57, opened 2024-12-10
v0.5 Preparation


qor-lb:
- [x] Define the term _incubation repository_ in the SCORE context
- [x] Create incubation repositories for the agreed components
- [x] Nominate team participants
- [x] #24 
- [x] #25 
- [x] Clarification of Architecture Community F2F meeting cadence

PR #63, opened 2024-12-11
docs: move features to top level

ramceb:
As feature will also contain tests docs is the
wrong location

Issue-ref: see  #57


ltekieli:
Was discussed with @hoe-jo that it would be better to have everything under ```/docs``` directory for better support of tooling.

PR #76, opened 2024-12-13
Improve line length settings

AlexanderLanin:
Currently "mandatory wrapping" is configured in the project settings, which is not what everyone wants.

Example: It took me a surprisingly long amount of time to understand a line of code, because it looked like two lines. I've not used wrapping in the last 10 years... or ever.

At the same time let's add a python ruler for 80 characters, since it's the de-facto-standard. Not only de-facto, it is the standard. See PEP-8.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,12 +4,19 @@
     "files.trimTrailingWhitespace": true,
     "editor.insertSpaces": true,
     "editor.tabCompletion": "on",
-    "editor.wordWrap": "wordWrapColumn",
-    "editor.wrappingIndent": "same",
-    "editor.wordWrapColumn": 100,
+
+    // Default for any filetype
     "editor.rulers": [
         100
     ],
+
+    // Python Settings
+    "[python]": {
+        // In python using 80 characters per line is the standard.
+        "editor.rulers": [
+            79
+        ],
+    },
 
     // RST Settings
     "[restructuredtext]": {


MaximilianSoerenPollak:
This makes sense I would say.
Further discussions regarding other rules that should be added to either the editor or ruff setup can be held accordingly when needed.

AlexanderLanin:
Could someone merge this? I don't have write access to score.

ltekieli (approved):

ltekieli:
@AlexanderLanin please just fix gitlint and we are ready to go.

AlexanderLanin:
Rebased + changed commit message

FScholPer:
gitlint check failed @ltekieli how do we handle these type of issues. message has to be changed?

ltekieli:
> gitlint check failed @ltekieli how do we handle these type of issues. message has to be changed?

Yes, the git guideline has all the necessary information for properly formatting a commit message:
- https://eclipse-score.github.io/score/process/guidelines/git/index.html

This is enforced by gitlint which is run in a github action.

MaximilianSoerenPollak:
> gitlint check failed @ltekieli how do we handle these type of issues. message has to be changed?

> IDE: ... 

the title pre-fix so IDE in this case has to be lowercase. uppercase is not allowed. Allowed regex: `(^[a-z_-]+: .+$)`

>Signed-off-by: Alexander Lanin <Alexander.Lanin@etas.com>
> Issue-Ref: see https://github.com/eclipse-score/score/pull/76

Have to change the `see` after to one of the accepted words:  `close / closes / closed` | `fix / fixes / fixed` | `resolve / resolves / resolved` or `none`, as well as changing the  `Ref` to lowercase. 

Regex: `Issue-ref: (see|close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved) #(\d+))`

The rest seems to be okay as far as I can tell.

AlexanderLanin:
My fault, I'll fix it. We would benefit from better error messages, but I'm not sure this is possible with git-lint.

AlexanderLanin:
Rebased + fixed commit message (I'm referencing this PR as there is no issue and #119 is not decided)

Issue #39, opened 2024-12-06
Initial version of Software Dev Process


antonkri:
Please introduce parts of the Software Development Process, that are already defined.

aschemmel-tech:
Plan to release the processes based on the precondition that we can use the standards for linking:
- Project Management Plan includes folder structure - see https://github.com/eclipse-score/score/pull/22 
- CODEOWNER file und groups
- Change Request / Contribution Request
- Iso26262 and ASPICE incl. tailoring
- Workproducts, Workflows, Roles

FScholPer:
@antonkri  can we close that "very old" ticket?

aschemmel-tech:
Processes are all in ML1 - sufficient for "initial"

PR #77, opened 2024-12-13
Improvement: Update CODEOWNERS

aschemmel-tech:
Issue-ref: see #39

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,46 @@
 # Comment out as not in score yet
-# *                                   @eclipse-score/automotive-score-committers
-# .github                             @eclipse-score/automotive-score-committers
+# *                                  @eclipse-score/infrastructure-tooling-community
+# .*                                 @eclipse-score/infrastructure-tooling-community
+# .github/CODEOWNERS                 @eclipse-score/technical-lead
 
-# docs                                @eclipse-score/automotive-score-committers
+# /docs                              @eclipse-score/process-community
+# /docs/concepts                     @eclipse-score/automotive-score-committers
+# /docs/glossary                     @eclipse-score/automotive-score-committers
+# /docs/features                     @eclipse-score/technical-lead
+# /docs/features/*/architecture      @eclipse-score/automotive-score-committers
+# /docs/features/*/requirements      @eclipse-score/technical-lead
+# /docs/features/*/safety_analysis   @eclipse-score/safety-manager
+# /docs/features/*/safety_planning   @eclipse-score/safety-manager @eclipse-score/technical-lead
+# /docs/features/*/verification      @eclipse-score/quality-manager @eclipse-score/safety-manager
+# /docs/manual                       @eclipse-score/automotive-score-committers
+# /docs/overview                     @eclipse-score/project-lead
+# /docs/pmp                          @eclipse-score/technical-lead @eclipse-score/process-community
+# /docs/pmp/quality_management       @eclipse-score/quality-manager
+# /docs/pmp/safety_management        @eclipse-score/safety-manager
+# /docs/pmp/security_management      @eclipse-score/security-manager
+# /docs/requirements/stakeholder     @eclipse-score/technical-lead
+# /docs/requirements/tool            @eclipse-score/infrastructure-tooling-community @eclipse-score/process-community
+# /docs/release                      @eclipse-score/quality-manager @eclipse-score/technical-lead
+# /docs/safety                       @eclipse-score/safety-manager @eclipse-score/technical-lead
+# /docs/security                     @eclipse-score/security-manager @eclipse-score/technical-lead
+# /docs/tutorials                    @eclipse-score/automotive-score-committers
+
+# /examples                          @eclipse-score/automotive-score-committers
+# /integration_tests                 @eclipse-score/automotive-score-committers
+# /tools                             @eclipse-score/infrastructure-tooling-community
+
+# in separate process_description repository:
+#
+# /process                  @eclipse-score/process-community
+
+# in separate <module_name> repositories:
+#
+# /docs                     @eclipse-score/process-community
+# /docs/manual              @eclipse-score/<module_name>/safety-manager
+# /docs/release             @eclipse-score/<module_name>/quality-manager @eclipse-score/<module_name>/module-lead
+# /docs/safety_plan         @eclipse-score/<module_name>/safety-manager @eclipse-score/<module_name>/module-lead
+# /docs/safety_analysis     @eclipse-score/<module_name>/safety-manager
+# /docs/verification        @eclipse-score/<module_name>/quality-manager @eclipse-score/<module_name>/safety-manager
+# /components               @eclipse-score/<module_name>/technical-lead
+# /components/*/            @eclipse-score/<module_name>/automotive-score-committers
 


.github/CODEOWNERS
```diff
@@ -1,6 +1,51 @@
 # Comment out as not in score yet
-# *                                   @eclipse-score/automotive-score-committers
-# .github                             @eclipse-score/automotive-score-committers
+# *                                              @eclipse-score/infrastructure-tooling-community
```
LittleHuba, line 2: I don't think the infrastructure-tooling-community should have final say about codeownership.

Imagine somebody introducing a new directory tree. A change on that level should require an approval from somebody high in the project and not only one of the communities.

Proposal: `@eclipse-score/technical-lead` or `@eclipse-score/project-lead`

.github/CODEOWNERS
```diff
@@ -1,6 +1,51 @@
...
+# .github                                        @eclipse-score/infrastructure-tooling-community
 
-# docs                                @eclipse-score/automotive-score-committers
+# docs/concepts                                  @eclipse-score/automotive-score-committers
+# docs/glossary                                  @eclipse-score/automotive-score-committers
+# docs/features/<feature-name>/architecture      @eclipse-score/automotive-score-committers
```
LittleHuba, line 7: Missing a rule for `docs/features`.
This would be the relevant rule when somebody wants to introduce a new feature.

.github/CODEOWNERS
```diff
@@ -1,6 +1,51 @@
...
+# docs/features/<feature-name>/requirements      @eclipse-score/automotive-score-committers @eclipse-score/technical-lead
+# docs/features/<feature-name>/safety_analysis   @eclipse-score/automotive-score-committers @eclipse-score/safety-manager
+# docs/features/<feature-name>/safety_planning   @eclipse-score/safety-manager @eclipse-score/technical-lead
+# docs/features/<feature-name>/verification      @eclipse-score/quality-manager @eclipse-score/safety-manager
+# docs/manual                                    @eclipse-score/automotive-score-committers
+# docs/overview                                  @eclipse-score/project-lead
+# docs/pmp/change_management                     @eclipse-score/technical-lead @eclipse-score/process-community
```
LittleHuba, line 14: Don't we want the full `docs/pmp` directory under `@eclipse-score/technical-lead @eclipse-score/process-community`
I don't really see anything that would go under that directory owned by anybody else. If it ever comes to that, we can still split it up.

.github/CODEOWNERS
```diff
@@ -1,6 +1,51 @@
...
+# in a separate repository:
+# process                                        @eclipse-score/process-community
+
+# in separate repositories:
+# modules/<module_name>/docs/manual              @eclipse-score/automotive-score-committers @eclipse-score/safety-manager
+# modules/<module_name>/docs/release             @eclipse-score/quality-manager @eclipse-score/module-lead
+# modules/<module_name>/docs/safety_plan         @eclipse-score/safety-manager @eclipse-score/module-lead
+# modules/<module_name>/docs/safety_analysis     @eclipse-score/automotive-score-committers @eclipse-score/safety-manager
+# modules/<module_name>/docs/verification        @eclipse-score/quality-manager @eclipse-score/safety-manager
+# modules/<module_name>/components               @eclipse-score/automotive-score-committers
```
LittleHuba, line 50: Here we probably want more fine-granularity in the long run. More or less a team of experts for a specific module.
But for now this should be sufficient.

.github/CODEOWNERS
```diff
@@ -1,6 +1,51 @@
...
-# .github                             @eclipse-score/automotive-score-committers
+# *                                              @eclipse-score/infrastructure-tooling-community
+# .github                                        @eclipse-score/infrastructure-tooling-community
 
-# docs                                @eclipse-score/automotive-score-committers
+# docs/concepts                                  @eclipse-score/automotive-score-committers
+# docs/glossary                                  @eclipse-score/automotive-score-committers
+# docs/features/<feature-name>/architecture      @eclipse-score/automotive-score-committers
+# docs/features/<feature-name>/requirements      @eclipse-score/automotive-score-committers @eclipse-score/technical-lead
+# docs/features/<feature-name>/safety_analysis   @eclipse-score/automotive-score-committers @eclipse-score/safety-manager
```
LittleHuba, line 9: Are you aware that putting more than one owner is an `OR` operation? To me this looks like you thought of an `AND` operation here.

LittleHuba (commented):
Some comments from my side

.github/CODEOWNERS
AlexanderLanin: Yeah we can have one rule for `docs/pmp` and specific rules for e.g. security_management which will involve the security-manager.

.github/CODEOWNERS
AlexanderLanin: Potentially the same problem for all top level directories, like /docs.

.github/CODEOWNERS
```diff
@@ -1,6 +1,51 @@
 # Comment out as not in score yet
-# *                                   @eclipse-score/automotive-score-committers
-# .github                             @eclipse-score/automotive-score-committers
+# *                                              @eclipse-score/infrastructure-tooling-community
+# .github                                        @eclipse-score/infrastructure-tooling-community
 
-# docs                                @eclipse-score/automotive-score-committers
+# docs/concepts                                  @eclipse-score/automotive-score-committers
```
AlexanderLanin, line 5: Generic: directories should start with `/` since we want to address `/docs/concepts` *only* on the root level with this rule.

```suggestion
# /docs/concepts                                  @eclipse-score/automotive-score-committers
```

.github/CODEOWNERS
AlexanderLanin: Those teams are module=repository specific anyway?!

AlexanderLanin (changes requested):

.github/CODEOWNERS
aschemmel-tech: I would agree to Alexander's reasoning, I think the tooling team would ask PL or TL if something of introduce new directory tree dimension would happen.

.github/CODEOWNERS
```diff
@@ -1,6 +1,51 @@
 # Comment out as not in score yet
-# *                                   @eclipse-score/automotive-score-committers
-# .github                             @eclipse-score/automotive-score-committers
+# *                                              @eclipse-score/infrastructure-tooling-community
+# .github                                        @eclipse-score/infrastructure-tooling-community
```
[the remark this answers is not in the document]
aschemmel-tech: accept

.github/CODEOWNERS
aschemmel-tech: ok, will change

.github/CODEOWNERS
aschemmel-tech: is this true? in other github projects I work in, this will lead to an "AND" condition on the reviews

.github/CODEOWNERS
```diff
@@ -1,6 +1,51 @@
...
+# docs/security                                  @eclipse-score/security-manager @eclipse-score/technical-lead 
+# docs/tutorials                                 @eclipse-score/automotive-score-committers
+
+# examples                                       @eclipse-score/automotive-score-committers
+# integration_tests/feature                      @eclipse-score/automotive-score-committers
+# integration_tests/platform                     @eclipse-score/automotive-score-committers
+# tools                                          @eclipse-score/infrastructure-tooling-community
+
+# in a separate repository:
+# process                                        @eclipse-score/process-community
```
masc2023, line 42: Should there not at least another gateing group, e.g. Quality Manager or also Safety and Security Manager?

.github/CODEOWNERS
AlexanderLanin: We had the same discussion some weeks ago in some other context. I also believe it's OR. But people said that they believe it's AND. Documentation says it's OR.

Maybe we need to just try it.

.github/CODEOWNERS
aschemmel-tech: ok, will solve

.github/CODEOWNERS
aschemmel-tech: ok, will change

.github/CODEOWNERS
aschemmel-tech: ok, will split up teams

.github/CODEOWNERS
masc2023: Would not agree, you mention it also above, so is that not a bottleneck then too? And if the person at the being are the same, the they can be part of e.g. safety managers and community too.

AlexanderLanin (approved):
Looks good enough for me. It is worth a try to start like that.

Before fully activating it (uncommenting):
* describe workarounds in case we misconfigured something here. Who shall be able to merge regardless of CODEOWNERS? That's typically some admin role.
* those teams are currently not meaningfully filled with people. Don't activate any ownership unless the team is actually set up.

.github/CODEOWNERS
```diff
@@ -1,6 +1,46 @@
...
+# in separate <module_name> repositories:
+#
+# /docs                     @eclipse-score/process-community
+# /docs/manual              @eclipse-score/<module_name>/safety-manager
+# /docs/release             @eclipse-score/<module_name>/quality-manager @eclipse-score/<module_name>/module-lead
+# /docs/safety_plan         @eclipse-score/<module_name>/safety-manager @eclipse-score/<module_name>/module-lead
+# /docs/safety_analysis     @eclipse-score/<module_name>/safety-manager
+# /docs/verification        @eclipse-score/<module_name>/quality-manager @eclipse-score/<module_name>/safety-manager
+# /components               @eclipse-score/<module_name>/technical-lead
+# /components/*/            @eclipse-score/<module_name>/automotive-score-committers
```
LittleHuba, line 45: @nradakovic @ltekieli when you set up the template repository for modules, could you already add this part there in the CODEOWNERS?

We would probably need a readme there that then tells what steps need to be done manually (e.g. adapting the group names,...)

LittleHuba (approved):

PR #84, opened 2024-12-13
Add contribution request guideline

LittleHuba:
Adds the contribution request guideline which describes how contributions to the SCORE project are to be done.

Currently missing is a rewording of the CONTRIBUTION.md. This shall be done in a follow-up PR once the rewording is aligned.


AlexanderLanin:
LGTM

The images could be simplified in the future, they show a GitHub actor which is not in the text (unnecessary detail).

antonkri (approved):

PR #85, opened 2024-12-13
IPC Contribution Request Initial Version

HartmannNico:
After setting up the IPC CR in it's intial version merge CR into main.

### DIFF
--- a/docs/features/communication/ipc/index.rst
+++ b/docs/features/communication/ipc/index.rst
@@ -19,10 +19,11 @@ Inter-process Communication (IPC) Framework handles the safe, secure and perform
 
 The IPC framework is independent of the application context it is used in.
 
-There are two fundamental information exchange concepts defined
+There are three fundamental information exchange concepts defined
 
-* Topics: A topic is a data entity identifyed by a name, being structured by a *data type* and containing zero to multiple instances of that type (a *value*).
-* Remote Procedures: A remote procedure is an invocable entity identifyed by a name, being structured by parameters of a *data type* and being activated by an invokation passing arguments, i.e. a value for each parameter.
+* **Topics**: A topic is a data entity identifyed by a name, being structured by a *data type* and containing zero to multiple instances of that type (a *value*).
+* **Remote Procedures**: A remote procedure is an invocable entity identifyed by a name, being structured by parameters of a *data type* and being activated by an invokation passing arguments, i.e. a value for each parameter.
+* **Events**: An event signals the change of the state of something. An event carries no data, only the information "the related state change occurred".
 
 Data Types
 ----------
@@ -172,6 +173,14 @@ Remote Procedures
 * Discovery
 * Invocation
 * Sync/Async
+
+Events
+------
+
+* Name
+* Publisher
+* Subscriber
+* Chains / Buffering
 
 
 Zero Copy
--- a/docs/features/communication/ipc/index.rst
+++ b/docs/features/communication/ipc/index.rst
@@ -14,3 +14,5 @@
 
 Inter-process Communication
 ###########################
+
+
--- a/docs/features/communication/ipc/index.rst
+++ b/docs/features/communication/ipc/index.rst
@@ -13,7 +13,7 @@
    # *******************************************************************************
    #
    # Authors
-   # @hartmannnico, 
+   # @hartmannnico,
 
 
 Communication Framework
@@ -21,7 +21,8 @@ Communication Framework
 
 The Communication Framework handles the safe, secure and performant exchange of information between software and/or hardware components of the Score stack.
 
-Communiction covers information exchange between endpoints residing in
+Communication covers information exchange between endpoints residing in
+
 * the same or different process, running on
 * the same or different operating systems, deployed on
 * the same or different compute devices, built into
@@ -75,13 +76,13 @@ The entirety of connected nodes within a mesh we call *fabric*.
 
 **Design Note**
 
-From a perspective of safety, a node also encapsulates a single safety domain. Links provide the means for separating safety domains and thus allow for mixed criticality applications. 
+From a perspective of safety, a node also encapsulates a single safety domain. Links provide the means for separating safety domains and thus allow for mixed criticality applications.
 
 
 Names
 -----
 
-A name is an UTF-8 tag of an element. The underlying Unicode standard is release 16.0.0 https://www.unicode.org/versions/Unicode16.0.0/. 
+A name is an UTF-8 tag of an element. The underlying Unicode standard is release 16.0.0 https://www.unicode.org/versions/Unicode16.0.0/.
 
 All codepoints are valid in a name with the exception of the following codepoints:
 
@@ -91,7 +92,7 @@ All codepoints are valid in a name with the exception of the following codepoint
 * the *wildcard* marker, codepoint ``ASTERISK``: ``*`` ``&#002A``
 * the *any* marker, codepoint ``QUESTION MARK``: ``?`` ``&#003F``
 
-Forbidden as first character in a name is 
+Forbidden as first character in a name is
 
 * the *reference* marker codepoint ``AMPERSAND``: ``&`` ``&#0026``. It's use is discouraged in names in general.
 
@@ -101,8 +102,8 @@ Element names prefixed with an underscore ``LOW LINE``: ``_`` ``&#005F`` are reg
 
 **Design Note**
 
-A name is not a property of an element itself. 
-Instead, a name acts as an *alias* to obtain an element *reference*. 
+A name is not a property of an element itself.
+Instead, a name acts as an *alias* to obtain an element *reference*.
 See _`References`.
 
 
@@ -113,7 +114,7 @@ A namespace is a named scope in which the definition of the elements topic, remo
 
 General rules of names apply to namespace names.
 
-Namespaces can be nested. 
+Namespaces can be nested.
 The path separator between the names is the unicode codepoint ``SOLIDUS``: ``/`` ``&#002F``.
 
 ::
@@ -121,7 +122,7 @@ The path separator between the names is the unicode codepoint ``SOLIDUS``: ``/``
    This/is/a/nested/namespace
 
 
-The namespace name ``/`` is reserved and refers to the global namespace. 
+The namespace name ``/`` is reserved and refers to the global namespace.
 *Global* here means visible with respect to a certain realm that is not further defined. A realm can be a vehicle with it's attached cloud environment or just an application context. It is up to deployment to define the scope of the global namespace.
 
 The namespace name ``super`` is reserved and refers to the parent namespace of the namespace where ``super`` is used. The use of ``super`` in the global namespace is an error.
@@ -206,9 +207,9 @@ Within a namespace elements from another namespace are visible without an explic
 Handles
 -------
 
-A handle is a numeric value that uniquely refers to an individual element in the communication system. 
+A handle is a numeric value that uniquely refers to an individual element in the communication system.
 
-A specific element in the communication system 
+A specific element in the communication system
 
 Data Types
 ----------
@@ -236,7 +237,7 @@ Float64   Floating   ``f64``   ``double``, ``float64_t``  8      An IEEE 756 64 
 BFloat16  Floating   ``bf16``  ``bfloat16_t``             2      A Google brain float 16 floating point
 Char      String     ``char``  ``char32_t``               4      A unicode codepoint (32 bit)
 String    String     ``str``   -                          n/a    A UTF-8 encoded text
-Handle    Reference  -         -                          8      A 64 bit unsigned integer handle  
+Handle    Reference  -         -                          8      A 64 bit unsigned integer handle
 ========= ========== ========= ========================== ====== ============================
 
 The type ``Byte`` may be used as alias for ``UInt8``.
@@ -311,57 +312,51 @@ A HashMap is an unordered collection of data element of the same type with varia
 Pointers
 ````````
 
-There are no data types defined for pointers to other data.
+There are no data types defined for pointers to data types.
 
 References
 ``````````
 
-| Generic references are not allowed as data types. Instead, there are distinct reference types defined:
+The communication framework allows for three potential classes of references:
 
-* Topic References
-* Remote Procedure References
-* Event References
+* References to data types
+   There are no data types defined for references to data types.
 
-The reference marker to an element shall be `AMPERSAND`: `&` `&#0026`.
+* References to information elements
 
-Values of reference
+   * Topic References, data type ``TopicRef``
+   * Remote Procedure References, data type ``FnRef``
+   * Event References, data type ``EventRef``
 
+* References to infrastructure elements
+   Currently we do not define references to infrastructure elements.
+   However, for conceptual symmetry reasons and application value they
+   might come up in future versions.
 
+The `AMPERSAND`: `&` `&#0026` as first character in a path is defined as the marker for references to information elements.
 
-Topic Reference
-```````````````
-
-A topic reference is a data type that provides a resolvable reference to a Topic.
-
-::
-
-   TopicRef
-
-The underlying implementation shall not use more than 64 bits for the representation of TopicRef.
-
-All properties of the referenced topics shall be retrievable through the TopicRef.
-
-Data access to the content of the Topic shall be possible through the TopicRef.
-
-
-RP Reference
-`````````````
-
-An RP Reference is a data type that provides a resolvable reference to a Remote Procedure.
+In names and paths the reference marker to an information element of a path.
 
 ::
 
-   RPRef
+   &/Body/Doors/Windows/LeftFront/Position
 
-The underlying implementation shall not use more than 64 bits for the representation of RPRef.
+References the topic of the top left windows's position.
 
-All properties of the referenced Remote Procedure shall be retrievable through the RPRef.
+A Reference shall have a corresponding unique handle. The communication framework shall be able to dispatch handles of references like any other value of a data type. The underlying value type for handles should be ``UInt64`` and must have lockfree atomic read and write operations available.
 
-Invocation of the referenced Remote Procedure shall be possible through the RPRef.
+The application should not have access to handles directly, but only to the references themselves. We call the conversion of a handle into a reference “resolution of the handle”.
+
+The operations granted through a reference to an information item shall be identical to the operations of the information item itself.
 
 
-Namespaces
-----------
+**Implementation Note**
+
+Internally, the communication framework may actually only pass TopicRef’s to the application. From a semantic view it makes no difference to hold a TopicRef or a Topic directly.
+
+**Implementation Note**
+
+A ``TopicRef`` is *not* the same as ``&Topic`` as it may require additional validity checks.
 
 
 Topics
@@ -405,3 +400,4 @@ Zero Copy
 
 Safety
 ------
+
--- a/docs/features/communication/ipc/index.rst
+++ b/docs/features/communication/ipc/index.rst
@@ -19,7 +19,7 @@
 Communication Framework
 #######################
 
-The Communication Framework handles the safe, secure and performant exchange of information between software and/or hardware components of the Score stack.
+The Communication Framework handles the safe, secure and performant exchange of information between software and/or hardware components of the S-CORE stack.
 
 Communication covers information exchange between endpoints residing in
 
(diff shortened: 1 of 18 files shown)


docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,397 @@
...
+* Infrastructure Elements
+
+While the former carry the information exchanged the latter provide the means through which the information is exchanged.
+
+Information Elements
+````````````````````
+
+There are three fundamental concepts of information exchange defined. One communication element represents each concept:
+
+* **Topics**: A topic is the information carrier for **data**. A unique Id identifies a topic while a *data type* defines it's memory layout. The topic carries zero or multiple *values*. A value represents a single instances of the data type. See _`Topics`, _`Names`, _`Data Types`.
```
FScholPer, line 51: Is this something with is aligned to existent technologies like Iceoryx, uprotocol etc.? I would propose that it would make sense to check which technologies are already there and not to start from scratch. We can also discuss this in the first IPC incubation round.

HartmannNico: Yes and no. It is in line with current available communication systems like iceoryx2 (in fact, they did exactly this). Yet I disagree with the argumentation. This is a contribution request to ask for what we would like to have. In that sense, it does not matter how exactly it matches with existing technologies which have proven to be not sufficient.

FScholPer: You are right. Lets take it over as next steps after the finished incubation

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+#######################
+
+* Purpose - why and what do we need this for
+* Scope and context diagram 
+* QoS requirements
+* Error handling 
+* ASIL requirements (hashing, integrity, ...)
+* Interoperability 
+
+The Communication Framework handles the safe, secure and performant exchange of information between software and/or hardware components of the Score stack.
```
olivembo, line 29: Please update all instances to "SCORE" for consistency. 😊

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+Communication covers information exchange between endpoints residing in
+
+* the same or different process, running on
+* the same or different operating systems, deployed on
+* the same or different compute devices, built into
+* the same or different devices.
```
olivembo, line 36: What is the distinction between a "compute device" and a "device" in this context?
Compute device = CPU, GPU, etc.
Device = SoC

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+In current communication designs (SomeIP, Protobuf, Zenoh) communication on networks is at the centerpoint of considerations. We believe this to be the wrong approach for software defined machines. Network based communication circles around protocols based on a wire based communication paradigm, requiring serialization and segmentation of data.
```
olivembo, line 38: Official is "SOME/IP"

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+Instead, we promote a memory centered core paradigm and put Inter-Process Communication (IPC) into our conceptional focus. This allows for true zero copy design, does not segment or serialize data and fosters significantly higher performance in algorithms. As a bonus, it provides easy gateway mechanisms into serial communication like Ethernet and CAN.
```
olivembo, line 40: It seems IPC (general concept) is being conflated with Zero-Copy Communication (specific mechanism). Could you adjust the phrasing to make this distinction clearer? For example, you might specify "Zero-Copy Inter-Process Communication (IPC)" instead.

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+General Architecture
+--------------------
+
+The communication framework defines two central groups of elements:
+
+* Information Elements
+* Infrastructure Elements
+
+While the former carry the information exchanged the latter provide the means through which the information is exchanged.
```
olivembo, line 52: There seems to be a missing comma after "While the former carry the information exchanged" to separate the dependent clause from the main clause.

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+Information Elements
+````````````````````
+
+There are three fundamental concepts of information exchange defined. One communication element represents each concept:
+
+* **Topics**: A topic is the information carrier for **data**. A unique Id identifies a topic while a *data type* defines it's memory layout. The topic carries zero or multiple *values*. A value represents a single instances of the data type. See `Topics`_, `Names`_, `Data Types`_.
```
olivembo, line 59: The explanation is clear, but there’s a small typo: "it's memory layout" should be "its memory layout" (no apostrophe).

olivembo: Two small issues:
* "A Id handle" should be "An Id handle".
* "it's" should be "its" (no apostrophe).

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+* **Remote Procedures**: A remote procedure is the information carrier for **execution** progress. A Id handle identifies the remote procedure together with an ordered set of named *parameters*. Each parameter defined by a data type. A caller of a remote procedure can cause it's activation by invoking the remote procedure with passed *arguments*. An argument is a single value instance for a parameter. See `Remote Procedures`_, `Names`_, `Data Types`_.
```
olivembo, line 60: Two small issues:

    "A Id handle" should be "An Id handle".
    "it's" should be "its" (no apostrophe).

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+* **Events**: An event is the information carrier for runtime **synchronization**. A unique Id identifies the event. It signals the change of state. There is no data conveyed with the event. See `Events`_, `Names`_.
+
+While the Id uniquely identifies an information element within the communication framework, it can also have a *name* as alias to conveniently identify the element. While the Id may not be publicly known, the *name* allows for public lookup.
```
olivembo, line 63: If the _name_ also needs to be unique, it seems like the _Id_ might be redundant. It would be helpful to clarify the distinct roles of the _Id_ and the _name_ in order to highlight the different purposes of both.

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+
+Infrastructure Elements
+```````````````````````
+
+Infrastructure elements provide the means through which the information exchange executes.
+We define three fundamental building elements:
+
+* **Endpoints**: Endpoints are both the source and the target of every information exchange in teh communication framework. An endpoint providing information is consequently called a *provider*. With the same logic a *consumer* is an endpoint consuming information. Endpoints have an *Id* that uniquely identifies the endpoint within a node.
```
olivembo, line 72: Typo: "teh" should be "the."
In this sentence, you use the first time the term _node_. Maybe add in parenthesis that the term is defined below.

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+* **Nodes**: A node is an entity in the communication system that hosts several endpoints. It is the central element of the communication fabric by connecting endpoints and routing data. Nodes have an *Id* that uniquely identifies the node within a fabric. A node itself is also an endpoint.
```
olivembo, line 73: The last sentence "A node itself is also an endpoint." implies according to the definition of _endpoints_ that a node can contain other nodes ("Endpoints have an *Id* that uniquely identifies the endpoint within a node."). Is that the case?

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+* **Links**: A link is the fundamental abstraction of a connection between any two nodes. A link conveys information between nodes.
+
+The combination of NodeId and EndpointId we also refer to as *address*. As nodes are also endpoints, they implicitly have an address.
+
+Nodes and endpoints may also be identified by a *name* that resolves into references to these elements. See `Names`_, `References`_.
+
+Connecting nodes though links creates a mesh of nodes that can mutually exchange information utilizing the above concepts. The boundary of the mesh is at the sole discretion of the deployment and may span from a single application into a connected cloud environment.
```
olivembo, line 80: Here, the term _mesh_ is used, above the term _fabric_ is used. Recommendation is not to use multiple terms for the same "thing".

olivembo, line 80: Typo: "though" should be "through"

olivembo, line 80: Can you explain the use case of "only" connecting nodes via links? I understand the use case behind connecting endpoints of different nodes, but not the use case of "only" connecting nodes.

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+   namespace B
+      use A/*     // makes One and Two visible in B
+
+      algorithm
+         access One  // valid, One is visible in B
+         access Two  // valid, Two is visible in B
+
+
+Within a namespace elements from another namespace are visible without an explicit use when a resolving path is given.
+
```
olivembo, line 204: "Within a namespace" -> "Within a namespace, "

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+A handle is a numeric value that uniquely refers to an individual element in the communication system.
+
+The numeric value can safely be transported through the communication system and shall be resolvable into a reference of the original element.
+
+Data Types
+----------
+
+Data types describe the inner structure of data entities known as values.
+A specific data type will always have the identical memory layout, independent from compiler, operating system and controller architecture.
```
olivembo, line 226: How should that be possible regarding different alignment, padding, endianess, etc. of the different platform, OSs, and so on?

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+- A topic can have zero or multiple subscribers. A subscriber consumes the data published into the topic.
+
+Queue
+`````
+
+A topic stores it's values in a queue with a given depth. The queue has a policy that defines the behavior of new data published. There are four policies when the queue is full:
+
+- Ignore: New data is ignored.
+- Overwrite oldest: New data overwrites the oldest element. This makes the topic a ring buffer.
+- Replace latest: New data replaces the latest element. This makes the latest update always available.
```
olivembo, line 394: Can you explain this? E.g. the queue size is 3 and the queue is full. Then a new value arrives. The latest in the queue is replaced by the new one. What happens to the two others?

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+Namespace and Name
+``````````````````
+
+A topic can have a name that exists within a namespace. The name is not an attribute of the topic itself. Instead, it is an alias to the topic's reference that allows access to the data of the topic.
+
+Lifetime
+````````
+
+Once created, the topic belongs to the communication framework which determines it's lifetime.
```
olivembo, line 405: Typo "it's" should be "its"

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+Name & Namespace
+````````````````
+
+A remote procedure may have a name that exists in a namespace. The name is not a property of the remote procedure, but acts as an alias to a unique RPC reference associated with the remote procedure.
+
+Publishing & Discovery
+``````````````````````
+
+Attaching a name to a remote procedure means to publish the remote procedure.
+The communication framework owns both the name and the namespace.
```
olivembo, line 439: General question: If the communication framework owns everything (name, namespace of the procedure, topics, etc.) what happens, if there is a failure in the communication framework? Isn't it then the [single point of failure](https://en.wikipedia.org/wiki/Single_point_of_failure) and FFI according to ISO 26262-1:2018 cannot be fulfilled?

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+Information Elements
+````````````````````
+
+There are three fundamental concepts of information exchange defined. One communication element represents each concept:
+
+* **Topics**: A topic is the information carrier for **data**. A unique Id identifies a topic while a *data type* defines it's memory layout. The topic carries zero or multiple *values*. A value represents a single instances of the data type. See `Topics`_, `Names`_, `Data Types`_.
+* **Remote Procedures**: A remote procedure is the information carrier for **execution** progress. A Id handle identifies the remote procedure together with an ordered set of named *parameters*. Each parameter defined by a data type. A caller of a remote procedure can cause it's activation by invoking the remote procedure with passed *arguments*. An argument is a single value instance for a parameter. See `Remote Procedures`_, `Names`_, `Data Types`_.
+* **Events**: An event is the information carrier for runtime **synchronization**. A unique Id identifies the event. It signals the change of state. There is no data conveyed with the event. See `Events`_, `Names`_.
```
vevmaki, line 61: What state will change when a corresponding event will "occur"

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+Connecting nodes though links creates a mesh of nodes that can mutually exchange information utilizing the above concepts. The boundary of the mesh is at the sole discretion of the deployment and may span from a single application into a connected cloud environment.
+
+The entirety of connected nodes within a mesh we call *fabric*.
+
+.. admonition:: Design Note
+
+   From a perspective of safety, a node also encapsulates a single safety domain. Links provide the means for separating safety domains and thus allow for mixed criticality applications.
+
+^^^^ End of Big Picture ^^^^
```
vevmaki, line 88: I am fine with the approach of endpoints and nodes as long as we are talking about network infrastructure. However, if we want to apply this to shared memory concepts for IPC I wonder wether a shared memory element is a node?

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+
+- An event exists on it's own. The framework's data communication system owns the event.
+- An event can have zero or one publishers. The publisher updates ('publishes') new occurrences of associated state changes into the event. This is called 'triggering the event'.
+- An event can have zero or multiple subscribers, also called 'listeners'. A subscriber consumes the state change notifications.
+
+Immediate Events & Queued Events
+````````````````````````````````
+
+An event is designed to convey an immediate notification of the associated state change. However for cases where a subscriber cannot react immediately an event occurrence may be latched in a queue for deferred processing. This is called an 'Event Queue'. The framework may opt to offer event queues on top of immediate event propagation.
+
```
vevmaki, line 476: In case of queued events, is there are there ordering criteria like time...

Is there a chance to associate topics with RPCs or topics?

docs/features/communication/ipc/index.rst
HartmannNico: Exactly. Better terms welcome. The compute device executes instructions -> CPU, MCU, GPU, NPU; "Chip" would be too broad.
The "device" is referring to the ECU or board the compute device is bonded to.

docs/features/communication/ipc/index.rst
HartmannNico: IPC is the communication between processes (or better: communication endpoints) with data transport by means of memory. Zero-Copy is the desired mode of implementation to execute IPC. IPC is a prerequisite to enable Zero-Copy.

docs/features/communication/ipc/index.rst
HartmannNico: The state is an abstractum here. The point to make is that an event shall occur. The meaning of this occurrence is the change of any related state. The state is only semantically connected to the event. Example: an interrupt is abstracted in an event that occurs/fires/triggers because some data was received which caused a state change in the receiving chip.

docs/features/communication/ipc/index.rst
HartmannNico: The name is unique within the namespace it is defined.
The two items Id and Name were conveived to declare the Id to be something that is more compute-like (a number) while a name should foster human interaction. The Id could be a hash from the name, or any other unique identifier for an element.
Furthermore, many elements do not need to have a name, the idea is to allow anonymous elements with no name. Those would have an Id, but no mapping to a name.

docs/features/communication/ipc/index.rst
HartmannNico: In communication there will be messages that are addressing a node. With the definition that only endpoints initiate and consume information exchanged (aka messages), follows that a node must be an endpoint.
The logic that because of this a node can have sub-nodes I do not see, as a node is an endpoint but an endpoint is not a node.
A node in my opinion should not have "sub-nodes".

docs/features/communication/ipc/index.rst
HartmannNico: Practical consideration when designing a network. Locally, endpoints attach to their node. Nodes interconnect through links with each other. So an endpoint initiates a message Send to the attached node with a target address (Node+Endpoint). Node resolves the address, if it is local: dispatch to local endpoint. Otherwise: Find a proper Route to the target node, utilize the Link referenced in this Route and dispatch through the link. Semantically, an endpoint communicates to another endpoint, the nodes and links in between are invisible.

docs/features/communication/ipc/index.rst
olivembo: "In computer science, interprocess communication (IPC) is the sharing of data between running processes in a computer system." (see https://en.wikipedia.org/wiki/Inter-process_communication)
There are different approaches (https://en.wikipedia.org/wiki/Inter-process_communication#Approaches) of implementing IPC, where Shared Memory is one of the to achieve a better performance.
So I know that you mean "IPC via Shared-Memory", when you use the term "IPC", but for readers who read that the first time it is confusing.

docs/features/communication/ipc/index.rst
HartmannNico: It is possible. And your points are valid. 
That is why we cannot use unspecified memory layouts as they are used by C++ or Rust. Instead, we need formalized layouts like C. 
This again constrains the choices of data structures we can exchange. Not too much though. 
Pointers and references are forbidden. Serialization is slow, often requires code generation and hence is bad. 
Instead, data has to be coherent and relocatable for zero-copy. For collections this requires individual implementations, for primitives, tuple, array and struct we are good with C layout.

Mixed endianess is a show-stopper for zero-copy as it requires data conversion one way or the other. All mixed endian designs with memory data exchange are messy and should be avoided at all cost. Meaning: Hardware requirement. We are, after all, software-defined here.
The consistent way to cope with mixed endianess is serialization. Expensive and not Zero-Copy. Unless we entertain the view that serializing into and from a buffer is zero-copy.

docs/features/communication/ipc/index.rst
HartmannNico: I admit this is a theoretical and most likely very rare scenario. The other two values would stay the same, so you have the oldest unprocessed values in the queue and the latest written.
The use case is, when you want to have the latest value under all circumstances _and_ need to know the oldest one you did not process.

docs/features/communication/ipc/index.rst
HartmannNico: I assume queued items will have timestamps. My opinion: The timestamp is part of the data and not part of the queue item.

What do you mean by "associate" topics with RPCs?

docs/features/communication/ipc/index.rst
HartmannNico: --> Discussion point.

docs/features/communication/ipc/index.rst
vevmaki: I am fine when the sorting of elements in a queue is done via a timestamp of the element. 

I am wondering if queued events can trigger functions. So an event element would have a timestamp and a reference to function which will be executed by the queue's server.

docs/features/communication/ipc/index.rst
qor-lb: Examples to control the interpretation/layout of a respective memory location an therefore enable such a type system are e.g. [Cap'n Proto](https://capnproto.org/) or [rkyv](https://rkyv.org/)

docs/features/communication/ipc/index.rst
qor-lb: @HartmannNico could you add this use case in the feature request to make it apparent to future readers?
```suggestion
- Replace Latest: New data replaces the latest element, ensuring that the most recent update is always available. A typical use case is when you need to have the newest value at all times while still keeping track of the oldest value that has not been processed.
```

docs/features/communication/ipc/index.rst
```diff
@@ -11,6 +11,541 @@
...
+**Discussion**: An alternative to intertwined RPCs is the use of a Future mechanism. The advantage of the RPC reference based approach is the symmetry in invocation and result transmission. One implementation achieves both.
+
+Result
+``````
+
+A remote procedure may produce a result that is returned to the caller. The result also has a data type and consists of a single value.
+
+.. note::
+
+   Instead of passing back the result from the procedure the caller may pass a result-return reference that is a remote procedure itself. This way the framework may have a straight-forward way of implementing a Future mechanism that completes upon reception of the response call.
```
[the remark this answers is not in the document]
qor-lb:
```suggestion
   Instead of passing back the result from the procedure, the caller may pass a result-return reference that is a remote procedure itself. This way the framework may have a straight-forward way of implementing a Future mechanism that completes upon reception of the response call.
```

docs/features/communication/ipc/index.rst
qor-lb: @vevmaki could you please resolve the conversation if your question is answered? Otherwise we can also add it to the agenda on the IPC CFT meeting.

qor-lb:
Consolidated into #229 by @HartmannNico & @LittleHuba.

Issue #78, opened 2024-12-13
Setup documentation tooling (sphinx, sphinx-needs)


ltekieli:
* Introduce initial highlighting, preview and linting for .rst and .md files (-> esbonio) -> covered by #30
* Introduce initial code completion for .rst and .md files (-> templates and/or sphinx-needs extension) -> covered by 30
* Introduce first checks like mandatory attributes (-> pydantic based metamodel) < partially blocked by #39 (expected Jan 2025). But we can start with the basics etc.
* Enable drawing images (-> plantuml, mermaid and/or drawio)
* Extend meta model with first few elements
* Initial support for the first multi-repo use cases

AlexanderLanin:
Consider including tool requirements in this ticket

AlexanderLanin:
Updated list:
[x] Introduce initial highlighting, preview and linting for .rst and .md files (-> esbonio) (clone of #30)
[ ] Introduce initial code completion for .rst and .md files (-> templates and/or sphinx-needs extension) (clone of #30)
[ ] Introduce first checks like mandatory attributes (-> pydantic based metamodel)
[ ] Enable drawing images (-> plantuml, mermaid and/or drawio)
[ ] Extend meta model with first few elements

out of scope:
* Initial support for the first multi-repo use cases

AlexanderLanin:
This is at least 95% done. We'll open new dedicated issues for whatever is remaining.

PR #96, opened 2024-12-16
docs: preview in PRs

### LINKED ISSUE #94: Remove comment artifact url in every commit
Currently every commit in the main SCORE repository ends up in a new artifact url comment. This is not really needed and I propose to remove it from the workflow.
Need some feedback for that idea

AlexanderLanin:
# Current State
Documentation is built during PRs, and currently, an artifact is uploaded, with a link to the artifact added to the PR. However, this generates numerous links (e.g., see #94), and to view the documentation, users must download and extract the artifact manually.

# Desired State
Provide a direct link to the documentation preview in PRs. This link should allow users to preview the documentation directly, without extra steps, and remain stable throughout the PR's lifetime, unaffected by new commits.

# Solution
Since GitHub Pages supports only a single folder or artifact per repository, and rebuilding all versions of documentation for every PR is inefficient, a new approach is needed. The gh-pages branch will act as persistent storage for documentation artifacts. Each PR will store its specific documentation in a dedicated subfolder within this branch.

When a PR is updated, the workflow will add or update the corresponding documentation in its designated subfolder. Once the PR is closed or merged, its documentation subfolder will be removed to keep the branch manageable. This approach ensures stable links, efficient builds, and clear documentation previews.

This probably belongs to #78. And it closes #94.

Note: if this works as intended, it can be merged now. Then we can change the github pages settings to branch-based.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -11,17 +11,16 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 
-name: Documentation build
+name: Documentation
 on:
   pull_request:
     types: [opened, reopened, synchronize]
   push:
-    branches:
-      - main
   merge_group:
     types: [checks_requested]
 jobs:
   docs-build:
+    name: Build documentation
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -11,63 +11,167 @@
...
+        run: |
+          if [[ ${{github.event_name}} == 'pull_request' ]]; then
+            echo "target_folder=pr-${{github.event.pull_request.number}}" >> $GITHUB_ENV
+          elif [[ ${{github.ref_name}} != 'main' ]]; then
+            echo "target_folder=${{github.ref_name}}" >> $GITHUB_ENV
+          else
+            echo "target_folder=/" >> $GITHUB_ENV
+          fi
+
+      # TBD: Maybe we should not split the build and deploy jobs, but rather have a single job?
```
ltekieli, line 89: When merge queue is introduced you will want to skip the whole deploy job.

.github/workflows/docs.yml
```diff
@@ -11,63 +11,167 @@
...
+        if: env.target_folder != '/'
+        run: |
+          # Add the target folder to the versions file, if it's not already there
+          # (-qx: quiet, exact line match)
+          if ! grep -qx "${{env.target_folder}}" versions; then
+            echo "${{env.target_folder}}" >> versions
+          fi
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@v4
```
ltekieli, line 115: Please use fixed version of the action, eg. v4.7.2

.github/workflows/docs.yml
```diff
@@ -11,63 +11,167 @@
...
 
-name: Documentation build
+name: Documentation
 on:
   pull_request:
-    types: [opened, reopened, synchronize]
+    types: [opened, reopened, synchronize, closed]
   push:
     branches:
       - main
```
ltekieli, line 20: This filter needs to be removed.

ltekieli:
gh pages generation is set to workflow currently:
- https://github.com/eclipse-score/.eclipsefdn/blob/main/otterdog/eclipse-score.jsonnet#L62-L70

FScholPer (dismissed):
Thank you for the fast fix. Really good job done :)

.github/workflows/docs.yml
AlexanderLanin: Ok, I'll keep it separated.

.github/workflows/docs.yml
ltekieli: I meant the "branches: main" part, but ignore it, I just thought this was the reason the documentation build didn't run, but there has to be another reason.

AlexanderLanin: Incorrect indentation made the yaml file unparsable. It seems there is no error message in such cases!

AlexanderLanin:
> gh pages generation is set to workflow currently:
> 
> * https://github.com/eclipse-score/.eclipsefdn/blob/main/otterdog/eclipse-score.jsonnet#L62-L70

I guess I'll change this PR to try uploading both ways - when building "latest".
Then we can change the other repository.
Then we undo those temporary changes.

AlexanderLanin:
This is quite ugly, I will split cleanup into a separate workflow, so it doesn't show up here.
image

AlexanderLanin:
One addition: someone needs to create a `gh-pages` branch here in the repo.

FScholPer (approved):

ltekieli (approved):

PR #117, opened 2024-12-17
git: Remove requirement for commit body to reference an issue

AlexanderLanin:
Referencing an issue is not required according to GD_GUIDL__Contr_Request_Guidance.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -32,11 +32,6 @@ regex=^[a-z_-]+: .+$
 # First line of the commit message body must be empty. (second line of the commit message)
 [body-first-line-empty]
 
-# Body must contain a reference to an issue
-# 'see' will act as a form of reference from commits to issues without closing them.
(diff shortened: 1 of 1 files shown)


ltekieli (changes requested):
Commit  message is described in git guidelines and still requires issue-ref:
- https://eclipse-score.github.io/score/process/guidelines/git/index.html

ltekieli:
I'd rather vote for some defined prefix that will skip the "Issue-ref" check. "fixup:", "chore:" that explicitly indicates that the official process is omitted. The sentence:
```For simple contributions, like fixing bugs or improvements, just creating a PR is sufficient.``` is to subjective, and just skipping the issue ref completely also is not really valid.

AlexanderLanin:
I guess this qualifies as a non simple contribution. So I'll create an issue :-)

Issue #32, opened 2024-12-05
Initial Structure of the Platform Management Plan


antonkri:
Provide initial structure of the Project Management Plan to the project

PR #122, opened 2024-12-17
Add initial platform management plan structure

antonkri:
Add initial platform management plan structure.

Issue-ref: #32


aschemmel-tech (approved):
no findings

PR #144, opened 2025-01-09
move docs tooling to _tooling

### LINKED ISSUE #139: Improvement/ADR: split docs tooling and docs content
## Current situation

Currently content and tooling of `docs` is mixed up, which makes it difficult for tool maintainers and tool users to find the relevant files. It's not as bad currently, since there is little tooling. But it will grow. Just have a look at the current situation:
> BUILD  -> tooling
> _assets -> docs
> _extensions -> tooling
> _templates -> tooling
> conf.py -> tooling
> features -> docs
> incremental.py -> tooling
> index.rst -> docs
> platform_management_plan -> docs
> process -> docs
> requirements -> docs
> requirements.txt -> tooling
> requirements_lock.txt -> tooling
> score_releases -> docs

## Proposal

Move all sphinx-runtime-related tooling to either `/.docs` or to `/docs/.tooling`.

## Constraints

* To ease working with python code, it would be sensible to have all sphinx-runtime-related python code in the same directory as `conf.py`. This way when we `import metamodel` or something like that, we don't need to go though bazel on every change.
* Moving all tooling to one directory should make reuse across repositories easier. Although that is currently not a concern. Yet.
* No additional constraints anticipated from CI, esbonio, incremental or "classic" build options.

AlexanderLanin:
closes #139 

I've left conf.py in docs due to a limitation by sphinx_docs (used by docs:docs), which assumes index.rst and conf.py are in the same directory. They are intentionally not using the conf-dir parameter 😢


docs/_tooling/incremental.py
```diff
@@ -12,31 +12,23 @@
...
-        "This script must be run from Bazel via `bazel run //docs:incremental`"
-        " (sphinx not installed)."
-    )
-
-workspace = os.getenv("BUILD_WORKSPACE_DIRECTORY")
-if not workspace:
-    sys.exit(
-        "This script must be run from Bazel via `bazel run //docs:incremental`"
-        " (BUILD_WORKSPACE_DIRECTORY not set)."
-    )
```
MaximilianSoerenPollak, line 30: Was this intentional that you deleted the 'safeguard' of executing it only from bazel?

docs/BUILD
```diff
@@ -74,15 +74,15 @@ sphinx_docs(
...
 )
 
 py_library(
     name = "extensions",
-    srcs = [
-        "_extensions/layouts.py",
-        "_extensions/metamodel.py",
-    ],
+    srcs =
+        glob(["_tooling/conf_extras/*.py"]) + glob([
```
MaximilianSoerenPollak, line 83: Is it possible to have it both in one glob? Or doesn't that work due to them being different subfolders?

MaximilianSoerenPollak (commented):
Just two questions that I had. Other than that this looks good.

AlexanderLanin:
fotce push 1: auto format is not enabled in vs code :(
force push 2: even the trivial commit needed a commit body :(
force push 3: prefix commit with "docs:"

docs/_tooling/incremental.py
MaximilianSoerenPollak: That makes sense.

MaximilianSoerenPollak (approved):
LGTM

pahmann (approved):
Peer reviewed with @AlexanderLanin

PR #145, opened 2025-01-09
git: update gitlint documentation

AlexanderLanin:
The documentation of enabled/disabled rules was a little off. And I was wondering why a body is required, until I found the commit message guidelines, so it seems to make sense to add that link.


MaximilianSoerenPollak (approved):
This seems to be consistent with what I can read in the git guidelines, so LGTM.

PR #146, opened 2025-01-09
docs: add recommended extensions

AlexanderLanin:
This adds drawio, rst and esbonio extensions. As those three are quite typically useful when editing documentation.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,2 +1,12 @@
 {
+    "recommendations": [
+        // Editing *.drawio.svg files directly in VS Code
+        "hediet.vscode-drawio",
+
+        // Some convenient extensions for editing reStructuredText files
+        "lextudio.restructuredtext",
+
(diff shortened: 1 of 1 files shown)


MaximilianSoerenPollak:
Question is if the rst extension is really needed as it offers (at least as far as I know right now) no additional benefit. Would it then not be better to leave it out? 
But could be that I'm assuming based on old information here.

AlexanderLanin:
I've checked in detail what the rst extension promises, that is not moved to esbonio. Documentation seems outdated, so I looked through the source code. At least I tried. The list may be incomplete / slightly wrong.

* linked documents are clickable
* drag n drop documents into rst files
* support for writing tables (add columns, rows, shift enter, alt enter, tab to next cell, ...)
* commands like unterline, createGrid, dataToTable
* snippets

So there is definitely something.

pahmann (approved):

PR #147, opened 2025-01-10
sphinx-needs: Added sphinx_extension folders

MaximilianSoerenPollak:
The sphinx_extension folders are moved over with this commit. This adds multiple warnings for different requirements. These warnings check if certain parameters are fulfilled when the documentation builds.

**Note: The warnings are NOW not interpreted as errors, to allow for a successfull build but give stakeholders the opportunity to see what checks for what reasons fail.**

Currently (in my local build) there are 198 warnings, mainly containing to workflows and workproducts. 

The tests that are inside of the `sphinx_extensions` folder can currently not be run as there is no testing framework setup yet.

The structure of this should be compatible with #144 as sphinx_extensions also is in `_tooling`


AlexanderLanin:
LGTM (the goal was to merge as-is and improve later)

ltekieli (approved):

PR #153, opened 2025-01-13
docs: registry modules

### LINKED ISSUE #125: Create bazel registry
- registry is set up and usable
- documentation is updated

ltekieli:
Adds details about:
- module entries in registry
- module versioning and compatibility

resolves #125

### DIFF
--- a/docs/features/infrastructure/bazel_registry.rst
+++ b/docs/features/infrastructure/bazel_registry.rst
@@ -12,6 +12,8 @@
    # SPDX-License-Identifier: Apache-2.0
    # *******************************************************************************
 
+.. _bazel_registry:
+
 Bazel registry
 ##############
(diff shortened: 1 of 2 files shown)


AlexanderLanin:
LGTM. I guess we'll describe it in more details in the future, but that's ok.

Regarding versioning, we should agree on some format, e.g. SemVer or date based across all our modules?!

antonkri (approved):

ltekieli:
> LGTM. I guess we'll describe it in more details in the future, but that's ok.

I would wait for the first actual module to be integrated and then discuss in detail.

> Regarding versioning, we should agree on some format, e.g. SemVer or date based across all our modules?!

Yes as long as it's compliant with the bazel format.

PR #175, opened 2025-01-16
python-testing: Added pytest capabilities

### LINKED ISSUE #159: Improvement: Integrate pytest
## Why is it needed? 

Currently we have no way of executing python defined tests via bazel.   
There are tests already inside `sphinx_extensions`, but we have no way of running these tests. 

## What should it do?

The tooling should be able allow for test to be run via 'pytests'. It should be possible to be called from a bazel BUILD files.
There also should be a way to test specific sub-folders / languages etc. 

### Tasks: 

- [ ] Integrate Pytest into bazel 
- [ ] Ensure current tests can be executed correctly
- [ ] Provide simple ways to call a subset of tests (for example, call all tests for python)
- [ ] Document how to use implementation for own developments as well as executing it
- [ ] integrate into IDEs like pytest normally does: https://pytest-with-eric.com/introduction/how-to-run-pytest-in-vscode/

MaximilianSoerenPollak:
Added functionality for pytest into bazel.
Possible to now add tests that get executed via  'bazel test //...'. 
Documented possibilty of adding tags to tests.
Added pytest integration into VSCode to autodiscover tests & execute those.

closes #159


README.md
```diff
@@ -64,6 +64,22 @@ point your IDE to the .venv_docs virtual environment.
...
+With the following command it is possible to run all available tests:
+
+```
+$ bazel test //...
+``` 
+
+However it's also doable to just run specific tests. For example all python tests:
+
+```
+$ bazel query 'kind(py.*, tests(//...))' | xargs bazel tests
```
AlexanderLanin, line 78: Can we also have that in usable? :D
I'm not even sure what this is filtering for. Doesn't it also run all tests?

docs/_tooling/sphinx_extensions/sphinx_extensions/checks.py
```diff
@@ -12,18 +12,20 @@
 # *******************************************************************************
 from sphinx_needs.api.configuration import add_warning
 from sphinx.application import Sphinx
-from sphinx_extensions.requirements.checks.id import (
+from docs._tooling.sphinx_extensions.sphinx_extensions.requirements.checks.id import (
```
AlexanderLanin, line 15: People will not know that they need to import this way. Since the old import doesn't show up as an error while writing the script.

tools/testing/pytest/MODULE.bazel
```diff
@@ -0,0 +1,46 @@
...
+)
+
+###############################################################################
+#
+# Python version
+#
+###############################################################################
+bazel_dep(name = "rules_python", version = "1.0.0")
+
+PYTHON_VERSION = "3.12"
```
AlexanderLanin, line 26: If this has an independ python version... it could happen that the tests pass with a python version we don't care about, but fail with the python version we are actually using.

Same problem for all the other dependencies.

tools/testing/pytest/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+load("@score_pytest//:defs.bzl", "py_pytest")
+
+py_pytest(
+    name = "test_my_first_check",
+    srcs = [
+        "test_my_first_check.py"
+    ],
+    plugins = [
+        # Specify optional plugins, that will register their fixtures
```
AlexanderLanin, line 21:
```suggestion
        # Optionally specify pytest plugins, that will register their fixtures
```

tools/testing/pytest/README.md
```diff
@@ -0,0 +1,45 @@
...
+    ],
+    args = [
+        # Specify optional arguments, ex:
+        "--basetemp /tmp/pytest",
+    ],
+    # Optionally provide pytest.ini file, that will override the default one
+    pytest_ini = "//my_pytest:my_pytest_ini",
+
+    # Optionally provide tags the test should have, in order to allow for execution grouping
```
AlexanderLanin, line 30: How would one use execution grouping? And what is it?

tools/testing/pytest/bazel/py_pytest.bzl
```diff
@@ -0,0 +1,50 @@
...
+    plugins = ["-p %s" % plugin for plugin in plugins]
+
+    py_test(
+        name = name,
+        srcs = [
+            pytest_bootstrap,
+        ] + srcs,
+        main = pytest_bootstrap,
+        args = args +
+               ["-c $(location %s)" % pytest_ini] +
```
AlexanderLanin, line 33: We have the script location here? Can we add that to sys.path?

tools/testing/pytest/README.md
```diff
@@ -0,0 +1,45 @@
...
+MODULE.bazel
+```
+bazel_dep(name = "score_pytest", version = "0.1")
+```
+
+BUILD
+```
+load("@score_pytest//:defs.bzl", "py_pytest")
+
+py_pytest(
```
AlexanderLanin, line 15: How about dependencies?

And how do we handle versioning?
e.g. when in general I'm using dependency A, B and C.
For testing I'm only using A and B.

It can easily happen that I get different versions for A and B for testing, then for my production script. 
Can we reuse the "normal lockfile"?

README.md
MaximilianSoerenPollak: It basically queries all test that are available to bazel for tests in `python` and then executes those only. 

One can still use tags to execute specific tests but those should not be language specific, but rather different grouping like 'smoke tests' etc. 

That's why this is the approach chosen for language specific test runs if that is wanted.

docs/_tooling/sphinx_extensions/sphinx_extensions/checks.py
MaximilianSoerenPollak: How would you best solve this? Document it for sure, but I'm unsure if that is enough.

tools/testing/pytest/MODULE.bazel
MaximilianSoerenPollak: I will do some testing and see if this can be removed & the one from the main one inherited.

tools/testing/pytest/README.md
MaximilianSoerenPollak: Will add some examples. 

Basically you can add tags to a test with the `tags = [...]` attribute.  For example `tags = ["smoke-test", "integration-test"]` 
Then you can execute ALL tests that have a tag you specify via the command line like so: 
`bazel test --test_tag_filters=smoke-tests` then all tests that have at least one of the tags mentioned (or if you add a `-` infront it negates the tag so only test that do not have that tag) are being run. 

The other way is the one mentioned above via querying different test kinds `bazel query 'kind(py.*, tests(//...))' | xargs bazel tests`

AlexanderLanin: I'd rather add feature-like tags like `docs`, `docs-build` or `docs-as-code` :-)

tools/testing/pytest/bazel/py_pytest.bzl
MaximilianSoerenPollak: This is the location of a potential different 'pytest_ini' that you can supply. So adding this to Path would make not much sense as we don't know if it 's supplied or not, but correct me if I'm wrong in my assumption here.

If it is not supplied the default implementation will be used which is defined like this: 
```
    if not pytest_ini:
        pytest_ini = Label("@score_pytest//:pytest.ini")
```

AlexanderLanin: Oh right, sorry. I haven't used that modulo string formatting operator in.... ever :D

tools/testing/pytest/MODULE.bazel
MaximilianSoerenPollak: It now inherits from the main MODULE. I'm testing if also requirements etc. are available.

tools/testing/pytest/README.md
MaximilianSoerenPollak: This now works. The pytest implementation relies on the packages that are used in `docs/_tooling/requirements.txt` and we can define packages there to also use in our tests. 

This should then guarantee that we will not have any versioning issues.

tools/testing/pytest/README.md
MaximilianSoerenPollak: This is changed in newest commit.

tools/testing/pytest/README.md
AlexanderLanin: Now it will use docs requirements for any kind of python test... we'll need a better concept in the future.

AlexanderLanin:
LGTM

AlexanderLanin (approved):

tools/testing/pytest/README.md
MaximilianSoerenPollak: Yes I agree.
Something that allows us to have python requirements in one place but also inherit doc requirements where needed.

PR #208, opened 2025-01-22
Docs-check-refactor

AlexanderLanin:
**Update**

As we have merged #226 this PR is now a draft of how the rest could be adapted. Note that it is incomplete and not working. We'll keep this PR for a few days as reference for a proper fix.


hoe-jo:
Shall we place the metamodel in a separate directory where process group is codeowner?

AlexanderLanin:
> Shall we place the metamodel in a separate directory where process group is codeowner?

I was even considering moving it completely to the process directory. We'll see.
Anyhow CODEOWNERS also works on individual files, not only directories.

docs/_tooling/extensions/score_metamodel/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+        check(needs, log)
+
+    if log.has_warnings:
+        log.warning("Some needs have issues. See the log for more information.")
+        # TODO: exit code
+
+
+def setup(app: Sphinx):
+    app.config.needs_id_required = True
+    app.config.needs_id_regex = "^[A-Za-z0-9_-]{6,}"
```
hoe-jo, line 51: Pls adapt ID to defined format

AlexanderLanin: My original goal was a no-change refactoring... but I failed horribly 😆 

In order to make the tests pass, I had to adjust them a little... and a little more... and a little more... and then I thought, it would just be easier to take a different approach 🤣 

I'll look into this setting as well. Potentially removing it completely, as ID format is better checked via the usual metamodel checks anyway.

PR #226, opened 2025-01-27
docs: score_metamodel

### LINKED ISSUE #207: Improvement: Combine Sphinx Needs Metamodel and Models for Check into one structure
Currently the model for the check is a separate model than the sphinx needs model. It needs to be restructured into one file so that it is easier to maintain.

AlexanderLanin:
Prepare score_metamodel as the new central place for the score metamodel.

This is a small subset of #208 which got out of hand.
Closes #207 

Changes:

* improve VS Code performance by eclusing some directories
* remove python.envFile as pytest is working fine without it after all
* add score_metamodel with everything metamodel related (e.g. checks)
* add ONE check to the score_metamodel
* merge two metamodel fles into one
* last but not least... add a warning to bazel pytest directive in case src is empty to avoid cryptic json errors


MaximilianSoerenPollak (approved):
LGTM

antonkri (approved):

PR #289, opened 2025-02-06
docs: Added decision records for PlantUML and Draw.io

### LINKED ISSUE #183: docs: document why we chose plantuml + drawio
PlantUML is being introduced in #158, however there is no explanation why PlantUML was chosen. We also support Draw.IO, again without any explanation why Draw.IO was chosen.

We need to document the decision, even if it's too late.

In case score process does not offer any advice on how to do that, have a look at https://github.com/joelparkerhenderson/architecture-decision-record

Note: it does not need to have an unreasonable level of detail or fake arguments. If it's a simple reason, document the simple reason and that's it.

dcalavrezo-qorix:
Created an ADR documenting the decision rationale for the tools used to reneder diagrams.

Fixes #183

### DIFF
--- a/tools/decision_records/plantUML_draw.md
+++ b/tools/decision_records/plantUML_draw.md
@@ -1,0 +1,52 @@
+# Use of PlantUML and Draw.IO in the Score Project
+
+## Status
+
+Accepted
+
+## Context
+
+The Score project requires a consistent and maintainable approach for creating and managing architectural and technical diagrams. The primary concerns driving this decision include:
+
+- **Version Control:** The ability to store and manage diagrams as code to ensure traceability and history tracking.
+- **Collaboration:** Allowing multiple contributors to edit and refine diagrams without requiring proprietary software.
+- **Automation:** Supporting automated documentation generation.
+- **Ease of Use:** Enabling contributors with varying levels of expertise to create and update diagrams efficiently.
+- **Ease of Integration with Sphinx and Sphinx-Needs:** Ensuring diagrams can be seamlessly incorporated into documentation workflows using Sphinx and Sphinx-Needs.
+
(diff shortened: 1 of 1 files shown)


tools/decision_records/plantUML_draw.md
```diff
@@ -0,0 +1,52 @@
...
+- **Positive Impacts:**
+  - Diagrams can be stored and managed within the Git repository alongside documentation.
+  - Contributors can choose between textual and graphical diagram creation.
+  - Enables automation in documentation pipelines.
+  - Ensures long-term maintainability and accessibility of diagrams.
+  - Seamless integration with **Sphinx and Sphinx-Needs**, improving documentation workflows.
+
+- **Challenges:**
+  - Requires contributors to be familiar with PlantUML syntax if they opt for code-based diagrams.
+  - Draw.IO files should be properly version-controlled and handled to avoid conflicts.
```
AlexanderLanin, line 50: We use the .drawio.svg extension for images. We don't need additional version control beyond that.

AlexanderLanin (changes requested):

AlexanderLanin:
@MaximilianSoerenPollak any idea where we should put this file? Otherwise we can merge it as-is and discuss it later.

AlexanderLanin:
@hoe-jo this is a perfect example for a mini comment + mini fix + no delta review possibility for the reviewer.

AlexanderLanin (approved):

AlexanderLanin:
@dcalavrezo-qorix FYI `Signed-off-by` currently has no meaning in score.

dcalavrezo-qorix:
> @dcalavrezo-qorix FYI `Signed-off-by` currently has no meaning in score.

Could be, but when I read the guide the following was stated ( more or less in this form): `To comply with the DCO, every commit message must include a Signed-off-by line`

MaximilianSoerenPollak:
> @MaximilianSoerenPollak any idea where we should put this file? Otherwise we can merge it as-is and discuss it later.

There is a `decision_records` that I planned on adding to `_tooling`. Just the question if that makes more sense instead of the `tooling` one as the draw.io & plantuml are used for the Docs mainly.

PR #302, opened 2025-02-07
doc: Update Score Process Model

masc2023:
Update model according definition in process_description


masc2023:
> Proposal to remove the RASIC relevant charts, as they clutter the process Image. Furthermore the fulfillment of the RASIC should be part of the SCORE Guidance.

Would not fully agree, as information is already in the current version, so please let us discuss to come to an agreement in next Community Meeting

FScholPer (approved):

PR #369, opened 2025-02-13
docs: Convert sphinx metamodel to yaml

### LINKED ISSUE #286: docs: convert metamodel to yaml
Currently `docs/_tooling/extensions/score_metamodel/metamodel.py` contains only configuration, there is really no benefit in having it as python code.

Task:
* Convert from `py` to `yaml`
* Allow comments in yaml! (This means you cannot naivly use `yaml` package)
* Provide content to sphinx_needs in `setup()` of `score_metamodel`

Proposed scope of this first task, so we don't need to change the entire yaml twice:
* needs_extra_options can be computed from the other existing fields (set of all required options)
* links should be a dict and not a list.... e.g.:
```
needs_extra_links:
  contains:
    incoming: is part of
    outgoing: consists of
  responsible:
    incoming: is responsible for
    outgoing: responsible
```
* Same for needs_types:
```
types:
  workproduct:
    title: Workproduct
    prefix: WP__
    color: "#DDDD00" # yellow
    style: artifact
    mandatory_options:
      id: ^WP_.*$
      status: ^(draft|valid|invalid|volatile)$
      tags: ^.*$
      relevant: ^.*$
    mandatory_links:
      compliance-wp: ^.*$
```
* In needs_types split mandatory_options and mandatory_links (see above)
* Provide schema.json. It can be referenced from yaml via `# yaml-language-server: $schema=./metamodel-schema.json`

Future, out of scope for now:
* Default-values for e.g. status? References as in schema.json?
* Replace regex with e.g. enums?
* For links replace regex with the directive that is being linked
* Include Prefix definition in ID check so that ducplicate definition can be removed

nicu1989:
# Improvement

## Description

Convert metamodel.py to metamodel.yaml since it only contains data structures.

## Related ticket

closes #286 (improvement ticket)


docs/_tooling/extensions/score_metamodel/__init__.py
```diff
@@ -82,12 +85,82 @@ def _run_checks(app: Sphinx, exception: Exception | None) -> None:
...
 def setup(app: Sphinx):
     app.config.needs_id_required = True
     app.config.needs_id_regex = "^[A-Za-z0-9_-]{6,}"
-    app.config.needs_types = metamodel.needs_types
-    app.config.needs_extra_links = metamodel.needs_extra_links
-    app.config.needs_extra_options = metamodel.needs_extra_options
+
+    # load metamodel.yaml via ruamel.yaml
+    here = os.path.abspath(os.path.dirname(__file__))
+    yaml_path = os.path.join(here, "metamodel.yaml")
```
AlexanderLanin, line 97: I'm personally in favor of `pathlib.Path` for such stuff

docs/_tooling/extensions/score_metamodel/__init__.py
```diff
@@ -82,12 +85,82 @@ def _run_checks(app: Sphinx, exception: Exception | None) -> None:
...
+
+
 def setup(app: Sphinx):
     app.config.needs_id_required = True
     app.config.needs_id_regex = "^[A-Za-z0-9_-]{6,}"
-    app.config.needs_types = metamodel.needs_types
-    app.config.needs_extra_links = metamodel.needs_extra_links
-    app.config.needs_extra_options = metamodel.needs_extra_options
+
+    # load metamodel.yaml via ruamel.yaml
```
AlexanderLanin, line 95: Can you move this to a function please

docs/_tooling/extensions/score_metamodel/__init__.py
```diff
@@ -82,12 +85,82 @@ def _run_checks(app: Sphinx, exception: Exception | None) -> None:
...
+
+        if "color" in directive_data:
+            one_type["color"] = directive_data["color"]
+        if "style" in directive_data:
+            one_type["style"] = directive_data["style"]
+
+        # mandatory_options => "req_opt" as list of (name, regex)
+        req_opts_yaml = directive_data.get("mandatory_options", {})
+        req_opt_list = [(k, v) for k, v in req_opts_yaml.items()]
+        one_type["req_opt"] = req_opt_list
```
AlexanderLanin, line 123: Note: keeping `req_opt` and `opt_opt` as-is is certainly possible, but not mandatory.

Those two we only use in our own checks. So if you want to adapt them to a dict or whatever. That's totally possible!

docs/_tooling/extensions/score_metamodel/__init__.py
```diff
@@ -82,12 +85,82 @@ def _run_checks(app: Sphinx, exception: Exception | None) -> None:
...
+
+        # optional_options => "opt_opt" as list of (name, regex)
+        opt_opts_yaml = directive_data.get("optional_options", {})
+        opt_opt_list = [(k, v) for k, v in opt_opts_yaml.items()]
+        one_type["opt_opt"] = opt_opt_list
+
+        # mandatory_links => "req_link"
+        mand_links_yaml = directive_data.get("mandatory_links", {})
+        if mand_links_yaml:
+            one_type["req_link"] = [(k, v) for k, v in mand_links_yaml.items()]
```
AlexanderLanin, line 133: Extracting `req_link` as a separate attribute will require adaptations to `check_options`.

docs/_tooling/requirements.txt
```diff
@@ -11,6 +11,7 @@ pydata-sphinx-theme
 sphinx-design
 sphinxcontrib-plantuml
 sphinx-autobuild
+ruamel.yaml
```
AlexanderLanin, line 14: This here lists the package as beta, is that correct?
https://pypi.org/project/ruamel.yaml/

docs/_tooling/extensions/score_metamodel/__init__.py
nicu1989: Done

docs/_tooling/extensions/score_metamodel/__init__.py
nicu1989: Done, looks better with pathlib

docs/_tooling/extensions/score_metamodel/__init__.py
nicu1989: done, moved to dict

docs/_tooling/extensions/score_metamodel/__init__.py
nicu1989: done, I had it on to do list but forgot :)

docs/_tooling/requirements.txt
nicu1989: true its beta, I dont see other option for round-trip comment preservation.

docs/_tooling/extensions/score_metamodel/metamodel.yaml
```diff
@@ -0,0 +1,388 @@
...
+  ##############################################################################
+  # Process Metamodel
+  ##############################################################################
+  # Standards
+  std_req:
+    title: "Standard Requirement"
+    prefix: "STD_REQ__"
+    mandatory_options:
+      opt_id: "^(STD_REQ_ISO26262|STD_REQ_ISO21434|STD_(G|B)P_ASPICE-40)__[0-9a-z_]*$"
+      opt_status: "^(valid)$"
```
AlexanderLanin, line 15: what is the opt prefix here? And what is opt_id?

nicu1989: It represents the req_opt id and req_opt status, both for id and for status I needed to add a prefix due to sphinx reserved option name. For example if I used only id:
`sphinx_needs.needs.NeedsConfigException: Extra option "id" already used internally.  Please use another name in your config (needs_extra_options).`

nicu1989: I think that the way to go. This way the yaml file will be kept "clean". I will open a PR.

Issue #318, opened 2025-02-10
Improvement: Add ASPICE standard references for process


masc2023:
Document in score repository (until the process_description repo is available)

Create documentation according to "template": https://github.com/eclipse-score/process_description/blob/main/requirements_engineering/index.rst

this includes workproducts definition and linking to standard
this includes guidance definition and linking to standard
Prepare presentation in audit session

PR #372, opened 2025-02-14
process: add aspice standard

masc2023:
Tailored processes of APSICE PAM 4.0 are added

Resolves: #318

### DIFF
--- a/docs/process/standards/aspice/aspice.rst
+++ b/docs/process/standards/aspice/aspice.rst
@@ -1,0 +1,431 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2025 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+ASPICE 4.0
+==========
+
+This incorporates the detailed description of the `Automotive SPICE PAM
+v4.0 <https://vda-qmc.de/wp-content/uploads/2023/12/Automotive-SPICE-PAM-v40.pdf>`__
+to support the performance of process assessments, so that the process
+assessment model can be used for its intended purpose.
+
+`Automotive SPICE PAM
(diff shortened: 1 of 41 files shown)


AlexanderLanin:
@hoe-jo why don't we have something like `.. standard-req::`, `.. norm-req::`, `.. iso-req::` or however we would call these?

masc2023:
@hoe-jo . can we allow Upper Letter for Standards, e.g. STD_BP_ASPICE-40__SWE-5-BP5

docs/process/standards/aspice/aspice.rst
```diff
@@ -0,0 +1,431 @@
...
+8. Interfaces between the involved parties are managed to ensure both effective communication and the assignment of responsibilities.
+
+
+Generic practices
+~~~~~~~~~~~~~~~~~
+
+.. std_req:: GP2.1.1: Identify the objectives and define a strategy for the performance of the process.
+   :id: STD_GP_ASPICE-40__GP-211
+   :status: valid
+   :links: STD_BP_ASPICE-40__IIC-19-01
```
aschemmel-tech, line 101: do we have "links" in the metamodel?

masc2023: No, is inbuild feature, already available in POC

docs/process/standards/aspice/aspice.rst
```diff
@@ -0,0 +1,431 @@
...
+6. Persons performing the process are prepared for executing their responsibilities.
+7. Physical and material resources for performing the process are identified, made available, allocated and used.
+8. Interfaces between the involved parties are managed to ensure both effective communication and the assignment of responsibilities.
+
+
+Generic practices
+~~~~~~~~~~~~~~~~~
+
+.. std_req:: GP2.1.1: Identify the objectives and define a strategy for the performance of the process.
+   :id: STD_GP_ASPICE-40__GP-211
```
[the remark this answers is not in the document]
masc2023: done

hoe-jo:
@AlexanderLanin we do have the directive std_req

@masc2023 i rather would not have an exception here, we could simply reformat the IDs with regex (also inside VScode possible) one way could be:
Search: \b(([A-Z0-9-]+_){1,4})_*(?=[a-z])
Replace: \L$1_

masc2023:
@hoe-jo , already changed it, please check if it is ok, currently the checks does still check for STD_REQ_......
> @masc2023 i rather would not have an exception here, we could simply reformat the IDs with regex (also inside VScode possible) one way could be: Search: \b(([A-Z0-9-]+_){1,4})_*(?=[a-z]) Replace: \L$1_

aschemmel-tech (approved):
Fine

Issue #393, opened 2025-02-17
C++: example repository


AlexanderLanin:
### Goal
As a module/component developer I would like to start working on my C++ code. I don't know here to start. I would like a template with a hello world application and and hello world requirement.

**Technically this means:**
Integrate the open sourced LLVM toolchain for Bazel into S-CORE to enable C++ compilation and execution.

### Description
Since there are still no clear requirements which host tool for compilation and execution we need to use, the best option is to go with community provided [LLVM Toolchain](https://github.com/bazel-contrib/toolchains_llvm).
With integrating this toolchain we are avoiding integration complexity that other toolchains can bring. In case later we need to integreate additional compiler (e.g. GCC) we can always keep this and as addtion add new toolchain (with proper configuration).
The current setup has already been done in PoC so we just need to migrate same solution to S-CORE.

### Minimal features
:exclamation: Successfully compile and run C++ code using LLVM.
:exclamation: Provide a simple Hello World example to validate functionality.

### Out of Scope (Optional / Not Required)
:grey_exclamation: IDE integration (C++ language server, compiler output visualization, etc.).
:grey_exclamation: clang-format (including IDE support).
:grey_exclamation: clang-tidy (including IDE support).
:grey_exclamation: Dependency management support (over bazel? via e.g. vcpkg or conan)

Note: we do not have an agreement on any of these tools yet, as there is no-one to agree them with. The problem is already addressed to tech lead circle.

For now we would probably do all of this in `score`. Later on we can move it to `tooling`. -> will be discussed in [next daily](https://github.com/orgs/eclipse-score/discussions/236#discussioncomment-12228486)

ltekieli:
Usable C++ toolchains are provided as a bazel submodule here: https://github.com/bazel-contrib/toolchains_llvm

AlexanderLanin:
That was one of the ideas. Would you align with @nradakovic and improve the issue description?

AlexanderLanin:
> [@AlexanderLanin](https://github.com/AlexanderLanin) I can't update description of ticket so I will basicly add stuff as comment. If you think this should be moved into description, feel free to do so.

Description updated.

FScholPer:
@antonkri can you please migrate the old stuff to the new repo?

FScholPer:
now we have two of them. scrample and example. I thin we can close it now

PR #413, opened 2025-02-18
llvm: Add LLVM Toolchain config

nradakovic:
Add LLVM Toolchain configuration for Bazel into score project:

Issue-ref: #393

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -79,3 +79,20 @@ bazel_dep(name = "buildifier_prebuilt", version = "7.3.1")
 #
 ###############################################################################
 bazel_dep(name = "aspect_rules_lint", version = "1.0.3")
+
+###############################################################################
+#
+# LLVM Toolchains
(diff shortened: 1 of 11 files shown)


ltekieli:
Does score repository require any toolchain config aside from python? IMHO not.

@nradakovic @AlexanderLanin maybe it would be better to provide a template repo with the toolchain setup? It's usually just a copy paste of the MODULE.bazel content.

Or maybe the toolchains should land in a separate repository.

nradakovic:
> Does score repository require any toolchain config aside from python? IMHO not.
> 
> @nradakovic @AlexanderLanin maybe it would be better to provide a template repo with the toolchain setup? It's usually just a copy paste of the MODULE.bazel content.
> 
> Or maybe the toolchains should land in a separate repository.

Well I would perfer that all toolchains are in separate repo, but IDK if we want to do this now?

Issue #406, opened 2025-02-17
GitHub PR Workflow


AlexanderLanin:
# How Should We Merge PRs in Score?

## Current State

Our [documentation](https://eclipse-score.github.io/score/main/process/guidance/contribution_request/index.html) states that a committer shall approve a PR, after which GitHub will add it to a merge queue. However, this is misleading:

- GitHub does not automatically add PRs to merge queue.
- We do not have a merge queue in place (which is a separate discussion).
- The actual steps required to merge a PR are not explicitly documented.

### Problem 1: The Merge Process is Unclear and Does Not Reflect Reality
- The documentation suggests an automated process that does not exist.
- The actual steps required to merge a PR are not clearly defined.
- There is ambiguity about whether we should be using a "merge commit," "rebase," or "squash" when merging.

### Problem 2: Managing "Fix Findings" Commits
To facilitate meaningful PR reviews, we must have commits that **only address review findings** (let’s call them "fix findings" commits for now). These commits should:
- Clearly isolate the changes that fix issues identified during review.
- Allow reviewers to verify that the requested changes have been applied.

However, this raises concerns:
- **GitLint rules**: Do these commits need to conform to our current commit message policies? Especially when we squash them anyway.
- **Git history cleanliness**: If every PR results in multiple "fix findings" commits, our history could become cluttered.

### 1. Clarify PR Merging Strategy
We should explicitly define how PRs should be merged. Options include:

1. **Squash & Merge**
   - Ensures a clean commit history (single commit per PR, -ff merge).
   - Avoids excessive intermediate commits in the main branch.
   - Reviewers can request changes through comments, and the final merge squashes all commits into one.
   - Results in problems when developers attempt to keep working on branches, which were squashed via the web interface.

2. **Rebase & Merge**
   - Preserves commit history.
   - More control over individual commits.
   - Requires a clear convention for handling review-related commits.

3. **Merge Commit**
   - Preserves full history but can lead to a messy commit graph.

4. **A custom approach**
   - Instead of using the web interface, we could merge upon e.g. a "/merge" comment in the PR.
   - While unconventional, this certainly allows us the most flexibility.

My personal recommendation:
* **Use "Squash & Merge"**, ensuring that the final commit message is well-structured and summarizes all review changes.
* Revisit this topic when we work on #205 

Would love to hear everyone's thoughts! 🚀

MaximilianSoerenPollak:
[...]

PR #424, opened 2025-02-19
docs: Update git guidance documentation.

johannes-esr:
Mention gitlint in the documentation and make clear that the body of a commit message is mandatory.


AlexanderLanin (approved):
That's a totally valid improvement, until we discuss #406

Issue #282, opened 2025-02-05
docs: Fix warnings in Docs Build


hoe-jo:
After implementing the checks and adapting the metamodel several warnings occurred concerning:
- Wrong UID
- Attributes renamed

Those warnings need to be fixed.

Currently following topics are open:

- Fix Check for optional_options, this code may be a starting point:

```python
    required_options: dict[str, str] = need_options.get("mandatory_options", {})
    optional_options: dict[str, str] = need_options.get("opt_opt", {})
    required_links: list[str] = [x[0] for x in need_options.get("req_link", ())]
    optional_links: list[str] = [x[0] for x in need_options.get("opt_link", ())]

    allowed_options = (
        list(required_options.keys())
        + list(optional_options.keys())
        + required_links
        + optional_links 
        + default_options_list
    )
``` 

- Stkh_Holder_Requirements + Tool Requirements -> fix id in files; activate check for 3 parts in ID again

PR #426, opened 2025-02-19
tooling: Correct upper case values in docs build errors.

Aymen-Soussi-01:
Replace all ids and options items by their value in small letters.

I used 2 regex to do the job (:[a-z]*: .*__.*) and (:[a-z]*: .*_.*)  and replaced them with \L$1 all using VS Code option to search and replace regex patterns to handle all options and id cases that not follow the pattern.
In another PR we will handle the case of id and options that start only with one "_" Plus there were some small mistakes like changing ASIL_D/B to asil_d/b and some small unnecessary ones that I fixed manually by returning them to their old value.
 
Related to: #282

### DIFF
--- a/docs/_tooling/extensions/score_metamodel/checks/id_format_and_length.py
+++ b/docs/_tooling/extensions/score_metamodel/checks/id_format_and_length.py
@@ -12,8 +12,8 @@
 # *******************************************************************************
 import os
 
-from sphinx_needs.data import NeedsInfoType
 from sphinx.application import Sphinx
+from sphinx_needs.data import NeedsInfoType
 
 from score_metamodel import CheckLogger, local_check
 
@@ -28,7 +28,7 @@ def check_id_format(app: Sphinx, need: NeedsInfoType, log: CheckLogger):
     # Split the string by underscores
     parts = need["id"].split("__")
 
-    if need["type"].startswith(("gd_", "wf_", "wp_")):
+    if need["id"].startswith(("gd_", "wf__", "wp__")):
         if len(parts) != 2 and len(parts) != 3:
             msg = "expected to consisting of one of these 2 formats:`<Req Type>__<Abbreviations>` or `<Req Type>__<Abbreviations>__<Architectural Element>`."
             log.warning_for_option(need, "id", msg)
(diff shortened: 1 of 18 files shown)


hoe-jo:
With the regexes you just proposed we loose the information where we need to set the first two underscores. It´s not clear for me how you want to check this later. My proposal would be:
1. ([A-Z]+_)+([A-Z]+__) -> \L$1\L$2

hoe-jo:
Just updated the regex with this we would fix the prefixes

Aymen-Soussi-01:
@hoe-jo  I manually reviewed every value I changed and didn't find any lost information. The 2 regex worked well and made the objective of PR. If you found any example of lost infos in this PR please mention it so that I can see how to correct it.
Also, I tried your regex and it doesn't give the expected results as it changes every single line of rst files and make most of the words if not all of them lost infos

hoe-jo:
Ok, let me please reformulate then:

Currently we can distinguish between prefix and keyword:
RL_contributor

If we change it to:
rl_contributor

How do we (later) introduce here the 2nd underscore? We already lost here the information what is prefix and what is keyword, other example:
STD_REQ_sometext -> std_req_sometext

To add here the 2nd (missing) underscore we would have to search for each workproduct itself later and correct it

With the Regex:
b(([A-Z0-9]+){1,4})*(?=[a-z]) -> \L$1_

This is done automatically in one step

Aymen-Soussi-01:
@hoe-jo the underscore handling will be in the next PR

AlexanderLanin:
We should use different need types for differentiation and not id-naming-rules.

AlexanderLanin (dismissed):

AlexanderLanin:
Mhh this PR does not actually reduce the number of warnings?!

Aymen-Soussi-01:
@AlexanderLanin We need to complete the underscore part to reduce the build errors as most of needs are not following the patters and not only are in big letters

AlexanderLanin:
But this PR does not reduce warnings. It's still ~1500 warnings, same as before.

Aymen-Soussi-01:
Yes before merging the main and resolving the conflicts somehow the errors where 1078 then they become >1800 after the merge and squach of the main I still dont get it as I resolve some small conflicts in 4 or 5 files which contains some valid ids and options

Aymen-Soussi-01:
@AlexanderLanin I just corrected one of the checks (check_id_format) and the errors now have been reduced to 1651 error

AlexanderLanin (approved):

PR #427, opened 2025-02-19
Adding functionality to link from source code to needs

### LINKED ISSUE #421: Improvement: Add linking of source code to needs
~~There shall be a way to link requirements to source code via a pre-determined template tag.~~

Edit: 

It should be the other way around, one should be able to link to requirements from source code via a template string.

The source code then should be linked inside the specified requirement under the 'Source Code Link' option.

MaximilianSoerenPollak:
Added functionality to link from source code to needs via tempalte strings. 

Improved speed of load_metamodel_data by reducing iterations
Added extra options into metamodel yaml to enable 'global' options to be defined easier.

closes #421 

--- 

Would welcome feedback on this, as some things probably aren't straight forward. 

Also would like to know (if that falls into this PR) where to put the rst TLDR user guide.


docs/_tooling/extensions/score_source_code_linker/README.md
```diff
@@ -0,0 +1,161 @@
+# Source Link Extension Details
```
MaximilianSoerenPollak, line 1: Documentation has not been updated.

docs/_tooling/extensions/score_metamodel/metamodel.yaml
```diff
@@ -1,5 +1,8 @@
 # yaml-language-server: $schema=./metamodel-schema.json
 
+needs_types_base_options:
+  optional_options:
+    source_code_link: "^https:\/\/github\.com\/eclipse-score\/score\/blob\/[a-zA-Z0-9]+\/[^#\s]+(?:#L\d+)?$"
```
MaximilianSoerenPollak, line 5: It matches what is expected better. 
Althought I would prefer for this to be much simpler if I'm honest, as longer / complicated regex aren't ideal. 

image

docs/_tooling/extensions/score_metamodel/metamodel.yaml
```diff
@@ -1,5 +1,8 @@
 # yaml-language-server: $schema=./metamodel-schema.json
 
+needs_types_base_options:
+  optional_options:
+    source_code_link: "^https://github.com/eclipse-score/score/blob/[a-zA-Z0-9]+/[^#\\s]+(?:#L\\d+)?$"
```
MaximilianSoerenPollak, line 5: Although I would prefer for this to be much simpler if I'm honest, as longer / complicated regex aren't ideal. But I have no solution at the moment for how to simplify except just allowing anything I guess.

image

docs/_tooling/extensions/score_source_code_linker/README.md
```diff
@@ -0,0 +1,161 @@
...
+   - Adds source_code_link option to matching needs
+   - Handles needs_string_links conversion
+
+3. Error Handling:
+   - Validates requirement IDs
+   - Provides descriptive warnings for missing IDs
+
+## Usage Guide
+
+> **This part will be deleted once moved to the correct RST file**
```
hoe-jo, line 73: Which would be the "correct" one?

docs/BUILD
```diff
@@ -43,6 +43,7 @@ load("@rules_pkg//pkg:mappings.bzl", "pkg_files")
 load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
 load("@rules_python//python:pip.bzl", "compile_pip_requirements")
 load("@rules_python//sphinxdocs:sphinx.bzl", "sphinx_build_binary", "sphinx_docs")
+load("//docs:_tooling/extensions/score_source_code_linker/requirement_links.bzl", "requirement_links")
```
AlexanderLanin, line 46: It's technically not specific to requirements... ;-)

docs/BUILD
```diff
@@ -68,6 +69,7 @@ sphinx_docs(
     config = ":conf.py",
     extra_opts = [
         "--keep-going",
+        "-Drequirement_links=$(location :requirement_links)",
```
AlexanderLanin, line 72:
```suggestion
        // do something...
        "-Drequirement_links=$(location :requirement_links)",
```

docs/BUILD
```diff
@@ -86,6 +89,14 @@ sphinx_build_binary(
     deps = sphinx_requirements,
 )
 
+requirement_links(
```
AlexanderLanin, line 92:
```suggestion
# Building/Using/Running this will...
requirement_links(
```

docs/BUILD
```diff
@@ -148,6 +160,12 @@ java_binary(
     ],
 )
 
+py_binary(
```
AlexanderLanin, line 163: Let's agree to put ~1 sentence next to each target?
Regardless of whether they are documented somewhere else.

docs/BUILD
```diff
@@ -188,3 +206,28 @@ filegroup(
     ],
     visibility = ["//visibility:public"],
 )
+
+py_library(
```
AlexanderLanin, line 210: Because when reading this, it's really hard to tell why there is a library and a binary which look rather identical.

docs/BUILD
```diff
@@ -188,3 +206,28 @@ filegroup(
...
+        ["_tooling/extensions/score_source_code_linker/**/*.py"],
+        exclude = ["_tooling/extensions/score_source_code_linker/tests/*.py"],
+    ),
+    imports = ["_tooling/extensions"],
+    visibility = ["//visibility:public"],
+)
+
+score_py_pytest(
+    name = "score_source_code_linker_test",
+    size = "small",
```
AlexanderLanin, line 222: What does size do?

docs/BUILD
```diff
@@ -188,3 +206,28 @@ filegroup(
...
+    srcs = glob(["_tooling/extensions/score_source_code_linker/tests/**/*.py"]),
+    args = [
+        "--basetemp /tmp/pytest",
```
AlexanderLanin, line 225:
```suggestion
       # set ... because ...
        "--basetemp /tmp/pytest",
```

docs/BUILD
```diff
@@ -188,3 +206,28 @@ filegroup(
...
+    ],
+    plugins = [
+        "sphinx.testing.fixtures",
```
AlexanderLanin, line 228: I've never explicitly enabled fixtures in pytest before. Are you sure this is needed?

docs/BUILD
```diff
@@ -188,3 +206,28 @@ filegroup(
...
+    ],
+    deps = [
+        ":score_source_code_linker",
+    ],
```
AlexanderLanin, line 232: Should we include the code under test via deps or via srcs? It would make sense to have a common pattern for our pytests. -> discussion

docs/_tooling/extensions/score_metamodel/metamodel.yaml
AlexanderLanin: We can try that, but why bother? This regex shall prevent people to somehow write something totally wrong in here.

`https://github.com/eclipse-score/.+` is not "totally wrong".

docs/BUILD
AlexanderLanin: Maybe make an explicit section and add a reference to docs/_tooling/extensions/score_source_code_linker/README.md?

docs/_tooling/extensions/score_source_code_linker/README.md
```diff
@@ -0,0 +1,161 @@
+# Source Link Extension Details
+
+A Sphinx extension for source code traceability for requirements. This extension works with the Bazel system and Sphinx-needs to provide automatic source code traceability.
+In a first step it parses the source code for requirement tags. All discovered tags including their file and line numbers are written in an intermediary file before the sphinx build.
+In a second step this intermediary file is parsed during sphinx build. If a requirement Id is found in the intermediary file a link to the source is added.
+
+## Implementation Components
+
+### Bazel Integration
+The extension uses two main components to integrate with bazel:
```
AlexanderLanin, line 10: Maybe we should add a dataflow diagram here?

docs/_tooling/extensions/score_source_code_linker/README.md
```diff
@@ -0,0 +1,161 @@
...
+   - Processes all files from provided deps
+   - Passes files as `--input` arguments to `requirement_links.py`
+   - Handles dependency tracking for incremental builds
+
+2. `requirement_links.py`
+   - Scans input files for template tags (e.g., `# req-traceability:`)
+   - Retrieves git information (hash, file location)
+   - Generates mapping file with requirement IDs and links
+
+**Note:** The extension also works when you use `bazel run //docs:incremental`
```
AlexanderLanin, line 22: Why is that Noteworthy? :D

docs/BUILD
MaximilianSoerenPollak: That is actually true. This probably should be a rename and called 'file_links'

docs/BUILD
MaximilianSoerenPollak: I don't quite understand this. Do you mean the `-D` argument has to be the last one in the list, and to indicate this via the comment?

docs/BUILD
MaximilianSoerenPollak: > Hopefully work? 🤞 

Will adapt this to give some context, as well as to the other py-library I have added by necessity

docs/BUILD
MaximilianSoerenPollak: It was needed in the old implimentation, but since we do the requirements differently here this most likely is no longer needed. 
I will test this and remove if doable.

docs/BUILD
MaximilianSoerenPollak: The `srcs` in score_py_pytest can only be pytest / test files if I rembmer correctly, therefore if you need 'other' sources for your tests (like the py_library so it can import) you have to do it via deps. 

BUt will also take a look at this again and check if I rembmer correctly here.

docs/_tooling/extensions/score_source_code_linker/README.md
MaximilianSoerenPollak: It wasn't a given in the past :D . This should be removed as 'incremental' now should work 100% same as build.

docs/BUILD
MaximilianSoerenPollak: Size is a parameter that allows you to 'filter' the tests in bazel as well. You basically can tell bazel test all 'small' tests for example. 

Unsure if this is something we want to incooperate, as it's also hard to define when something is 'small, medium, big' 
https://bazel.build/reference/test-encyclopedia#role-test-runner for reference.

docs/BUILD
MaximilianSoerenPollak: I think this makes sense. Have each target explained or at least referenced where it belongs to will make the BUILD more approachable in the long run. 

Maybe also 'grouping' different things together makes some sense to get an easier overivew.

docs/_tooling/extensions/score_source_code_linker/README.md
MaximilianSoerenPollak: That might make it easier to follow. Have never really done that but I take a look into it and add it.

docs/BUILD
AlexanderLanin: No, sorry. My intention was for you to document what -D does. I've now googled it :-)

Can we use `--define` instead of  `-D`? That would already make it clearer

docs/BUILD
AlexanderLanin: I see. Let's move this to a separate discussion. Google has a definition for 'small, medium, big' which we could reuse/adapt. I personally like this a lot more than "unit test, integration test, system test" since I typically don't care what kind it is. I care how fast it is.

docs/BUILD
MaximilianSoerenPollak: I have to test if this works, as the `-D` parameter is from sphinx itself, not from bazel. But yes that would make it much clearer for sure.

docs/BUILD
MaximilianSoerenPollak: I have renamed this much more explicit, wondering though if it now might be too explicit / long, as I don't want to start introducing Akronyms to make it shorter. Once the commit is up, happy to hear your feedback.

docs/BUILD
MaximilianSoerenPollak: Added references to the README and more comments to explain.

docs/_tooling/extensions/score_source_code_linker/README.md
```diff
@@ -0,0 +1,159 @@
+# Source Link Extension Details
+
+A Sphinx extension for source code traceability for requirements. This extension works with the Bazel system and Sphinx-needs to provide automatic source code traceability.
+In a first step it parses the source code for requirement tags. All discovered tags including their file and line numbers are written in an intermediary file before the sphinx build.
+In a second step this intermediary file is parsed during sphinx build. If a requirement Id is found in the intermediary file a link to the source is added.
```
AlexanderLanin, line 5: Would it make sense to use needextend for that?

AlexanderLanin (changes requested):

AlexanderLanin:
@MaximilianSoerenPollak some old findings in here that still apply?

docs/_tooling/extensions/score_source_code_linker/README.md
MaximilianSoerenPollak: I can't quite follow, in which way do you mean?

AlexanderLanin: "In a second step this intermediary file is parsed during sphinx build. If a requirement Id is found in the intermediary file a link to the source is added." = needextend

MaximilianSoerenPollak: Needextend needs to be inside the rst file, as far as I know. 
So we would need to change the underlying rst in order to get this right, which is something we didn't want to do in a 'build' step. 
Also I'm not sure if the needextend is getting parsed before or after the strings are converted to links.

And you would need a 'needextend' for each need that is referenced as they all have a different link.

Seems not ideal, unless I'm missunderstanding.

MaximilianSoerenPollak:
> @MaximilianSoerenPollak some old findings in here that still apply?

Yes I didn't yet add the Dataflow diagram not quite sure what a good approach for that would be.
Also didn't yet remove the note of it working with incremental.

MaximilianSoerenPollak:
@hoe-jo there are still links to 'old' requirements that do not exist yet in score. Should I just remove these for now?

docs/_tooling/extensions/score_source_code_linker/README.md
MaximilianSoerenPollak: Would this be something like you had in mind? Or at least in the direction? 
image

AlexanderLanin: Perfect! Can you add test to the arrows between the boxes? e.g. from file discover to tag processing is potentially a "list of files".

MaximilianSoerenPollak: Should I put that inside an 'asset' folder in there? Or how do we best display this? I could put it as 'UML' code but then you won't see if you read the README ?

AlexanderLanin: For README.md it's best to store the png... but how do we keep it editable? PlantUML embedds the code into png giles. But no-one will know that... Still the best choice?!

docs/_tooling/extensions/score_source_code_linker/parse_source_files.py
```diff
@@ -0,0 +1,123 @@
...
+import collections
+import functools
+import json
+from typing import Optional
+from pathlib import Path
+
+
+TAGS = [
+    "# req-traceability:",
+    "# req-Id:",
```
AlexanderLanin, line 25: Should be discussed after PR is merged. `#` sounds rather python specific. The space is strange. etc.

docs/_tooling/extensions/score_source_code_linker/parse_source_files.py
```diff
@@ -0,0 +1,123 @@
...
+]
+
+GITHUB_BASE_URL = "https://github.com/eclipse-score/score/blob/"
+
+
+def extract_requirements(
+    source_file: str, git_hash_func: Optional[callable] = None
```
AlexanderLanin, line 32:
```suggestion
    source_file: str, git_hash_func: Optional[callable] = get_git_hash
```

If that's the default, can't  we set it as a default?

docs/_tooling/extensions/score_source_code_linker/parse_source_files.py
```diff
@@ -0,0 +1,123 @@
...
+    get_hash = git_hash_func if git_hash_func is not None else get_git_hash
+    with open(source_file, "r") as f:
+        for line_number, line in enumerate(f):
+            line_number = line_number + 1
+            line = line.strip()
+            if any(x in line for x in TAGS):
+                hash = get_hash(source_file)
+                cleaned_line = (
+                    line.replace("'", "").replace('"', "").replace(",", "").strip()
+                )
```
AlexanderLanin, line 65: This will not work for `According to # req-Id: REQ_123 we need to...`

Proposal: merge as-is, and address later.

docs/_tooling/extensions/score_source_code_linker/parse_source_files.py
```diff
@@ -0,0 +1,123 @@
...
+    Returns:
+        (str): Full 40char length githash of the latest commit this file was changed.
+
+        Example:
+                3b3397ebc2777f47b1ae5258afc4d738095adb83
+    """
+    try:
+        abs_path = Path(file_path).resolve()
+        # HACK: could not think of a better way to get the root directory.
+        score_path = abs_path.__str__().split("score")[0] + "score"
```
AlexanderLanin, line 90: We have different approaches throughout the code base. We should align how we do that.

Proposal: merge as-is, and address later.

AlexanderLanin: Note: here you can also simply use `pathlib.Path(abs_path).parent` as you don't really need the git root.

AlexanderLanin (approved):
Only minor findings. Fix or merge?

docs/_tooling/extensions/score_source_code_linker/README.md
AlexanderLanin: It's still rendered on github and in vs code preview. Both valid approaches to reading markdown.

I feel you can just store the png. And later we describe somewhere how to edit such pngs.

docs/_tooling/extensions/score_source_code_linker/parse_source_files.py
MaximilianSoerenPollak: Ah, you mean as have the `tempalte string` inside normal text or comments, instead of on it's own line? 
Hmm that's true.

MaximilianSoerenPollak:
> Only minor findings. Fix or merge?

I think fixing all findings except [this one](https://github.com/eclipse-score/score/pull/427#discussion_r1971307064) makes sense, as that one I fear will need more time & energy to figure out correctly and isn't a quick fix.

docs/_tooling/extensions/score_source_code_linker/parse_source_files.py
```diff
@@ -28,8 +28,43 @@
...
+
+    Returns:
+        (str): Full 40char length githash of the latest commit this file was changed.
+
+        Example:
+                3b3397ebc2777f47b1ae5258afc4d738095adb83
+    """
+    try:
+        abs_path = Path(file_path).resolve()
+        # HACK: could not think of a better way to get the root directory.
```
MaximilianSoerenPollak, line 46: remove comment, change var name

AlexanderLanin (approved):

PR #438, opened 2025-02-21
docs: bugfix uids metamodel

hoe-jo:
This PR fixes another bunch of uids


aschemmel-tech (approved):
Only format changes, ok

PR #446, opened 2025-02-24
docs: fix uids

hoe-jo:
fix second bunch of ids
- deactivate check for opt_opts
- deactivate check for 3 parts for "stkh_req__", "tool_req__", "doc__"
- increase length check to 45 chars

remaining topics:
- safety plan needs more fixing


aschemmel-tech (approved):
checked ok in 4-eye

PR #449, opened 2025-02-24
docs: initial draft for python debugging in bazel

hoe-jo:
- added module to include debugpy
- added vscode launch config


hoe-jo:
Currently not needed for docs build

Issue #348, opened 2025-02-12
OSS/Dash: Enable automatic clearance


AlexanderLanin:
We already have a job running (see #155), but currently the output (see #347) is only an info. There is no way to "fix" the problems.

We've already got a private_key for our tooling to auto-create eclipse-tickets for whitelisting the dependencies that dash is currently reporting.

Current idea:
- Extend #347 with something like "comment 'XYZ' to auto create whitelisting ticket".
- on-comment (given sufficient access rights, e.g. contributor), run dash in "auto-create-ticket-mode"
- Create new comment

Idea two:
- Make the tool runnable locally

Token:
> A GitLab bot user for the eclipse-score project (username: score-bot) and an API token have been created.
The API token has been added to eclipse-score organization as organization secret (ID: ECLIPSE_GITLAB_API_TOKEN)

Reference: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5590
Reference: https://github.com/eclipse-score/.eclipsefdn/commit/d4adfcdf3dfc896a939c60e2fe278fe06ff37440

Example:
* https://github.com/orgs/eclipse-kuksa/repositories?q=lang%3Arust&type=all
* https://github.com/eclipse-kuksa/kuksa-actions

dcalavrezo-qorix:
We need to find out h w Dash tracks submitted tickets

- does Dash cache or track already submitted requests, preventing duplicate submissions?
- do we have to check for existing tickets

We don't want to flood them with too many requests.

AlexanderLanin:
@SebastianSchildt any idea?

SebastianSchildt:
This is no issue in our experience. The first time dash finds a dependency not yet in the curation db, it creates a ticket.
The next times, weil the ticket exists, but is not cleared yet (e.g. due to more pushes to a PR) dash will try to create ticket but it will detect that there is already one (it will also tell you so in its output), and not create another one.

So basically you do not need to take care of this yourself. Just let dash fire away

dcalavrezo-qorix:
> This is no issue in our experience. The first time dash finds a dependency not yet in the curation db, it creates a ticket. The next times, weil the ticket exists, but is not cleared yet (e.g. due to more pushes to a PR) dash will try to create ticket but it will detect that there is already one (it will also tell you so in its output), and not create another one.
> 
> So basically you do not need to take care of this yourself. Just let dash fire away

Perfect! Thank you @SebastianSchildt

PR #518, opened 2025-02-26
dash: enable automatic license clearance via review action

dcalavrezo-qorix:
Enabled the execution of `review` for the dash tool in the CI pipeline. This won't work on local development unless developers have their own DASH PATs. Important notice:  secrets remain in the CI workflow, not in Bazel BUILD files, they won't be exposed.

Fixes: #348

### DIFF
--- a/tools/dash/dash.bzl
+++ b/tools/dash/dash.bzl
@@ -47,7 +47,17 @@ def dash_license_checker(
         runtime_deps = [
             "@dash_license_tool//jar",
         ],
-        args = ["$(location :{}2dash)".format(name)],
+        # We'll build up "args" in the order: [ static options ] + select() + [ file last ]
+        args = [
+            # If we have any always-on flags, put them here
+        ] + select({
+            # If CI_BUILD=true, add "-review"
+            ":ci_build": ["-review"],
+            "//conditions:default": [],
+        }) + [
+            # The file is last
+            "$(location :{}2dash)".format(name),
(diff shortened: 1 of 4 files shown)


tools/dash/dash.bzl
```diff
@@ -47,7 +47,17 @@ def dash_license_checker(
         runtime_deps = [
             "@dash_license_tool//jar",
         ],
-        args = ["$(location :{}2dash)".format(name)],
+        # We'll build up "args" in the order: [ static options ] + select() + [ file last ]
+        args = [
+            # If we have any always-on flags, put them here
+        ] + select({
+            # If CI_BUILD=true, add "-review"
```
AlexanderLanin, line 54: I need to think about:
* Using "CI" from env, instead of a custom flag
* Using PERFORM_CLEARANCE instead of generic CI_BUILD
* Pass `-review` same as the other parameters, without select

AlexanderLanin: Third options sounds most reasonable? Especially since the 3 other parameters are passed directly anyway.

AlexanderLanin: I think I would have simply used -review. But that's only my gut feeling. So let's keep it as is. 
Works for now!
Doesn't matter whether we adjust it in the PR or later on demand - if there is ever such a demand.

AlexanderLanin (approved):

Issue #243, opened 2025-01-28
python: code formatting / ruff linting


AlexanderLanin:
Scope: 
* Provide ruff configuration for formatting based on common python guidelines
* Provide ruff configuration for linting based on common python guidelines
* Enable ruff extension?
* Ensure formatting behaves the same with the current `bazel test //:format.check` and `bazel run //:format.fix` commands.
* Format existing code

Out of Scope:
* Add to bazel, that will be handled by #506
* Fix linter warnings

Issue #508, opened 2025-02-25
python: provide pyright with initial config


AlexanderLanin:
Scope: 
* Provide pyright configuration based on common python guidelines
* Enable pyright extension?

Out of Scope:
* Add pyright to bazel, that will be handled by #506
* Fixing existing warnings
* When to write data types?
* When to use custom types?

PR #528, opened 2025-02-27
Python linting

MaximilianSoerenPollak:
Add initial python linting configuration


AlexanderLanin (approved):
Closes #243, #508 

Note: lots of out-of-scope topics
* Fixing existing warnings (obvious errors visible!)
* When to write data types?
* When to use custom types?

PR #543, opened 2025-02-28
dash: support for Rust

dcalavrezo-qorix:
Added support for RUST dependencies check via DASH

fixes # 349

### DIFF
--- a/tools/testing/pytest/tests/test_dash_format_converter.py
+++ b/tools/testing/pytest/tests/test_dash_format_converter.py
@@ -1,0 +1,115 @@
+# *******************************************************************************
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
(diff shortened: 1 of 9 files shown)


tools/testing/pytest/tests/test_dash_format_converter.py
```diff
@@ -0,0 +1,115 @@
...
+
+@pytest.fixture(scope="session")
+def converter_path():
+    """
+    Determines the correct way to call dash_format_converter.
+
+    - If running inside Bazel, use TEST_SRCDIR to find the `py_binary` target.
+    - If running manually, call the Python script directly.
+    """
+    if "TEST_SRCDIR" in os.environ:
```
AlexanderLanin, line 38: Doesn't matter much, but somehow I feel it would be nice if all our tools follow the same approach.
Whether we want to test via subprocess or via call to main().

tools/testing/pytest/tests/test_dash_format_converter.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # Invoke the converter with -t cargo
+    subprocess.run(
+        f"{converter_path} -i {input_file} -o {output_file} -t cargo",
+        shell=True,
+        check=True,
+    )
+
+    # Read the output and verify correctness
+    actual = output_file.read_text().splitlines()
```
AlexanderLanin, line 110: Suggestion: wrap it into something like
```suggestion
    actual = run_converter(cargo_lock_content, "-t cargo")
```

AlexanderLanin (approved):

PR #548, opened 2025-03-03
docs: add graph checks

### LINKED ISSUE #527: docs: implement general graph checks scheme
Currently all graph checks are implemented as a 1:1 solution in python.

Here it would be very helpful to have a more generalized approach which uses the already existing metamodel.yaml

A proposal for that could be that we add here a section which follows some scheme like:

```
graph checks
  safety_check:
    apply:
      include: "comp_req"
      condition:
        "safety != QM"
    check:
      relation: "satisfies"
      condition:
        and:
          - "safety != QM"
          - "status == valid"
```

hoe-jo:
- add general graph checks
- added inition check in metamodel

resolves #527


docs/_tooling/extensions/score_metamodel/metamodel.yaml
```diff
@@ -396,3 +396,29 @@ needs_extra_links:
...
+#   - condition: defines (together with apply) the condition which the needs need to fulfill
+#     - [and / or / xor / not]
+# - check: defines the check that should be applied
+#   - relation: defines the relation which is used derive the type of the parent need
+#   - condition: defines the condition that should be checked
+#     - [and / or / xor / not]
+##############################################################
+graph_checks:
+  safety_check:
+    apply:
```
AlexanderLanin, line 416:
```suggestion
    when:
```
Is this the when clause?

AlexanderLanin: It's the same with the other parameters I'm afraid. They are not as self explanatory as you hoped ;-)

docs/_tooling/extensions/score_metamodel/checks/graph_checks.py
```diff
@@ -0,0 +1,139 @@
...
+        if (re.match(pattern, need["type"])) and (
+            eval_need_condition(need, selection["condition"])
+        ):
+            selected_needs.append(need)
+
+    return selected_needs
+
+
+@graph_check
+def check_metamodel_graph(
```
AlexanderLanin, line 112: Can you add some tests here? More for examples and readability, than for verification.

AlexanderLanin (changes requested):

docs/_tooling/extensions/score_metamodel/checks/graph_checks.py
hoe-jo: #553 created

docs/_tooling/extensions/score_metamodel/metamodel.yaml
hoe-jo: Maybe we can change the yaml to:

```
needs:
  include: comp_req
  condition: safety != QM
check:
  and:
    satisfies: safety != QM
    satisfies:  status == valid 
```

AlexanderLanin: This does not state that there must an satisfies link at all.

hoe-jo: This is checked by the mandatory attributes check

AlexanderLanin: Mhh, interesting. So we have split the checks for the same topic across different technological aspects. That's a discussion for the future.

AlexanderLanin:
Another point for the future: how do we add readable error messages to that yaml code.

Suggestion: fix copyright issues and we merge this as-is.

AlexanderLanin (approved):

PR #550, opened 2025-03-03
tooling: Integrate standards charts and Correct tests for checks

Aymen-Soussi-01:
Integrate the standards charts in the documentation and correct format of the tests for checks as import of metamodel.


docs/_tooling/extensions/score_metamodel/checks/standards.py
```diff
@@ -27,7 +27,14 @@ def get_standards_needs(needs: list[NeedsInfoType]) -> dict:
     return {
```
AlexanderLanin, line 27: we have a dedicated type, don't we? Something like `.. standard-req::`. So why do we need to differentiate by id at all?

docs/_tooling/extensions/score_metamodel/tests/test_check_options.py
```diff
@@ -131,9 +134,11 @@ def test_unknown_option_present_in_req_opt(self):
         )
 
         logger = fake_check_logger()
-
+        app = Mock(spec=Sphinx)
+        app.config = Mock()
+        app.config.needs_types = self.NEED_TYPE_INFO
```
AlexanderLanin, line 139: We need a better testing interface. But that's out of scope of this PR.

docs/process/standards/index.rst
```diff
@@ -22,3 +22,22 @@ Standards
    iso26262/iso26262
    isopas8926/isopas8926
    aspice_40/index
+
+.. needpie:: Percentage of standard requirements which are linked to at least one item
+   :labels: Connected Requirements, Not Connected Requirements
+   :legend:
+   :colors: green, red
+   :filter-func: _tooling.extensions.score_metamodel.checks.standards.my_pie_linked_standard_requirements
```
AlexanderLanin, line 30: Sound like this should not be in score_metamodel. Definitely not in checks.

AlexanderLanin (approved):
I think all 3 findings need to be followed up on. But let's merge this as-is, as it is an improvement as-is.

Issue #566, opened 2025-03-05
bazel: add workflow for bazel tests


hoe-jo:
It is required to setup a workflow for bazel tests. For the start it should include the test targets defined in /docs

PR #569, opened 2025-03-05
ci: add bazel test workflow

hoe-jo:
added workflow to run the tests in the docs environment

fixes: #566

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,26 @@
+# *******************************************************************************
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+name: Run Bazel Tests
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+jobs:
+  code:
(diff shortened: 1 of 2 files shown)


hoe-jo:
@MaximilianSoerenPollak Can you please take a look at the tests for source code linker, those are currently failing, thats why we cannot merge the bazel test job:
//docs:score_source_code_linker_test

MaximilianSoerenPollak:
> @MaximilianSoerenPollak Can you please take a look at the tests for source code linker, those are currently failing, thats why we cannot merge the bazel test job: //docs:score_source_code_linker_test

Well that's interesting. 
The test got broken by a formatter in this [commit](https://github.com/eclipse-score/score/commit/b0f81ce9f3b5f9926df6f9275bac4364e913b2cc#diff-57039746719e644fc63a6f5688e49b5750a671b695bfa5589f56f05985e588a5R158-R159)

I think there I just have to change the warning that is given out then it is correct as I think now the formatting is correct. 
How should I fix this, in another PR? @hoe-jo

.github/workflows/test.yml
```diff
@@ -0,0 +1,26 @@
...
+    types: [opened, reopened, synchronize]
+jobs:
+  code:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4.2.2
+      - name: Run test targets
+        run: |
+          bazel test //docs/...
```
MaximilianSoerenPollak, line 26: Shoudl this just be `//docs/...` or should it be `/...` to test everything?

hoe-jo: for now I would start with //docs there we have working test cases

I would go the incremental approach and add targets as soon as they are ready

ltekieli (approved):

Issue #415, opened 2025-02-18
docs: not all plantuml types supported


AlexanderLanin:
Currently PlantUML works fine in docs. But not all diagram types are working.
Reference: https://github.com/eclipse-score/score/pull/374#issuecomment-2665206780

The reason is, that the diagram types we tested so far do not require dot.
But others rely on dot.
Reference: https://chatgpt.com/share/67b46957-d804-800a-aedd-b148068caab9

Task:
* Add example diagram
* Ensure that it is working
* Hint: add dot via bazel

MaximilianSoerenPollak:
A bit weird (in the GPT summary) as the diagramms in the UML functions #414 also create class diagramms, but I never gotten an Error. 

I guess easiest would be to just add it via bazel, so that all diagramms no matter which engine is used, and also legaycy diagramms, are supported.

dcalavrezo-qorix:
No archive from the releases contain the `linux` binaries

In theory we should have the following bazel binary linked which we could pass to the planuml jar when invoked

```
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "graphviz",
+    urls = ["https://gitlab.com/api/v4/projects/4207231/packages/generic/graphviz-releases/10.0.1/graphviz-10.0.1.tar.gz"],
+    strip_prefix = "graphviz-10.0.1",
+    build_file_content = """
+package(default_visibility = ["//visibility:public"])
+
+sh_binary(
+    name = "dot_bin",
+    srcs = ["bin/dot"],
+    visibility = ["//visibility:public"],
+)
+    """,
+)
```

AlexanderLanin:
@dcalavrezo-qorix Have you already tried it? Does graphviz/dot have dependencies itself?

@ltekieli Can you please have a look above to see whether that's the right approach? And the correct goal? Usually, graphviz/dot is installed via apt. See also https://gitlab.com/graphviz/graphviz/-/releases

ltekieli:
@AlexanderLanin I'd start with adding the installation with apt in GH actions, and after that explore how this could be added via bazel.

AlexanderLanin:
That's quite true, we could make a hotfix in `score`. Here we are already looking into how to do that properly with bazel.

AlexanderLanin:
> In theory we should have the following bazel binary linked which we could pass to the planuml jar when invoked

[...]

PR #619, opened 2025-03-10
dot: install graphivz

dcalavrezo-qorix:
Install graphviz in the CI workflow and in the devcontainer

Fixes: #415


AlexanderLanin:
Looks good, but doesn't quite close the issue. We need at least documentation. And potentially some sort of detection / warning.

PR #639, opened 2025-03-11
process: add process area overview

masc2023:
Add image with process areas showing maturity level


docs/process/index.rst
```diff
@@ -34,6 +34,13 @@ Process Model
 The SCORE process model is described by workflows, executed by roles and is creating work products.
 Workflow activities are supported by guidances.
 
+.. figure:: _assets/score_process_area_overview.drawio.svg
+  :width: 100%
+  :align: center
+  :alt: Process area overview for **S-CORE** platform
+
+  Process area overview for **S-CORE** platform
+
```
aschemmel-tech, line 43: I think we should add an explanation what the Maturity Levels mean. And if these correspond to some document status?

aschemmel-tech (changes requested):

docs/process/index.rst
masc2023: Initial definition added

aschemmel-tech (approved):
Fine now

PR #643, opened 2025-03-12
ide: Ignore CLion IDE Aspect directory

LittleHuba:
This directory is created by the bazel plugin, and should not persist over different users.


hoe-jo (approved):
LGTM

PR #646, opened 2025-03-12
infra: Add starpls as bazel LSP

nicu1989:
Add bazel LSP via bazel in vscode.


AlexanderLanin (approved):

AlexanderLanin:
For the record: starpls is listed as dual licensed under Apache-2.0 and MIT

PR #595, opened 2025-03-06
Docs bzl rules

MaximilianSoerenPollak:
This is a WIP PR that should demonstrate and start the moving and abstracting of bazel rules and targets in order to make  the multi-repo easier and further move us towards a viable template.

Developing this in Collaboration with @AlexanderLanin

PR #653, opened 2025-03-12
docs: bunch of random fixes

AlexanderLanin:
Collection small, localized fixes to make #595 smaller.


docs/_tooling/extensions/score_plantuml.py
```diff
@@ -85,23 +52,25 @@ def get_runfiles_dir() -> Path:
...
+                sys.exit("Could not find git root.")
 
         runfiles_dir = git_root / "bazel-bin" / "docs" / "ide_support.runfiles"
-        if not runfiles_dir.exists():
-            sys.exit(
-                f"Could not find ide_support.runfiles at {runfiles_dir}. "
-                "Have a look at README.md for instructions on how to build docs."
-            )
+
+    if not runfiles_dir.exists():
```
MaximilianSoerenPollak, line 63: I think just for future consitency and concepts I like an assert here more. 
Same comment etc, but since we use asserts in other places already this would make it consitent that we do 'worldview' checks via them.

docs/_tooling/extensions/score_source_code_linker/__init__.py
```diff
@@ -23,50 +23,17 @@
...
-            {
-                "version": "0.1",
-                "parallel_read_safe": True,
-                "parallel_write_safe": True,
-            }
-    """
-    app.connect("env-updated", add_source_link)
     app.add_config_value("source_code_linker_file", "", rebuild="env")
+    if app.config.source_code_linker_file:
+        LOGGER.info("Loading source code linker...")
```
MaximilianSoerenPollak, line 28: Question: 
Probably should add the type here? Although I think we only need it in warnings.

MaximilianSoerenPollak (commented):
Other than the assert thing, this looks good 👍

docs/_tooling/extensions/score_plantuml.py
AlexanderLanin: So far I've mostly used sys.exit for such cases. Everywhere, not only here. The reason is simply that assert statements provide rather unreadable output.

docs/_tooling/extensions/score_source_code_linker/__init__.py
AlexanderLanin: What type?

MaximilianSoerenPollak (approved):
LGTM

Issue #174, opened 2025-01-16
docs-bug: warnings via esbonio have incorrect paths


AlexanderLanin:
File paths of warnings are missing the `docs/` prefix:

![Image](https://github.com/user-attachments/assets/e69ae0ce-c20b-4195-becc-19bba926038a)

Effect:
* they do not show up via error lens
* they are not clickable in the output or the problems tab of VS Code

MaximilianSoerenPollak:
Guessing this is due to the `docs/` we removed some time ago.

MaximilianSoerenPollak:
#572  is a proposal on how to deal with this. 
Looking for feedback currently, as we think this 'manual' upkeep isn't great, but also could not think of another solution that works, as the `docs` folder is completely swallowed by sphinx/sphinx-needs itself in the 'needs' objects, therefore can not access it either as `Path.parent` or otherwise.

AlexanderLanin:
With #359 fixed: Does the same problem appear in esbonio 1.0?

MaximilianSoerenPollak:
> With [#359](https://github.com/eclipse-score/score/issues/359) fixed: Does the same problem appear in esbonio 1.0?

I did not test that yet, can take a look at it :)

AlexanderLanin:
> could not think of another solution that works, as the `docs` folder is completely swallowed by sphinx/sphinx-needs itself in the 'needs' objects, therefore can not access it either as `Path.parent` or otherwise.

```py
if running_in_esbonio:
  # prefix `docs/` since esbonio does not account for WORKDIR vs esbonio.sphinx.srcDir
  path_of_warning = "docs/" + path_of_warning
```

nicu1989:
How do we form the variable `running_in_esbonio` ? I`m thinking the other way around: check if running from bazel with:

```
            if 'RUNFILES_DIR' in os.environ or 'RUNFILES_MANIFEST_FILE' in os.environ:
                matching_file = f"{need['docname']}{need['doctype']}"
            else:
                current_directory = Path.cwd()
                matching_file = next(current_directory.rglob(f"{need['docname']}{need['doctype']}"))

            return f"{matching_file}:{need['lineno']}"
```

I tested this for:

1. $ bazel run //docs:incremental
`process/workflows/architecture_design.rst:24: WARNING: wf__cr_mt_featarch: is missing required option: `status`.`
2. $ bazel build //docs:docs
[...]

PR #655, opened 2025-03-13
docs: Fix warnings via esbonio

nicu1989:
# Bugfix

## Description

Esbonio warnings have correct file path while keeping a correct file path for bazel build and run.

## Related ticket

closes [#174] (bugfix ticket)

### DIFF
--- a/docs/_tooling/extensions/score_metamodel/__init__.py
+++ b/docs/_tooling/extensions/score_metamodel/__init__.py
@@ -64,7 +64,7 @@ def _run_checks(app: Sphinx, exception: Exception | None) -> None:
 
     logger.debug(f"Running checks for {len(needs_all_needs)} needs")
 
-    log = CheckLogger(logger)
+    log = CheckLogger(logger, app)
 
     # Need-Local checks: checks which can be checked file-local, without a
     # graph of other needs.
--- a/docs/_tooling/extensions/score_metamodel/__init__.py
+++ b/docs/_tooling/extensions/score_metamodel/__init__.py
@@ -64,7 +64,9 @@ def _run_checks(app: Sphinx, exception: Exception | None) -> None:
 
     logger.debug(f"Running checks for {len(needs_all_needs)} needs")
 
-    log = CheckLogger(logger, app)
+    prefix = Path(app.srcdir).relative_to(Path.cwd())
+
+    log = CheckLogger(logger, prefix)
 
     # Need-Local checks: checks which can be checked file-local, without a
     # graph of other needs.
--- a/docs/_tooling/extensions/score_metamodel/log.py
+++ b/docs/_tooling/extensions/score_metamodel/log.py
@@ -10,34 +10,47 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
+import os
+from pathlib import Path
+
+from sphinx.application import Sphinx
 from sphinx_needs.data import NeedsInfoType
 from sphinx_needs.logging import SphinxLoggerAdapter
 
 
 class CheckLogger:
-    def __init__(self, log: SphinxLoggerAdapter):
+    def __init__(self, log: SphinxLoggerAdapter, app: Sphinx):
         self._log = log
         self._count = 0
+        self._app = app
 
     @staticmethod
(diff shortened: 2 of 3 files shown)


docs/_tooling/extensions/score_metamodel/log.py
```diff
@@ -10,34 +10,47 @@
...
 
         if get("docname") and get("doctype") and get("lineno"):
             # Note: passing the location as a string allows us to use readable relative paths,
             # passing as a tuple results in absolute paths to ~/.cache/.../bazel-out/..
-            return f"{need['docname']}{need['doctype']}:{need['lineno']}"
+            if "RUNFILES_DIR" in os.environ or "RUNFILES_MANIFEST_FILE" in os.environ:
+                matching_file = f"{need['docname']}{need['doctype']}"
+            else:
+                prefix = Path(app.srcdir).relative_to(Path.cwd())
+                matching_file = f"{prefix}/{need['docname']}{need['doctype']}"
```
AlexanderLanin, line 39: I would prefer this logic outside of CheckLogger, as introducing a dependency to "app" is a huge dependency?!

docs/_tooling/extensions/score_metamodel/__init__.py
```diff
@@ -64,7 +64,9 @@ def _run_checks(app: Sphinx, exception: Exception | None) -> None:
 
     logger.debug(f"Running checks for {len(needs_all_needs)} needs")
 
-    log = CheckLogger(logger, app)
+    prefix = Path(app.srcdir).relative_to(Path.cwd())
```
AlexanderLanin, line 67: is ` if "RUNFILES_DIR"` no longer needed?

AlexanderLanin: Ah found it. Wouldn't it be clearer to have that if here? As the prefix is only applies in some cases.

AlexanderLanin: Not sure that counts as readable, but basically:
```suggestion
    prefix = Path(app.srcdir).relative_to(Path.cwd()) if "RUNFILES_DIR" in os.environ or "RUNFILES_MANIFEST_FILE" in os.environ else ""
```

nicu1989: `if "RUNFILES_DIR"` is logic belonging to CheckLogger class - how to handle log warning paths. Moving it outside, to score_metamodel/__init__.py will also have no effect to CheckLogger constructor signature.

AlexanderLanin: Mhh, from a dependency perspective it should be outside. But from a logical point of view, we should indeed pass the app, or at least the srcdir.
I don't like low level classes depending on huge objects or external environment values.

So it's logical grouping vs design.

Not sure.

@MaximilianSoerenPollak any idea?

MaximilianSoerenPollak: Hmm I don´t like copying 'app' into the Class, it's such a big object, copying isn´t ideal. We only call the thing once currently, so it might not be a big deal at the moment though. 

Having the logic inside the checker though feels a bit weird just from a perspective of how the code looks, not saying it doesn´t belong there as currently I can´t think of another place where we would need this. 
Having the class depended on environment variables especially  inside the class itself can make it harder to see the scope / dependency.
I would prefer to know it's depended on it, but then again it's not like it's hidden very deep inside the class or will have any user facing interaction so. 


So I guess to sum up all the non talking I just did:
I think all approaches can work here, I'd be fine with either or but would personally prefer if I would know that environment variables are 'requiered' inside the Class.

AlexanderLanin: Same here. I would prefer to have all inputs on the level of extension init. And having the rest of the extension as much "functional" as possible. No hidden inputs.

No access to environment from anywhere else. As little access to big objects like app as possible from anywhere else.


But it's ok either way. It's not like it's a huge extension and we loose sight of what is going on where.

AlexanderLanin (approved):

AlexanderLanin (approved):

PR #661, opened 2025-03-14
docs: bugfix css files not found

MaximilianSoerenPollak:
Css files added back due to fixing of static path.
Deleted nolonger needed `layouts.py` 

Issue was that the `css` files were not found. 
Only solution that worked was to put the `html_static_path` into the conf.py otherwise it never found the css files.


AlexanderLanin:
Let's merge this as a quick fix and investigate later.

AlexanderLanin (approved):

PR #667, opened 2025-03-14
docs: add dd to metamodel

hoe-jo:
- static architecture
- dynamic architecture
- sw unit
- adapt Example to fit metamodel


aschemmel-tech (dismissed):

aschemmel-tech (approved):
good change

PR #706, opened 2025-03-17
process: add implementation links to verification process

pahmann:
closes: # 705


masc2023 (approved):
Removing ToDos with concrete Links is fine

PR #714, opened 2025-03-17
docs: update standard pie charts

hoe-jo:
- requirements and workproduct overview is available
- for every standard separately

### DIFF
--- a/docs/_tooling/extensions/score_metamodel/checks/standards.py
+++ b/docs/_tooling/extensions/score_metamodel/checks/standards.py
@@ -13,10 +13,10 @@
 from sphinx.application import Sphinx
 from sphinx_needs.data import NeedsInfoType
 
-from score_metamodel import (
-    CheckLogger,
-    graph_check,
-)
+# from score_metamodel import (
+#    CheckLogger,
+#    graph_check,
+# )
 
 
 def get_standards_needs(needs: list[NeedsInfoType]) -> dict:
@@ -169,11 +169,20 @@ def get_compliance_wp_needs(needs) -> set:
 def my_pie_linked_standard_requirements(needs, results, **kwargs):
     """
     Function to render the chart of check for standard requirements linked to at least an item via compliance-gd.
+    Passed arguments can be accessed via kwargs['arg<position>']
+    See: https://sphinx-needs.readthedocs.io/en/latest/filter.html#arguments
     """
     cnt_connected = 0
     cnt_not_connected = 0
 
-    standards_needs = get_standards_needs(needs)
+    standard = kwargs["arg1"]
+
+    all_standards_needs = get_standards_needs(needs)
+    standards_needs = {
+        k: v
+        for k, v in all_standards_needs.items()
+        if k.startswith(f"std_req__{standard}__")
+    }
     compliance_req_needs = get_compliance_req_needs(needs)
 
     for need in standards_needs.values():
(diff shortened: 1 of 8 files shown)


docs/_tooling/extensions/score_metamodel/checks/standards.py
```diff
@@ -19,41 +19,26 @@
 )
 
 
-def get_standards_needs(needs: list[NeedsInfoType]) -> dict:
+def get_standards_needs(needs: list[NeedsInfoType], standard: str) -> dict:
```
MaximilianSoerenPollak, line 22: I'd prefer to keep this function generic and filter inside the `my_pie_linked_standard_requirements` function, like so:

```python
# inside my_pie_linked_standard_requirements
all_standards_needs = get_standards_needs(needs)
standards_needs = {k:v for k,v in all_standards_needs.items() if k.startswith(standard)}
```



Then not everyone who uses `get_standards_needs` and wants to get all standards has to call the function 3 times with hardcoded strings. 


This though also depends on where we might want to use this.

docs/_tooling/extensions/score_metamodel/checks/standards.py
```diff
@@ -195,7 +180,9 @@ def my_pie_linked_standard_requirements(needs, results, **kwargs):
     cnt_connected = 0
     cnt_not_connected = 0
```
MaximilianSoerenPollak, line 182: We should add a comment here why this is  `arg1` and maybe a link to the [docs](https://sphinx-needs.readthedocs.io/en/latest/filter.html#arguments)

```suggestion
# Passed arguments can be accessed via kwargs['arg<position>'] 
# See: https://sphinx-needs.readthedocs.io/en/latest/filter.html#arguments 
```

docs/_tooling/extensions/score_metamodel/checks/standards.py
```diff
@@ -215,8 +202,9 @@ def my_pie_linked_standard_workproducts(needs, results, **kwargs):
     cwp_connected = 0
     cwp_not_connected = 0
 
-    standard_workproducts = get_standards_workproducts(needs)
+    standard = kwargs["arg1"]
 
+    standard_workproducts = get_standards_workproducts(needs, standard)
```
MaximilianSoerenPollak, line 207: Same as the comment on `get_standards_needs`

docs/process/standards/iso26262/iso26262.rst
```diff
@@ -1611,3 +1611,6 @@ Part 9: Automotive safety integrity level (ASIL)-oriented and safety-oriented an
    but the related requirements are not included as these are not needed to be covered.
 
    A mapping table to ISO 26262 numbering can be provided by SCORE project on request for every holder of a ISO 26262 standard license.
+
+.. needextend:: "standards/iso26262" in docname
```
MaximilianSoerenPollak, line 1615: Might be nicer just to add these 'needextends' to `process/standards/index.rst` ?

MaximilianSoerenPollak (changes requested):
Just some comments that I had.

docs/process/standards/iso26262/iso26262.rst
hoe-jo: True, but therefore it would make sense to define a tagging concept first?

MaximilianSoerenPollak: Probably yes, so I think for now this is fine until we have that concept. Simple change anyway and functionality is the same.

MaximilianSoerenPollak (dismissed):
Looks good, thanks for integrating the changes.

MaximilianSoerenPollak (approved):

PR #746, opened 2025-03-20
docs: redefine how-to merge pull request

pahmann:
It is practice to do squash commits, but this may not be always possible.
This commit adds a draft for the rules concluded in #406.

### DIFF
--- a/docs/contribute/general/git.rst
+++ b/docs/contribute/general/git.rst
@@ -53,7 +53,9 @@ one) logical commits. This is done following the rules below.
 
 Use ``Squash & Merge`` in case:
 
-   #. there is one initial commit + n "fix findings" commits
+   #. there are multiple commits on top of the initial commit
+
+      - Reasons for multiple commits are fixing of review findings or work in progress PRs.
 
       **AND**
    #. all commits by the same author.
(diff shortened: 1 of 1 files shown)


docs/contribute/general/git.rst
```diff
@@ -49,7 +49,26 @@ So this file must at least include the following lines:
 ***************
 
 Before merging a PR all commits shall be squashed into few (desired only
-one) logical commits.
+one) logical commits. This is done following the rules below.
+
+Use ``Squash & Merge`` in case:
+
+   #. there is one initial commit + n "fix findings" commits
```
AlexanderLanin, line 56: Or any other form of commits that we do not want to preserve: "work in progress" + "blah" + "blub" + "still work in progress"

docs/contribute/general/git.rst
```diff
@@ -49,7 +49,26 @@ So this file must at least include the following lines:
...
+      **AND**
+   #. all commits by the same author.
+
+Use ``Rebase & Merge`` or ``Merge Commit`` in case:
+
+   #. the commits address different topics
+      (sort by topics and combine each topic to a single commit)
+
+      **OR**
+   #. the commits have different authors
```
AlexanderLanin, line 67: Our own old discussion is slightly wrong here. We often have PRs with commits by different authors that we squash. That happens rather frequently when collaborating or even pair programming. I know, crazy. But sometimes we collaborate 🗡️ 

VS-Code and GitHub-UI will typically auto-append `Co-authored-by: name <name@example.com>` in such cases.

docs/contribute/general/git.rst
```diff
@@ -49,7 +49,26 @@ So this file must at least include the following lines:
...
+
+   #. there is one initial commit + n "fix findings" commits
+
+      **AND**
+   #. all commits by the same author.
+
+Use ``Rebase & Merge`` or ``Merge Commit`` in case:
+
+   #. the commits address different topics
+      (sort by topics and combine each topic to a single commit)
```
AlexanderLanin, line 64: Add something like: "advice: create separate PRs for separate topics"

docs/contribute/general/git.rst
pahmann: I learnt today that it is demanded by one committer company that it has to be clearly stated which parts belong to which company for IP topics, liability and so.

AlexanderLanin: I'm sorry, but that needs to be clarified further as we typically work in pair programming across companies.

AlexanderLanin: And I see your updated PR. But it doesn't quite solve the problem 100%. As the Author vs Co-Author are somewhat random for each commit.

AlexanderLanin:
Let's say this is better than before and merge it. We can discuss any further improvements after merging.

AlexanderLanin (approved):

Issue #788, opened 2025-03-25
Header-service implementation


MaximilianSoerenPollak:
Implementation of #205

PR #793, opened 2025-03-25
tooling: integrate header service

mmr1909:
Add service to automatically fill out review and approval information in the header of an approval record.

Fixes https://github.com/eclipse-score/score/issues/788


docs/_tooling/extensions/score_header_service/test/test_header_service.py
```diff
@@ -0,0 +1,319 @@
+# *******************************************************************************
```
MaximilianSoerenPollak, line 1: Not a fan of mocks, but I think here this is the only choice.

docs/_tooling/extensions/score_header_service/test/test_header_service_integration.py
```diff
@@ -0,0 +1,220 @@
...
+        mock_basic_conf,
+        basic_needs,
+        template_needs,
+    )
+    try:
+        app.build()
+        html_content = (app.outdir / "index.html").read_text()
+        mock_extract_merge_commit_data.assert_called_once()
+        assert "<td><p>John Doe</p></td>" in html_content
+        assert '<div class="line">aprover_1</div>' in html_content
```
MaximilianSoerenPollak, line 216: It might be better to just assert the actual values here like  `John Doe` as this then eliminates the risk of this test failing if sphinx changes something in their engine on how stuff is converted to HTML.

docs/_tooling/extensions/score_layout/sphinx_options.py
```diff
@@ -38,6 +38,10 @@
             "footer_right": [],
         },
     },
+    "focus": {
```
MaximilianSoerenPollak, line 41: Might be good to add a comment or link next to here to explain why this is different or what this layout does that you want to achieve.

MaximilianSoerenPollak (commented):
Looks overall good, just docstrings and some small things I'd like to see adressed.

MaximilianSoerenPollak (approved):
Changes since last review looked over in call. 👍

PR #822, opened 2025-03-31
Feature Request: JSON Parser

alexandruiulian10:
Provides a proposal for a safe JSON parser feature.


docs/features/communication/json/index.rst
```diff
@@ -0,0 +1,132 @@
...
+
+
+Basic Functionality
+----------------------
+The component shall support reading from a .json file or a buffer and convert that information to a compatible user format.
+
+
+Future Extensions
+----------------------
+The component shall suppot a future extension for converting from user format into .json format and storing it into a file or buffer.
```
simon-acc, line 51: typo

```suggestion
The component shall support a future extension for converting from user format into .json format and storing it into a file or buffer.
```

alexandruiulian10: Fixed in the latest commit.

rattenking52 (changes requested):

docs/features/communication/json/index.rst
```diff
@@ -0,0 +1,132 @@
...
+
+Backwards Compatibility
+=======================
+
+As there is currently no previous solution in S-CORE, no backwards compatibility is required.
+
+Security Impact
+===============
+
+The module will likely work with input and output streams.
```
masc2023, line 116: so what is the impact? Access Control needed?, please set also tag security to yes

alexandruiulian10: For access control and manipulation prevention (e.g. dm-verity) the hosting process and system configuration need to take care of it.

rattenking52 (changes requested):

docs/requirements/stakeholder/index.rst
```diff
@@ -539,6 +539,17 @@ Communication
    * Diagnostic jobs
 
 
+.. stkh_req:: JSON-Library
+   :id: stkh_req__communication__json_parser
```
lammjo, line 543: I am not too familiar with the S-CORE process yet, but intuitively I would propose to make the wording match:
```suggestion
.. stkh_req:: JSON-Library
   :id: stkh_req__communication__json_library
```
(Of course this entails changing docs/features/baselibs/json/requirements/index.rst where the IDs are used)

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. feat_req:: JSON parser ASIL level
+   :id: feat_req__json__asil
+   :reqtype: Functional
+   :security: YES
+   :safety: ASIL_B
+   :satisfies: stkh_req__communication__json_parser
+   :status: valid
+
+   The JSON parser shall support safe communication up to ASIL-B.
```
lammjo, line 91: As suggested below, I would propose to make the wording match:
```suggestion
.. feat_req:: JSON library ASIL level
   :id: feat_req__json__asil
   :reqtype: Functional
   :security: YES
   :safety: ASIL_B
   :satisfies: stkh_req__communication__json_library
   :status: valid

   The JSON library shall support safe communication up to ASIL-B.
```

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+General Requirements
+====================
+
+.. feat_req:: JSON Validation
+   :id: feat_req__json__validation
+   :reqtype: Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: stkh_req__communication__json_parser
```
alexandruiulian10, line 26:
```suggestion
   :satisfies: stkh_req__communication__json_library
```

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+   :status: valid
+
+   The JSON-Library shall provide a service to check the well-formedness of JSON data.
+
+.. feat_req:: JSON Deserialization
+   :id: feat_req__json__deserialization
+   :reqtype: Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: stkh_req__communication__json_parser
```
alexandruiulian10, line 36:
```suggestion
   :satisfies: stkh_req__communication__json_library
```

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Full testability for the user facing API
+========================================
+
+.. feat_req:: Fully mockable public API
+   :id: feat_req__json__testability_mock_api
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: QM
+   :satisfies: stkh_req__communication__json_parser
```
alexandruiulian10, line 75:
```suggestion
   :satisfies: stkh_req__communication__json_library
```

docs/features/baselibs/json/index.rst
```diff
@@ -0,0 +1,130 @@
...
+   # *******************************************************************************
+
+JSON-Library
+===========================
+
+.. document:: JSON-Library
+   :id: doc__json
+   :status: valid
+   :safety: ASIL_B
+   :tags: contribution_request, feature_request
```
masc2023, line 22: contribution_request is deprecated, shall be removed

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. feat_req:: JSON library ASIL level
+   :id: feat_req__json__asil
+   :reqtype: Functional
+   :security: YES
+   :safety: ASIL_B
+   :satisfies: stkh_req__communication__json_library
+   :status: valid
+
+   The JSON library shall support safe communication up to ASIL-B.
```
masc2023, line 91: so why not link to that requirement [stkh_req__communication__safe](https://eclipse-score.github.io/score/main/requirements/stakeholder/index.html#stkh_req__communication__safe)

docs/requirements/stakeholder/index.rst
```diff
@@ -539,6 +539,17 @@ Communication
...
 
+.. stkh_req:: JSON-Library
+   :id: stkh_req__communication__json_library
+   :reqtype: Functional
+   :security: NO
+   :safety: ASIL_B
+   :rationale: tbd
+   :status: valid
+
+   The software platform shall provide a JSON-Library with parsing functionality.
```
masc2023, line 550: Does this no derive from [stkh_req__functional_req__safe_config](https://eclipse-score.github.io/score/main/requirements/stakeholder/index.html#stkh_req__functional_req__safe_config)?

docs/features/communication/json/index.rst
masc2023: then may add this as assumption of use in the feature request, otherwise it get lost, by the way, this is more a component_request as part of feature:baselib

docs/features/baselibs/json/index.rst
4og: I agree that the JSON lib should be a component request. It also means that this file has to follow the https://eclipse-score.github.io/score/main/process/process_areas/change_management/guidance/change_management_component_template.html I.e. Feature flag section is not needed for a component.

alexandruiulian10: I updated that tag to component_request and removed the Feature flag section in the last commit

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: stkh_req__communication__safe is ASIL-D, while this feature is meant to go only up to ASIL-B

docs/requirements/stakeholder/index.rst
```diff
@@ -539,6 +539,17 @@ Communication
    * Diagnostic jobs
 
 
+.. stkh_req:: JSON-Library
```
4og: Discussed in https://github.com/orgs/eclipse-score/discussions/1239

This stakeholder req should be converted into a feature req of the Base libs

docs/features/baselibs/json/index.rst
```diff
@@ -0,0 +1,122 @@
...
+   # *******************************************************************************
+
+JSON-Library
+===========================
+
+.. document:: JSON-Library
+   :id: doc__json
+   :status: valid
+   :safety: ASIL_B
+   :tags: component_request, feature_request
```
4og: Delete feature_request

4og (changes requested):

docs/requirements/stakeholder/index.rst
alexandruiulian10: Changed in the latest commit, as discussed

docs/features/baselibs/json/index.rst
alexandruiulian10: Removed in the latest commit

docs/features/baselibs/json/index.rst
```diff
@@ -0,0 +1,122 @@
...
+.. toctree::
+   :hidden:
+
+   requirements/index.rst
+
+
+Abstract
+========
+
+This feature request proposes the integration of a safe JSON-Library with parsing functionality.
```
4og: This component request...

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+User friendly API for information exchange
+==========================================
+
+.. comp_req:: Support for programming language idioms
+   :id: comp_req__json__lang_idioms
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: QM
+   :satisfies: feat_req__baselibs__json_library
```
4og: The [feat_req__baselibs__consistent_apis](https://eclipse-score.github.io/score/main/features/baselibs/docs/requirements/index.html#feat_req__baselibs__consistent_apis) would fit here as well.

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Safety Impact
+=============
+
+.. comp_req:: JSON library ASIL level
+   :id: comp_req__json__asil
+   :reqtype: Functional
+   :security: YES
+   :safety: ASIL_B
+   :satisfies: feat_req__baselibs__json_library
```
4og: Please also add [feat_req__baselibs__safety](https://eclipse-score.github.io/score/main/features/baselibs/docs/requirements/index.html#feat_req__baselibs__safety)

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. comp_req:: JSON Deserialization
+   :id: comp_req__json__deserialization
+   :reqtype: Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: feat_req__baselibs__json_library
+   :status: valid
+
+   The JSON-Library shall provide a service to parse JSON data according to RFC8259.
```
4og: This RFC8259 doesn't support large integers. Do we want a support for large ints in S-CORE at all?

lennartbecker-d: You are right. While RFC8259 only supports integers in the range from [-(2^53)+1, (2^53)-1], the JSON-Library from Niels Lohmann also handles [-(2^63), (2^64)-1] (see unit-class_parser_diagnostic_positions.cpp). We thought about precising this enlargement in the argumentation for evidence.

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. comp_req:: Fully mockable public API
+   :id: comp_req__json__testability_mock_api
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: QM
+   :satisfies: feat_req__baselibs__json_library
+   :status: valid
+
+   The public API shall be fully mockable.
```
lennartbecker-d: According to the [coding guidelines](https://github.com/eclipse-score/score/blob/ba8cc391e75d332f0dd4c44a2d21b04d81b99d7e/docs/contribute/development/python/coding_guidelines.rst#L59) mocking seems to be an anti-pattern at least for Python. Does this hold true in this case, too? 
Which public API is meant here?

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. comp_req:: Support for programming language idioms
+   :id: comp_req__json__lang_idioms
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: QM
+   :satisfies: feat_req__baselibs__json_library
+   :status: valid
+
+   Each public API shall support the idioms of the programming language it is written in.
```
lennartbecker-d: What is the intention here? Which public API is meant here. Do you mean the one implemented in [baselibs](https://github.com/eclipse-score/baselibs/tree/main/score/json) or the one provided by nlohmann? 
Just considering how this might be supported with evidence? Is an SME review necessary here?

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. comp_req:: Use programming language infrastructure
+   :id: comp_req__json__lang_infra
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: QM
+   :satisfies: feat_req__baselibs__json_library
+   :status: valid
+
+   Each public API shall use core infrastructure of its programming language and accompanying standard libraries,
```
lennartbecker-d: This will be proven to be fulfilled by SME review or on codebasis?  Which API is meant here again? The requirement appears to be very general and vague.

lennartbecker-d (commented):
To sum things up: 
Which public API is meant in the different requirements?

docs/features/baselibs/json/index.rst
alexandruiulian10: Corrected in the latest commit.

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: Added in the latest commit.

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: Added in the latest commit.

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+   :reqtype: Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: feat_req__baselibs__json_library
+   :status: valid
+
+   The JSON-Library shall provide a service to parse JSON data according to RFC8259.
+
+User friendly API for information exchange
+==========================================
```
LittleHuba: We should add something about a strictly typed API. E.g. when a user requests the value of a key, then it is only returned if the used API returns a type that is compatible to the actual data within the JSON.

Example: If a user retrieves key "abc" with value `-10` from the library and requests a return type "uint8_t" then he should get an error.

Same for values exceeding the range of a type or other type mismatches.

alexandruiulian10: Added a requirement for this in the latest commit.

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: As clarified offline, the public API mentioned in these requirements is the API of the JSON library provided by baselibs, which will likely be a wrapper over the nlohmann JSON implementation.

I'm not familiar with mocking being an anti-pattern. To me the referenced statement seems to suggest that it may be used consciously and perhaps sparingly, rather than not at all. Thus it would make sense to keep the option open in case it may be needed. There is a similar mockability requirement in the score communication solutions and it seemed reasonable to add it here also, but I don't have a strong feeling about it.

LittleHuba: When doing dependency injections you normally do so with some kind of test double (dummy, fake, stub, mock,...).
The wording here falls short and is therefore misleading. What it should state is: 

> The public API of the library shall support dependency injection with test doubles.
> Note: This enables full testability of the user code.

I'll talk with @AlexanderLanin to rephrase the the coding guidelines. Calling mocking an anti-pattern is also misleading. There are very good cases where mocking is the correct way. Frightening developers away from it will only lead to worse solutions. What we should state is along the lines:

> Do not overuse mocks because it makes your tests brittle. Consider using other test doubles instead if you don't need the specific capabilities of a mock (verification of calls).

alexandruiulian10: I added this change with the latest commit.

docs/features/baselibs/json/requirements/index.rst
sankurm, line 1: The requirements around the 'parsing functionality' of JSON have been captured here. The outcome of the parsing shall be a **parsed representation of the JSON**. Let me refer to this as the `parsed_json`. I did not find any requirements that relate to the `parsed_json`. I would like to note some high-level aspects related to the `parsed_json`.

The `parsed_json` shall
1. be a [regular type](https://en.cppreference.com/w/cpp/concepts/regular.html)
2. not expose the implementation library types (as this can restrict future changes/evolution)

Some of the above requirements may relate to other components & modules too. Such high-level requirements should be considered to be elevated as guidance (e.g. thou shalt not expose implementation-level data-types across the component boundaries).

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,106 @@
...
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: QM
+   :satisfies: feat_req__baselibs__json_library
+   :status: valid
+
+   The public API shall use core infrastructure of its programming language and accompanying standard libraries,
+   whenever possible and meaningful.
+
+   Note: This includes error handling.
```
sankurm: Our implementations should have a greater guarantee. Error handling should not be left to a Note in my opinion.

I would like to suggest the following requirements: (again, if found generic, we could promote to overall Eclipse S-CORE level)
1. Error encountered (due to incorrect input, system error, or otherwise) should be reported back to the user
2. Error reporting via non-throwing mechanisms is preferrable (like [`optional<T>`](https://en.cppreference.com/w/cpp/utility/optional.html), [`std::expected<T>`](https://en.cppreference.com/w/cpp/utility/expected.html), [`score::cpp::optional`](https://github.com/eclipse-score/baselibs/blob/main/score/language/futurecpp/include/score/optional.hpp#L70), [`score::Result<T>`](https://github.com/eclipse-score/baselibs/blob/main/score/result/result.h#L35))
3. In case of error, it should be difficult for the users to carry out operations on the parsed representation of the JSON as if it was successful. The API shall allow foe easy error handling in case of an error. This is to make correct implementations easier than not.

Sample code:
```c++
//Let's say the application wants to apply config read from a JSON
void apply(score::json);

//Possible API offered by this component
using score::json = ...defined by this component...;
score::Result<score::json> deserialize(const std::string_view json_str) { ... }

//Such an API cannot be used incorrectly
auto ill_formatted_json = "xyz"sv;
auto my_config = deserialize(ill_formatted_json);

//The following must not even compile as it is trying to use the return as json
apply(my_config);

//Error handling must be possible
score::json default_config = { {"threshold-min", 80}, {"threshold_max", 95} };  //should be possible to construct directly
apply(my_config.value_or(default_config));
```

docs/features/baselibs/json/requirements/index.rst
sankurm, line 1: A possible requirement could be to ensure correctness by default by allowing compile-time constructs & checks.

I would put it as below:
1. to be able to parse a JSON at compilation-time. This can help have compile-time tests (with facilities like `static_assert`). This will block compilation if the implementation does develop an error or incorrectness. This will ensure that a component with errors cannot exist even in the absence of a pipeline (yes, the tests should be extensive). Plus, generally, it will shift-left catching errors.

I understand that the nlohmann library does not offer compile-time parsing. So, this can be kept as a good to have requirement. Or be dropped, if we decide to not support.

sankurm, line 1: Do we want JSON schema validation as a requirement?

Something like this:
1. It should be possible to validate a json against a json-schema.

This could be lower in priority as compared to the core functionality depending on the need across all partners.

sankurm (commented):
Thanks a lot for the great progress in this area. 

I classify my comments with the following priority/importance (must/should/could/may):
1. Must: `parsed_json` reqs
2. Must: Error handling reqs
3. Could: JSON schema reqs
4. May: Compile-time parsing

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: I added a requirement for this in the latest commit.

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: This is a good point, but I think we should clarify if it should be added here or somewhere else. It seems to me more like a general S-CORE guideline than something specific to the JSON library.

docs/features/baselibs/json/requirements/index.rst
4og: ok

docs/features/baselibs/json/index.rst
```diff
@@ -0,0 +1,123 @@
+..
```
4og: The content of this file should be placed in `docs/modules/baselibs/json/docs/index.rst`

docs/features/baselibs/json/requirements/index.rst
```diff
@@ -0,0 +1,118 @@
+..
```
4og: The file should be relocated to `docs/modules/baselibs/json/docs/requirements.rst`

4og (dismissed):

docs/features/baselibs/json/index.rst
alexandruiulian10: Moved in the latest commit.

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: Moved in the latest commit.

docs/features/communication/json/index.rst
aschemmel-tech: Checked: is now a component request, but AoU is still missing. Additionally we should have a seperate AoU that states that the integrity of the input JSON data is with the user (i.e. if it comes from a file it must be protected)

docs/features/baselibs/json/index.rst
aschemmel-tech: checked ok

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: What has the json library to do with communication? Rather configuration.

docs/requirements/stakeholder/index.rst
aschemmel-tech: Yes. please add this stakeholder requirement as satisfies.

docs/requirements/stakeholder/index.rst
aschemmel-tech: Checked ok

docs/features/baselibs/json/index.rst
aschemmel-tech: Checked ok

docs/features/baselibs/json/index.rst
aschemmel-tech: Checked ok

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: Checked ok

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: Checked ok

docs/features/baselibs/json/index.rst
aschemmel-tech: Checked ok

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: Checked ok

docs/requirements/stakeholder/index.rst
alexandruiulian10: This requirement was removed in the meantime. It was decided to have only one stakeholder requirement for the base libraries and have all contained features link to it - [stkh_req__functional_req__base_libraries](https://eclipse-score.github.io/score/main/requirements/stakeholder/index.html#stkh_req__functional_req__base_libraries).

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: We discussed this already in the past and then the decision was to not have schema validation. Can be added to the component requirements later if really required. Also here: add this to the component request in https://eclipse-score.github.io/score/pr-822/modules/baselibs/json/docs/index.html#future-extensions with the addition that this might even be an additional component (request).

aschemmel-tech: We should only have requirements which can be fulfilled by our implementation (OSS component we select). Can be added to the component requirements later if really required. So we should add this to the component request in https://eclipse-score.github.io/score/pr-822/modules/baselibs/json/docs/index.html#future-extensions

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: I rephrased it in the latest commit.

docs/features/communication/json/index.rst
alexandruiulian10: I added this in the latest commit.

docs/modules/baselibs/json/docs/requirements/index.rst
```diff
@@ -0,0 +1,118 @@
...
+.. comp_req:: Enforce strict type compatibility
+   :id: comp_req__json__type_compatibility
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: feat_req__baselibs__json_library
+   :status: valid
+
+   The public API shall enforce strict type compatibility. When a user requests a value, the API shall 
+   validate that the requested return type is compatible with the type of the stored JSON data.
```
LittleHuba:
```suggestion
   validate that the requested return type is compatible with the type and value of the stored JSON data.
```

docs/features/baselibs/json/requirements/index.rst
LittleHuba: Thanks, I added a new review for your added requirement.

docs/features/communication/json/index.rst
aschemmel-tech: Sorry about not making this clear: The AoU should be done as a requirement using template https://eclipse-score.github.io/process_description/main/process_areas/requirements_engineering/guidance/requirements_templates.html#gd_temp__req__aou_req

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: checked ok

docs/requirements/stakeholder/index.rst
aschemmel-tech: There is the feature requirement https://eclipse-score.github.io/score/pr-822/features/baselibs/docs/requirements/index.html#feat_req__baselibs__json_library - and this should also link to [stkh_req__functional_req__safe_config](https://eclipse-score.github.io/score/main/requirements/stakeholder/index.html#stkh_req__functional_req__safe_config) as it covers this partly

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: same comment as above

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: Requirement is "Non-functional" so a "SME review" is appropriate. After discussion with @LittleHuba we concluded we wanted this kind of requirement (which is applicable for every user API) rather defined in a coding guideline. But this is a future PR so let's keep this here for the moment.

docs/modules/baselibs/json/docs/requirements/index.rst
```diff
@@ -0,0 +1,118 @@
...
+   :status: valid
+
+   The public API shall use core infrastructure of its programming language and accompanying standard libraries,
+   whenever possible and meaningful.
+
+   Note: This includes error handling.
+
+.. comp_req:: Enforce strict type compatibility
+   :id: comp_req__json__type_compatibility
+   :reqtype: Non-Functional
```
aschemmel-tech: I would say this is a Functional requirement, as could be tested.

aschemmel-tech (changes requested):
see also updated inline comments

docs/features/baselibs/json/requirements/index.rst
aschemmel-tech: Also here, I discussed with @LittleHuba and said we would not have this in every component's requirements but add to a Coding Guideline. No actions in this PR. https://github.com/eclipse-score/score/issues/1533 created

docs/modules/baselibs/json/docs/requirements/index.rst
```diff
@@ -0,0 +1,118 @@
...
+
+   Note: This enables full testability of the user code.
+
+Safety Impact
+=============
+
+.. comp_req:: JSON library ASIL level
+   :id: comp_req__json__asil
+   :reqtype: Functional
+   :security: YES
```
PhilipPartsch: I believe this requirement is not security relevant.

alexandruiulian10: Changed to :security: NO in the latest commit.

docs/modules/baselibs/json/docs/requirements/index.rst
alexandruiulian10: Changed to :reqtype: Functional in the latest commit.

docs/features/communication/json/index.rst
alexandruiulian10: I added the two AoU requirements in the latest commit.

aschemmel-tech: checked ok

docs/modules/baselibs/json/docs/requirements/index.rst
aschemmel-tech: Checked OK

docs/modules/baselibs/json/docs/requirements/index.rst
aschemmel-tech: Checked OK

docs/features/baselibs/json/requirements/index.rst
alexandruiulian10: I added this point to the future extensions section.

alexandruiulian10: I added this point to the future extensions section.

aschemmel-tech: checked ok

aschemmel-tech: checked ok

antonkri (approved):

PR #847, opened 2025-04-03
git: Update gitlint error message

nicu1989:
Use correct link to docs in gitlint error message.


dcalavrezo-qorix (approved):
just consider defining a reusable workflow too

MaximilianSoerenPollak (approved):
Looks good, questions got answered in call.

PR #852, opened 2025-04-03
cicd: use CR checker from registry

dcalavrezo-qorix:
Decouple the copyright checker and use the module from our custom BZL registry

Addresses: #813


MaximilianSoerenPollak (approved):
Looks good. 

Thanks for the implementation, good to see our work is slowly visible.

Issue #273, opened 2025-02-03
Feature Request Orchestration


vinodreddy-g:
Acceptance Criteria:
- Feature Request written according to the [Change Management](https://eclipse-score.github.io/score/main/process/process_areas/change_management/change_management_concept.html) & [Feature Request Template](https://eclipse-score.github.io/score/main/process/process_areas/change_management/guidance/change_management_feature_template.html)
- Feature requirements written according to the [Requirements Engineering](https://eclipse-score.github.io/score/main/process/process_areas/requirements_engineering/requirements_concept.html)
- If necessary: extend the stakeholder requirements written according to the [Requirements Engineering](https://eclipse-score.github.io/score/main/process/process_areas/requirements_engineering/requirements_concept.html)

FScholPer:
@vinodreddy-g  Can you please add subissues with the gaps which are identified and not already there

PR #857, opened 2025-04-05
docs: added orchestration feature request

qor-lb:
closes: #273

### DIFF
--- a/docs/features/orchestration/index.rst
+++ b/docs/features/orchestration/index.rst
@@ -1,0 +1,261 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2025 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+.. _orchestration_feature:
+
+Orchestration
+#############
+
+.. document:: Orchestration
+   :id: doc__orchestration
+   :status: valid
+   :safety: ASIL_B
+   :tags: change_management
+
+
+.. toctree::
+   :hidden:
+
+   requirements.rst
+
+
+Feature flag
+============
+
+To activate this feature, use the following feature flag:
+
+``experimental_orchestration``
+
+
+Abstract
+========
+
+In response to the increasing complexity of modern centralized E/E architectures and the need to support hundreds of applications, this feature request proposes a comprehensive orchestration framework for managing concurrency in high-performance computing (HPC) systems. The motivation for this proposal is rooted in the significant performance penalties incurred by conventional thread-based approaches, where an excessive number of threads leads to costly context switching in operating systems.
+
+The proposed solution introduces user-level scheduling through cooperative multi-tasking, allowing task switches to occur in the nanosecond range instead of microseconds. By treating tasks as the fundamental unit of concurrency and enabling multiple tasks to share the same OS thread, the framework significantly reduces overhead and simplifies resource allocation.
+
+Overall, this orchestration feature is designed to provide a safe and secure, deterministic, user-friendly interface that streamlines concurrent execution, optimizes resource utilization, and facilitates the reliable integration of mixed-criticality applications in complex automotive and embedded systems.
+
+
+Motivation
+==========
+
+The rapid evolution of high-performance computing (HPC) architectures and the increased availability of multi-core processors in centralized E/E systems have driven the need for a platform capable of hosting advanced ADAS/AD applications while upintegrating existing ECU functionalities. This approach facilitates the deployment of loosely coupled applications that support flexible maintenance and updates, with the goal of hosting between 100 and 150 individual applications.
+
+
+Motivation: Increasing Performance in a Highly Concurrent System
+-----------------------------------------------------------------
+
+In existing platforms for microprocessors (µP), each application is expected to interact with approximately 15 system services or daemons - such as ``ara::diag`` for diagnostics and ``ara::com`` for SOME/IP communication. Under a straightforward implementation, this interaction model results in the creation of around 15 threads per application. When scaled to 100-150 applications, this amounts to roughly 1500 to 2250 threads solely managing inter-process communication, excluding those performing the core application tasks.
+
+Given that the HPC's µP typically provides between 2 and 16 cores, only a limited number of threads can be processed in parallel. In POSIX-like operating systems, threads become the currency for concurrency, meaning that when the thread count far exceeds available cores, the system must rely on context switching. With context switching times estimated between 2µs and 4.5µs [#f1]_ [#f2]_ [#f3]_, even a 100ms time slice could spend between 3% and 10% of its duration on context switching alone - assuming each thread is scheduled once. This overhead increases dramatically if threads are forced to switch more frequently due to competition for processing time.
+
+Furthermore, reducing context switching overhead not only improves raw performance but also enhances overall safety and reliability. By minimizing the number of context switches, the CPU can devote more time to executing critical application logic and monitoring system health, thereby reducing the risk of timing violations in safety-critical functions. This reduction in overhead also minimizes the likelihood of cascading delays and resource starvation, which bolsters system resilience and ensures predictable, real-time execution for safety-critical tasks.
+
+
+Motivation: Enabling a User-Friendly and Deterministic Interface for Concurrent Programs
+-----------------------------------------------------------------------------------------
+
+Concurrent programming in our target environment spans multiple scopes. At one level, concurrency exists within the algorithms of individual applications or system services, while at another level, multiple applications must execute concurrently across the platform. The challenge is to offer an interface that is not only simple and expressive but also deterministic and reliable - an important requirement in safety-critical systems.
+
+Traditional thread-based concurrency in POSIX-like environments introduces complexities such as deadlocks, livelocks, and starvation. These issues, coupled with the inherent difficulties in debugging and validating thread-based systems, can compromise both performance and reliability. [#f4]_ [#f5]_ [#f6]_ Moreover, current designs often separate the management of timing requirements, monitoring, and error handling from the control flow. Integrating these aspects closer to the application logic would promote higher cohesion and lower coupling, enabling more effective debugging and validation, particularly when addressing application-specific scenarios.
+
+
+Motivation: Managing Reliable Integrations in Mixed-Criticality Environments
+-----------------------------------------------------------------------------
+
+The broader objective of the HPC platform is to support the concurrent integration and maintenance of multiple applications, potentially sourced from different vendors. Unlike traditional microcontroller (µC) platforms that are statically configured, this platform must allow dynamic updates, upgrades, and new deployments throughout the vehicle's lifecycle. This introduces significant complexity in managing concurrency, particularly in mixed-criticality environments where applications of varying criticality levels must coexist without interference.
+
+Currently, application developers are expected to manage runtime scheduling directly via the operating system, controlling thread lifecycles, priorities, and affinities. Such an approach ties the configuration too closely to a specific deployment scenario and often leads to discrepancies between behaviors observed during development and those on the target system. This misalignment complicates integration efforts, as integrators must repeatedly iterate with the application developer to meet the system's reliability requirements. An orchestrator that abstracts these complexities could alleviate these challenges by offering a uniform, deterministic interface for managing runtime scheduling, thus ensuring that lower-criticality applications do not adversely affect the performance or reliability of higher-criticality ones.
+
+
+.. Rationale
+.. ==========
+
+
+Specification
+=============
+
+Now that we know the motivation behind the feature request, let's dive deeper into the proposed solution for the problem statements described above.
+
+
+Specification: How to increase performance in a highly concurrent system?
+--------------------------------------------------------------------------
+
+To address the performance challenges described in the motivation - where an excessive number of threads leads to significant context-switch overhead - the proposed solution is to adopt user-level scheduling through cooperative multi-tasking. This approach minimizes overhead by reusing existing threads within a process instead of creating one thread per session or task within a service or application component. By handling task switching at the user level, the system avoids the costly transition between user space and kernel space, reducing the typical context switch cost from the microsecond to the nanosecond range. [#s1]_ [#s2]_ [#s3]_ [#s4]_
+
+The proposed solution provides a framework where tasks, rather than threads, serve as the fundamental unit of concurrency. In this model, concurrent tasks can share the same OS thread, thereby significantly reducing the number of threads and the associated context-switch overhead. To enable parallelism for concurrent tasks, the solution will implement a configurable thread pool. In the target application of the framework, additional thread pools will only be introduced when necessary - for instance, in cases where application tasks differ in criticality or are separated by process boundaries. This reuse of threads not only minimizes the performance penalty associated with managing a vast number of threads but also simplifies resource allocation and scheduling in the final target scenario.
+
+The solution includes the following key functionalities:
+   - User-level scheduling to efficiently handle task switches in the nanosecond range.
+   - Cooperative multi-tasking, allowing multiple concurrent tasks to use the same OS thread.
+   - Provision of a configurable thread pool to enable parallelism for concurrent tasks.
+   - Introduction of additional thread pools only when necessary, such as when tasks differ in criticality or require separation by process boundaries.
+
+.. admonition:: Related Requirements
+
+   - :need:`feat_req__orchestration__user_level`
+   - :need:`feat_req__orchestration__cooperative_multi`
+   - :need:`feat_req__orchestration__configurable_thread`
+
+
+.. _how_to_enable_a_user_friendly_deterministic_interface_for_concurrent_programs:
+
+Specification: How to enable a user-friendly & deterministic interface for concurrent programs?
+------------------------------------------------------------------------------------------------
+
+To harness the benefits of user-space multi-tasking while still providing a user-friendly and deterministic interface for concurrent programs, this proposal advocates for a nested task-based programming framework. The choice for a nested structure over a graph-based one is driven by the need to design reliable programs and enable straightforward control flow analysis - a requirement that becomes critical during safety inspections. Although graph-based structures may have a gentler learning curve and offer rapid initial results, they often become limiting when more complex scheduling descriptions are needed, such as conditional branching, time monitoring, and error handling paths.
+
+The envisioned programming framework will be:
+   - Free from explicit thread management.
+   - Free from complex synchronization mechanisms.
+   - Capable of expressing both concurrent and sequential dependencies.
+   - Capable of expressing conditional branching within the program.
+   - Capable of expressing timing constraints and error handling paths directly within the program.
+   - Expressive, transparent, and deterministic to facilitate straightforward control-flow analysis.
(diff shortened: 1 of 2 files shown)


docs/features/orchestration/index.rst
```diff
@@ -0,0 +1,261 @@
...
+
+``experimental_orchestration``
+
+
+Abstract
+========
+
+In response to the increasing complexity of modern centralized E/E architectures and the need to support hundreds of applications, this feature request proposes a comprehensive orchestration framework for managing concurrency in high-performance computing (HPC) systems. The motivation for this proposal is rooted in the significant performance penalties incurred by conventional thread-based approaches, where an excessive number of threads leads to costly context switching in operating systems.
+
+The proposed solution introduces user-level scheduling through cooperative multi-tasking, allowing task switches to occur in the nanosecond range instead of microseconds. By treating tasks as the fundamental unit of concurrency and enabling multiple tasks to share the same OS thread, the framework significantly reduces overhead and simplifies resource allocation.
```
armin-acn, line 46: There are existing frameworks for user-level scheduling based on cooperative multitasking in many languages including Rust and C++. As you know, Rust even provides language elements for it. Unfortunately, the frameworks I know of (e.g. "Tokio" for Rust) are not safety certified.

Having a (safety-certified) framework for task-based asynchronous programming could be very beneficial for the reasons you are pointing out below. I think, such a framework should enable application development in the common async programming model using async functions, spawning of tasks, and "await"-ing of futures (as, e.g. with "Tokio" in Rust). 

According to my understanding, the functionality requested in this feature request more or less needs such a framework (at least a minimal one) as a basis. Thus, it could make sense to create a separate feature request for a safety-certified async-framework and then base remainder of the present feature request on top of that.

docs/features/orchestration/requirements.rst
```diff
@@ -0,0 +1,228 @@
...
+
+.. feat_req:: Cooperative Multi Tasking
+   :id: feat_req__orchestration__cooperative_multi
+   :reqtype: Functional
+   :security: NO
+   :safety: QM
+   :satisfies: stkh_req__execution_model__processes
+   :status: invalid
+
+   The system **SHALL** support cooperative multi-tasking, allowing multiple concurrent tasks to share the same OS thread.
```
armin-acn, line 64: See above: This could be achieved by providing an async framework similar to the known ones.

docs/features/orchestration/requirements.rst
```diff
@@ -0,0 +1,228 @@
...
+.. feat_req:: Configurable Thread Pool
+   :id: feat_req__orchestration__configurable_thread
+   :reqtype: Functional
+   :security: NO
+   :safety: QM
+   :satisfies: stkh_req__execution_model__processes
+   :status: invalid
+
+   The system **SHALL** provide a configurable thread pool for executing concurrent tasks. Additional thread pools **MAY** be introduced only when necessary (e.g., when tasks differ in criticality or require separation by process boundaries).
+
```
armin-acn, line 79: See above: This could be achieved by providing an async framework similar to the known ones.

docs/features/orchestration/requirements.rst
```diff
@@ -0,0 +1,228 @@
...
+.. feat_req:: Deterministic and Expressive Programming Framework
+   :id: feat_req__orchestration__deterministic_and
+   :reqtype: Functional
+   :security: NO
+   :safety: QM
+   :satisfies: stkh_req__execution_model__processes
+   :status: invalid
+
+   The programming framework **SHALL** allow developers to express concurrent and sequential dependencies, conditional branching, timing constraints, and error handling paths while abstracting explicit thread management and complex synchronization.
+
```
armin-acn, line 93: See above: This could be achieved by providing an async framework similar to the known ones. There is no need for a "nested" language to achieve this.

docs/features/orchestration/index.rst
mthiede-acn2, line 46: I think we had a broad agreement in score that if possible, the native means of the programming language or programming environment should be used. In case of user space scheduling and Rust, Rust async and the corresponding well known and wide spread API already solves this problem. Of course this needs a safe async runtime and working on that would be a very useful thing to do.

docs/features/orchestration/requirements.rst
```diff
@@ -0,0 +1,228 @@
...
+.. feat_req:: Tracing and Profiling
+   :id: feat_req__orchestration__tracing_and
+   :reqtype: Functional
+   :security: NO
+   :safety: QM
+   :satisfies: stkh_req__execution_model__processes, stkh_req__dev_experience__tracing_of_exec
+   :status: invalid
+
+   The system **SHALL** provide hooks for tracing and profiling task execution to verify behavior and control flow of the integrated system.
```
armin-acn, line 120: The tracing and logging frameworks already provide means for this.

docs/features/orchestration/requirements.rst
```diff
@@ -0,0 +1,228 @@
...
+
+.. feat_req:: Trusted Intra Process Code Assumption
+   :id: feat_req__orchestration__trusted_intra
+   :reqtype: Non-Functional
+   :security: YES
+   :safety: QM
+   :satisfies: stkh_req__execution_model__processes, stkh_req__dependability__security_features
+   :status: invalid
+
+   The orchestration feature **SHALL** assume that all code executing within a process is trusted.
```
armin-acn, line 136: I think this is not a requirement to the orchestrator but rather a requirement on how developers shall partition their applications (i.e., independent of the existence of the orchestration feature).

docs/features/orchestration/requirements.rst
```diff
@@ -0,0 +1,228 @@
...
+
+.. feat_req:: Homogeneous ASIL Levels Within a Process
+   :id: feat_req__orchestration__homogeneous_asil
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: stkh_req__execution_model__processes, stkh_req__dependability__automotive_safety
+   :status: invalid
+
+   All tasks within a single process **SHALL** share the same ASIL level.
```
armin-acn, line 168: I think this is not a requirement to the orchestrator but rather a requirement on how developers shall partition their applications (i.e., independent of the existence of the orchestration feature).

docs/features/orchestration/requirements.rst
```diff
@@ -0,0 +1,228 @@
...
+
+.. feat_req:: Priority Based Preemption Between Thread Pools
+   :id: feat_req__orchestration__priority_based
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: stkh_req__execution_model__processes, stkh_req__dependability__automotive_safety
+   :status: invalid
+
+   The system **SHALL** implement priority-based preemption between thread pools to ensure that lower-priority programs cannot interfere with higher-priority programs.
```
armin-acn, line 224: Is it intended to provide means for that in the orchestration feature? Or is there a separate mechanism for that (e.g. based on OS tools)?

docs/features/orchestration/index.rst
```diff
@@ -0,0 +1,261 @@
...
+-----------------------------------------------------------------------------------------
+
+Concurrent programming in our target environment spans multiple scopes. At one level, concurrency exists within the algorithms of individual applications or system services, while at another level, multiple applications must execute concurrently across the platform. The challenge is to offer an interface that is not only simple and expressive but also deterministic and reliable - an important requirement in safety-critical systems.
+
+Traditional thread-based concurrency in POSIX-like environments introduces complexities such as deadlocks, livelocks, and starvation. These issues, coupled with the inherent difficulties in debugging and validating thread-based systems, can compromise both performance and reliability. [#f4]_ [#f5]_ [#f6]_ Moreover, current designs often separate the management of timing requirements, monitoring, and error handling from the control flow. Integrating these aspects closer to the application logic would promote higher cohesion and lower coupling, enabling more effective debugging and validation, particularly when addressing application-specific scenarios.
```
mthiede-acn2, line 72: Can you explain how the orchestrator helps to avoid dead locks, livelocks and starvation? In particular also in the case where priority based OS scheduling is used as described below.

docs/features/orchestration/index.rst
```diff
@@ -0,0 +1,261 @@
...
+
+Currently, application developers are expected to manage runtime scheduling directly via the operating system, controlling thread lifecycles, priorities, and affinities. Such an approach ties the configuration too closely to a specific deployment scenario and often leads to discrepancies between behaviors observed during development and those on the target system. This misalignment complicates integration efforts, as integrators must repeatedly iterate with the application developer to meet the system's reliability requirements. An orchestrator that abstracts these complexities could alleviate these challenges by offering a uniform, deterministic interface for managing runtime scheduling, thus ensuring that lower-criticality applications do not adversely affect the performance or reliability of higher-criticality ones.
```
mthiede-acn2, line 80: Can you explain how the Orchestrator helps to integrate multiple applications and other services? For example, what does an integrator have to do when adding another orchestrator-enabled application to an existing system? I assume such an application comes with a description according to the "nested task-based programming framework". How will this new description be merged into the existing descriptions? How can it be ensured that the existing applications will still meet their timing requirements?

mthiede-acn2, line 80: Besides the potential upsides of the approach of an algorithm independent description of scheduling requirements, there is the downside, that the application logic is split into two parts: The part in the algorithm and the part in the scheduling description. I think this does not improve user friendliness from the perspective of an application developer. Therefore, the pros and cons have to be weighted carefully against each other.

docs/features/orchestration/index.rst
```diff
@@ -0,0 +1,261 @@
...
+------------------------------------------------------------------------------------------------
+
+To harness the benefits of user-space multi-tasking while still providing a user-friendly and deterministic interface for concurrent programs, this proposal advocates for a nested task-based programming framework. The choice for a nested structure over a graph-based one is driven by the need to design reliable programs and enable straightforward control flow analysis - a requirement that becomes critical during safety inspections. Although graph-based structures may have a gentler learning curve and offer rapid initial results, they often become limiting when more complex scheduling descriptions are needed, such as conditional branching, time monitoring, and error handling paths.
+
+The envisioned programming framework will be:
+   - Free from explicit thread management.
+   - Free from complex synchronization mechanisms.
+   - Capable of expressing both concurrent and sequential dependencies.
+   - Capable of expressing conditional branching within the program.
+   - Capable of expressing timing constraints and error handling paths directly within the program.
```
mthiede-acn2, line 125: Could you please explain what timing constraints you envision and what exactly the error handling paths would do? Is the error handling the reaction to a failed timing constraint check or is this error handling something more general? How does user code interact with the error handling?

docs/features/orchestration/index.rst
```diff
@@ -0,0 +1,261 @@
...
+
+In existing platforms for microprocessors (µP), each application is expected to interact with approximately 15 system services or daemons - such as ``ara::diag`` for diagnostics and ``ara::com`` for SOME/IP communication. Under a straightforward implementation, this interaction model results in the creation of around 15 threads per application. When scaled to 100-150 applications, this amounts to roughly 1500 to 2250 threads solely managing inter-process communication, excluding those performing the core application tasks.
+
+Given that the HPC's µP typically provides between 2 and 16 cores, only a limited number of threads can be processed in parallel. In POSIX-like operating systems, threads become the currency for concurrency, meaning that when the thread count far exceeds available cores, the system must rely on context switching. With context switching times estimated between 2µs and 4.5µs [#f1]_ [#f2]_ [#f3]_, even a 100ms time slice could spend between 3% and 10% of its duration on context switching alone - assuming each thread is scheduled once. This overhead increases dramatically if threads are forced to switch more frequently due to competition for processing time.
+
```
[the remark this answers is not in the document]
qor-lb: Yes correctly, an async executor like e.g. tokio-rs is one possible solution to implement the functionality (and in fact this is also the direction we went). I wanted to focus however on the functionality that we intend to achieve and leave the implementation details for the feature architecture/component decomposition. Other possible solutions besides an async executor could also be the introduction of "green-threads" like e.g. in the [fiber-framework](https://code.google.com/archive/p/fiber-framework/) of Google.

I also thought of introducing two feature requests, one for the cooperative executor and one for the higher level orchestration, however, besides the standard async executor functionality we also need to take into account e.g. potential "special" tasks that can preempt or escalate on errors in a cooperative context. These requirements come become only visible if the executor is designed with the context of the higher level orchestration in mind. This is the reason why I decided to combine the "executor" and "orchestration" functionalities in one FR. My expectation is that during the feature architecture we will decompose the functionality accordingly and end up with two components.

docs/features/orchestration/requirements.rst
qor-lb: Yes correct, logging and tracing provide the means to log and trace, however, the specific place and information that has to be traced can only be known by the orchestration feature. The goal is that this is actually implemented so that abstractions on top of the OS threads can be traced in accordance to the traces exposed by the system kernel.

PS: see the [answer on martin's comment](https://github.com/eclipse-score/score/pull/857/files#r2042466163) for more context

docs/features/orchestration/requirements.rst
qor-lb: The goal of this requirement is to define the assumed context for the implementation/design of the functionality. This is intended to be the input to the later required security analysis. What is your opinion? How/where should we define these "assumed" safety/security contexts?

docs/features/orchestration/requirements.rst
qor-lb: The goal of this requirement is to define the assumed context for the implementation/design of the functionality. This is intended to be the input to the later required safety analysis. What is your opinion? How/where should we define these "assumed" safety/security contexts?

docs/features/orchestration/requirements.rst
qor-lb: The goal is to utilize preemption to achieve that lower priority programs cannot interfere with higher-priority programs. Ideally this is achieve by using the means of the OS, i.e. prioritizing threads in the thread pool.

docs/features/orchestration/index.rst
```diff
@@ -0,0 +1,261 @@
...
+============
+
+To activate this feature, use the following feature flag:
+
+``experimental_orchestration``
+
+
+Abstract
+========
+
```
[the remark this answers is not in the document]
qor-lb: Short answer: both directions.

The goal is to provide user-level scheduling and configuration of tasks and threads within a single process. Additionally to that it should provide the building blocks that enable you to orchestrate the functionality within the ECU and across processes. Because this targets only the building blocks it should not force the user to do this in a fully centralized nor fully decentralized approach. 

The reason for why I would like to combine these functionalities in one PR are elaborated in the answer to [your comment below](https://github.com/eclipse-score/score/pull/857/files#r2040168457). I would propose a quick breakout session so that we can align how to proceed with the FR.

docs/features/orchestration/index.rst
```diff
@@ -0,0 +1,261 @@
...
+This can be achieved by leveraging the proposed programming framework to express application structures without direct dependence on kernel threads. Integrators would then define a shared thread pool that executes the routines of these programs via cooperative multi-tasking. This design allows integrators to manage the number, affinity, and priorities of threads in the pool, providing full control over system resources. Consequently, integrators can assign computing resources during deployment without the need for iterative fine-tuning with each application developer.
+
+Additionally, the system will use priority-based preemption between thread pools to ensure that lower-priority programs cannot interfere with higher-priority ones. Error handling is managed independently of the application's normal execution paths, with misbehaving code preemptively handled to maintain overall system stability. This mechanism empowers developers to build robust safety mechanisms on top of the orchestration feature.
```
simon-acc, line 143: Could you clarify what you envision here? Launching threads in thread pools with different OS priorities / in different cgroups?

docs/features/orchestration/index.rst
simon-acc: Hm what would the system do if deadlines are passed / errors occur in the "reaction branches" to other passed deadlines / errors? I could imagine that this introduces a good amount of complexity. And application developers would need to learn about it to implement the right reaction branches - but compared to OS-level scheduling, it is new, unknown and potentially not as mature as OS-level scheduling.

docs/features/orchestration/index.rst
simon-acc: > Currently, application developers are expected to manage runtime scheduling directly via the operating system, controlling thread lifecycles, priorities, and affinities.

This also passes the necessary tools to developers using a well-matured, time-tested scheduling approach which they already know. I could imagine that the effort both for application developers and integrators is increased if they have to learn a new, restricted scheduling approach instead of using what they know and what has been optimized over time.

> This misalignment complicates integration efforts, as integrators must repeatedly iterate with the application developer to meet the system's reliability requirements.

I do not think that the iteration for system reliability requirements disappears with an added orchestration layer. Threads and thread pools of the orchestration feature would still interfere with each other and with other processes and threads on the OS. The orchestration feature cannot offer guarantees like e.g. a realtime OS kernel when it comes to scheduling.

docs/features/orchestration/index.rst
simon-acc: > Traditional thread-based concurrency in POSIX-like environments introduces complexities such as deadlocks, livelocks, and starvation. 

Reading the rest of your FR, it does not sound like the proposed feature would tackle those concurrency challenges. Instead, it would introduce a new way how to react to timeouts and errors (which themselves might timeout and error...). Could you explain how this actually **reduces** the complexity for developers, instead of replacing a well-known set of concurrency challenges with a new, unknown set?

docs/features/orchestration/index.rst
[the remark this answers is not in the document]
simon-acc: The scheduling in operating systems has been optimized heavily. There are a few mayor motivations for task-based concurrency which share threads (like `async` in Rust):

1. Avoiding the overhead of **spawning** and **reaping** threads (e.g. not spawning and reaping one thread per connection when serving thousands of connections per second).
2. Minimizing load in IO-bound workloads (if most of what the code does is network IO, we can save thread switches because they do not pale in comparison)
3. Allowing imperative-style programming for concurrently executing code (every `await` is a potential hand-off to another execution path).

I do not see how our expected use-cases could harness those advantages. We will not have webserver-like applications which accept thousands of new connections every second and most applications are either heavier on the compute than the IO part or they do IO continuously without switching between different IO sources all the time. Furthermore, the nested language which you propose does not bring the benefit of easy imperative-style programming for concurrent code as `Future`s in Rust or e.g. `Promise`s in JS.

docs/features/orchestration/requirements.rst
```diff
@@ -0,0 +1,228 @@
...
+
+.. feat_req:: Configurable Error Handling Mechanisms
+   :id: feat_req__orchestration__configurable_error
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: stkh_req__execution_model__processes, stkh_req__dependability__automotive_safety
+   :status: invalid
+
+   The orchestration feature **SHALL** include configurable error handling mechanisms that are insulated from the effects of misbehaving tasks (e.g., tasks in an infinite loop).
```
ramceb, line 182: I see here a problem that misbehaving in this case is limited to temporal misbehaving. As far as I understood there is no mechanism to protect tasks from spatial misbehaving of other tasks

docs/features/orchestration/index.rst
```diff
@@ -0,0 +1,261 @@
...
+
+The solution includes the following key functionalities:
+   - User-level scheduling to efficiently handle task switches in the nanosecond range.
+   - Cooperative multi-tasking, allowing multiple concurrent tasks to use the same OS thread.
+   - Provision of a configurable thread pool to enable parallelism for concurrent tasks.
+   - Introduction of additional thread pools only when necessary, such as when tasks differ in criticality or require separation by process boundaries.
```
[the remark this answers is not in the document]
qor-lb: @qor-lb add example

Issue #709, opened 2025-03-17
Improvement: Document (Cyber)-Security process


aschemmel-tech:
Document in score repository (until the process_description repo is available)

Create documentation according to "template": https://github.com/eclipse-score/process_description/blob/main/requirements_engineering/index.rst

based on standard IsoSae 21434 and S-CORE tailoring

this includes workproducts definition and linking to standard
this includes guidance definition and linking to standard
Prepare presentation in audit session

PandaeDo:
Epic (initial) Security Management is done.

Issue #696, opened 2025-03-17
Improvement: Document Tailoring of ISOSAE21434


masc2023:
Tailoring for the relevant clauses of the ISOSAE 21434 shall be done as part of the platform management plan, security management plan.

Consider als communication from VDA,
that use of clause/sub-clauses numbers and identification of work products in Open Source is possible without asking for approval

PR #867, opened 2025-04-07
process: initial security management process

masc2023:
Resolves: #709, #696


docs/process/_assets/score_process_area_overview.drawio.svg
pahmann, line 1: Is there any reasoning behind the re-ordering of process areas? (I am just asking, it is not a finding to change this.)

masc2023: Just some thinking in Project Management activities, including Safety/Security and Change Request first, then Problem and finally more Infrastructure Config, Tool, Release, but open to discuss different orders

PR #889, opened 2025-04-10
tooling: Correct Pyright new warnings

Aymen-Soussi-01:
Fix all  reportPrivateUsage  and reportMissingParameterType, and more.

More than 120 Warnings in total have been fixed in this PR. 


AlexanderLanin (approved):

AlexanderLanin (dismissed):
> AttributeError: <module 'score_header_service.header_service' from '/home/runner/.cache/bazel/_bazel_runner/8873f4afacf2293fa7b0b8ed28a49529/sandbox/processwrapper-sandbox/10/execroot/_main/bazel-out/k8-fastbuild/bin/docs/header_service_test.runfiles/_main/docs/_tooling/extensions/score_header_service/header_service.py'> does not have the attribute '_append_approver_teams'

docs/_tooling/extensions/score_header_service/header_service.py
```diff
@@ -49,7 +49,7 @@
 #     }
 
 
-def register(app: Sphinx, env: BuildEnvironment, _: str) -> None:
+def register(app: Sphinx, env: BuildEnvironment, _: str | None) -> None:
```
MaximilianSoerenPollak, line 52: Why is this str/None?

docs/_tooling/extensions/score_header_service/header_service.py
```diff
@@ -251,7 +253,7 @@ def _extract_approvers(
     approvers_set: set[str] = {
         review.user.login for review in reviews if review.state == "APPROVED"
     }
-    approvers_list: list[str] = _append_approver_teams(list(approvers_set), team_info)
+    approvers_list: list[str] = append_approver_teams(list(approvers_set), team_info)
```
MaximilianSoerenPollak, line 256: This should probably be a private function.

docs/_tooling/extensions/score_header_service/test/test_header_service.py
```diff
@@ -11,13 +11,15 @@
...
 from unittest.mock import ANY, MagicMock, patch
-import score_header_service.header_service as hs
+
 import pytest
+import score_header_service.header_service as hs
 
 
 @pytest.fixture(scope="session", autouse=True)
-def add_metadata(record_testsuite_property):
+def add_metadata(record_testsuite_property: Callable[[str, str | list[str]], None]):
```
MaximilianSoerenPollak, line 22: Currently correct, theoretically list[str], tuple[str], set[str] and str are allowed here in the 2nd param. But I think we can leave it as is

docs/_tooling/extensions/score_header_service/test/test_header_service.py
```diff
@@ -157,38 +164,38 @@ def test_extract_merge_commit_data(run_mock):
...
 
 
-@patch("score_header_service.header_service._extract_approvers")
-@patch("score_header_service.header_service._extract_reviewers")
-@patch("score_header_service.header_service._extract_team_info")
-@patch("score_header_service.header_service._extract_org")
-@patch("score_header_service.header_service._extract_repo")
-@patch("score_header_service.header_service._extract_pull_request")
-@patch("score_header_service.header_service._extract_github_token")
+@patch("score_header_service.header_service.extract_approvers")
```
MaximilianSoerenPollak, line 180: we shouldn't make private/protected functions public, if they were protected before we should make them private maybe, or ignore the warning. 
But making stuff public isn't ideal as it was private assumingly for a reason

docs/_tooling/extensions/score_header_service/test/test_header_service.py
```diff
@@ -313,7 +321,9 @@ def test_extract_reviewers():
     mock_review1.user.login = "reviewer1"
     mock_review2.user.login = "reviewer2"
     mock_review3.user.login = "reviewer3"
-    assert (
-        hs._extract_reviewers([mock_review1, mock_review2, mock_review3], "author1")
-        == "reviewer1; reviewer3"
+    mock_paginated_list = MagicMock()
+    mock_paginated_list.__iter__.return_value = iter(
+        [mock_review1, mock_review2, mock_review3]
```
MaximilianSoerenPollak, line 326: What does this do exactly?

docs/_tooling/extensions/score_header_service/test/test_header_service_integration.py
```diff
@@ -167,54 +177,50 @@ def test_header_service_integration_github_data(
...
     }
     mock_generate_hash.return_value = "abcdef"
     mock_extract_merge_commit_data.return_value = github_data
     mock_basic_conf = basic_conf(use_github_data=False)
-    app = sphinx_app_setup(
-        mock_basic_conf,
-        basic_needs,
-        template_needs,
+    app: SphinxTestApp = sphinx_app_setup(
+        mock_basic_conf, basic_needs, template_needs, None
```
MaximilianSoerenPollak, line 215: Don't think it makes sense to explicitly pass None here?

docs/_tooling/extensions/score_header_service/test/test_header_service_integration.py
```diff
@@ -167,54 +177,50 @@ def test_header_service_integration_github_data(
...
     )
     try:
         app.build()
-        html_content = (app.outdir / "index.html").read_text()
+        html_content: str = (app.outdir / "index.html").read_text()
         mock_extract_merge_commit_data.assert_called_once()
         assert "John Doe" in html_content
-        assert "aprover_1" in html_content
+        assert "approver_1" in html_content
```
MaximilianSoerenPollak, line 222: This was changed to 'approver1' inside the tests further up, so this won't work

docs/_tooling/extensions/score_layout/__init__.py
```diff
@@ -24,7 +26,7 @@ def setup(app: Sphinx) -> dict[str, str | bool]:
     }
 
 
-def update_config(app: Sphinx, _config):
+def update_config(app: Sphinx, _config: Any):
```
MaximilianSoerenPollak, line 29: Config is a dictionary that has 'dict[str,Any]' but I guess this is okay for rn

docs/_tooling/extensions/score_header_service/header_service.py
Aymen-Soussi-01: Fixed. Please Resolve.

docs/_tooling/extensions/score_header_service/test/test_header_service.py
Aymen-Soussi-01: ok

docs/_tooling/extensions/score_header_service/header_service.py
Aymen-Soussi-01: The third parameter is not used, but I guess it can be either a string either a None, as we used to call the function right now in other places, and the string is in case we will modify the function in the future, because it doesn't make sense to leave the parameter type to only None

docs/_tooling/extensions/score_header_service/test/test_header_service_integration.py
Aymen-Soussi-01: Fixed. Please resolve.

docs/_tooling/extensions/score_layout/__init__.py
Aymen-Soussi-01: ok

docs/_tooling/extensions/score_header_service/test/test_header_service.py
Aymen-Soussi-01: here we modified the code to solve 3 problems:
- Mocked PaginatedList instead of using a raw list to simulate the real input type expected by the function
- Used MagicMock with __iter__ |to make the mock iterable like the real object
- Solved potential incompatibility to avoid issues if the function does type checking or uses other list-like behavior

docs/_tooling/extensions/score_header_service/header_service.py
MaximilianSoerenPollak: Technically this is never suppose to be called directly, the 3rd parameter is there as the 'app.connect' form setup with that 'event' gives 3 parameters. Sphinx, BuildEnviroment, Docname(str). 
So we have to accept the 3rd param but we assigned it to '_' as we wanted to signal we do not use it. 
I guess we can keep it as is for now as nobody should use it anyway.

Aymen-Soussi-01: ok

docs/_tooling/extensions/score_header_service/test/test_header_service.py
AlexanderLanin: Having _extract_reviewers accept an Iterable might be easier. But it's fine either way.

MaximilianSoerenPollak (approved):
Looks alright for now, with continuous improvements I think we are on a good way.

Issue #736, opened 2025-03-19
POC for FileChecker


AlexanderLanin:
Discussion: https://github.com/eclipse-score/score/blob/FileChecker/docs/_tooling/extensions/score_metamodel/decision_records/FileChecker.md

Blocks: #553

AlexanderLanin:
# Idea: Define tests for the metamodel in .rst files instead of in python files

Instead of writing lots and lots of unit tests, we could define the tests in a more human-readable format.
This would allow us to:
* Run this file through sphinx and check the output
* Run this file as a unit test replacement
* Allow process experts to read the tests
* Allow process experts to write the tests!!

Scope:
* Testing warnings (or absence of warnings) in the sphinx output
* Nothing else!!

## References

* [FileCheck](https://llvm.org/docs/CommandGuide/FileCheck.html)
* [Example of FileCheck](https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/test/clang-tidy/checkers/bugprone/chained-comparison.cpp)


## Comparison

### Current approach

```python
def test_check_linkage_status_positive(self):
    logger = tests.fake_check_logger()
    app = Mock(spec=Sphinx)

    need_1 = NeedsInfoType(
        id="TOOL_REQ__1",
        status="valid",
        satisfies=[
            "feat_req__2",
        ],
    )

    need_2 = NeedsInfoType(
        id="feat_req__2",
        status="valid",
    )
    needs = [need_1, need_2]

    check_linkage_status(app, needs, logger)
    logger.assert_no_warnings()
```

Obviously we can optimize the above code a little bit, but the idea is to have a human-readable format that is easy to understand.
The above code will never be readable or even writeable by a non-programmer.


### Proposal

Hint: `#EXPECT` always applies to the next line without an `#` prefix.

```rst
#EXPECT-NOT: id is not lower case
.. stkh_req:: parent need
   :id: feat_req__2
   :status: draft

#EXPECT: id is not lower case
#EXPECT: has a parent requirement(s): `feat_req__2` with an invalid status.
.. stkh_req:: need for testing
   :id: TOOL_REQ__1
   :status: valid
   :satisfies: feat_req__2

-------------------------

no technical assurance that test cases don't interfere with each other!

#EXPECT: status is missing
.. stkh_req:: another need
   :id: TOOL_REQ__2

---

#EXPECT-NOT: status is missing
.. stkh_req:: another need
   :id: TOOL_REQ__3

.. needextend::
[...]

PR #933, opened 2025-04-15
tooling: Add docu and change matching function

mmr1909:
Add documentation for the file based rule checks.
Test function checks file name and line number
separately from warning message now.

Fixes: https://github.com/eclipse-score/score/issues/736


MaximilianSoerenPollak (approved):
Looked through in call. 
Looks good, thanks for the work

PR #934, opened 2025-04-15
docs: include doxygen

hoe-jo:


hoe-jo:
obsolete since tooling moved to other repository

Issue #948, opened 2025-04-16
Improvement: Add overviews for roles, workflows in process description


masc2023:
Overview list for work products exists already.

DoD:
Overview list for roles and workflows are added

PR #949, opened 2025-04-16
process: add overviews for roles and workflows

masc2023:
update process overview color code to latest status

Resolves: #948


aschemmel-tech (approved):
fine

Issue #735, opened 2025-03-19
Provide non-fatal metamodel checks


AlexanderLanin:
Whenever we introduce a new check we get warnings from rst files, and cannot merge it without fixing the rst fils.

Ideas so far:
* (disable checks via DISABLED_ prefix in the check name. (branch exists somewhere... -> @AlexanderLanin))
* provide a log.non_fatal_warning() function which will print the warning, but not count it as a "warning as error"

AlexanderLanin:
@Aymen-Soussi-01 can you look into this?

PR #976, opened 2025-04-22
tooling: Log errors related to new checks as Infos

Aymen-Soussi-01:
Implement a new function that logs errors related to new checks as Info's and not normal warnings, with a message requesting to fix them before the new project version release.

A request for all team members to understand the new implementation of the log function and use it correctly whenever you introduce new checks in the project  

close: #735


AlexanderLanin:
I'm not 100% sure what we actually need. So this is good enough, until we have a real life use case!

AlexanderLanin (approved):

AlexanderLanin:
FYI: eclipse/eca infrastructure is down. Hopefully we can merge this tomorrow.

PR #986, opened 2025-04-25
tooling: Enable and document Python coverage

mmr1909:
Enable Python coverage in MODULE.bazel
Document how to generate coverage reports

Fixes: https://github.com/eclipse-score/tooling/issues/23


MaximilianSoerenPollak (approved):
Thanks. Once we have more coverage language support we might move the README part. But this looks good for now .

Issue #920, opened 2025-04-14
add starpls to tooling & use it in score, template & process repos


AlexanderLanin:
Note: We might also need buildifier?

MaximilianSoerenPollak:
The PR / jumping off point is here: https://github.com/eclipse-score/tooling/pull/14

PR #1037, opened 2025-05-06
tooling: Add starpls from tooling

nicu1989:
Replace in-score starpls tool with the one from tooling repo.

Related: [#920](https://github.com/eclipse-score/score/issues/920)


MaximilianSoerenPollak (approved):
Thanks for this. We are moving slowly to use all our tooling. Glad to see it.

PR #1042, opened 2025-05-06
bugfix: correct needs link to workflow to enable documentation build

masc2023:
Two overlapping merges conflicts in workflow definition, this bugfixes resolves the topic


aschemmel-tech (approved):
ok

PR #1045, opened 2025-05-06
bugfix: remove irrelevant folder tool concept

masc2023:
Tool Management is now defined, thus irrelevant Tool Evaluation Concept can be removed Corrected references


aschemmel-tech (approved):
ok

Issue #1039, opened 2025-05-06
Improvement: Remove term "Contribution Request"


aschemmel-tech:
> [!IMPORTANT]
> Make sure to link this issue with the PR for your improvement.

The term "Contribution Request" was discussed originally in context of the "Feature Request". The correct term according to the change mangement process is "Change Request" which has several types (feature request, feature modification, component request, component modification).
But in the documentation https://eclipse-score.github.io/score/main/contribute/index.html there is still a "Contribution Request Guideline" - the term should be replaced by  "Contribution Guideline"

PR #1048, opened 2025-05-07
process: remove term contribution request

masc2023:
Correct term is now Change Request with types
Feature request, modification, et.

Resolves: #1039


aschemmel-tech (approved):
ok, did not know it was so widespread

PR #1099, opened 2025-05-16
time feature request

qor-lb:


docs/features/time/index.rst
```diff
@@ -0,0 +1,138 @@
...
+.. _time_feature:
+
+Time
+#############
+
+.. document:: Time
+   :id: doc__time
+   :status: valid
+   :safety: ASIL_B
+   :tags: change_management
```
masc2023, line 24: feature_request

docs/features/time/index.rst
```diff
@@ -0,0 +1,138 @@
...
+
+
+Motivation
+==========
+
+Efficient and accurate time management is critical in automotive systems, affecting functionality, safety, and reliability across various applications. Generally, time relevance can be categorized into the following levels:
+
+* External Time Synchronization
+* In-Vehicle Time Synchronization
+* Clocks, Accuracy, and Reading Current Time
```
armin-acn, line 52: Could this item be defined a bit more clearly? E.g.

* Access to clocks including reading of current time
* Supporting calculus with points in time and time differences 
* Determining accuracy of time read from each clock

docs/features/time/index.rst
```diff
@@ -0,0 +1,138 @@
...
+---------------------------------------------
+
+Software stacks contain multiple clock sources:
+
+- Local Clocks: These include monotonic, system, steady, and high-resolution clocks, providing internal time references independent of external synchronization.
+- Synchronized Clocks: These clocks are synchronized to external or internal standards and are critical for tasks requiring accurate timestamps or coordinated execution.
+- Secure Clocks: Used specifically in cryptographic contexts to validate digital certificates or ensure secure transaction timestamps; these require protection against tampering and robust synchronization.
+
+Programming languages typically abstract these clocks (e.g., ``std::chrono`` in C++, ``std::time`` in Rust). Therefore, the integration of S-CORE with existing language-specific abstractions is essential for consistency and ease of use.
+
```
armin-acn, line 85: The integration should also enable calculus with points in time and time differences.
And it should be possible to serialize points in time and differences, so that they can be communicated with data via mw::com as well as be used during replay.

docs/requirements/stakeholder/index.rst
```diff
@@ -557,6 +557,83 @@ Communication
...
+
+.. stkh_req:: Synchronize the HW clock with Vehicle Time
+   :id: stkh_req__time__hw_clock_sync
+   :reqtype: Functional
+   :security: NO
+   :safety: QM
+   :rationale: Enables the system to compare events from different ECUs chronologically, using the same time base for timestamping ingress and egress frames.
+   :status: valid
+
+   The software platform shall synchronize the local HW clock to vehicle time.
```
PhilipPartsch, line 592: I'm think we do not have access to the HW clock.
I belive we should synchronize the OS clock.

johannes-esr:
@qor-lb, can you please have a look on the pending review comments?

docs/requirements/stakeholder/index.rst
qor-lb: You are correct. Currently we do not have a system defined in S-CORE, which is why we cannot define specific requirements to the HW integration. The decision to synchronize the HW clock and use the timestamp for ingress/egress ethernet frames can only be taken at system level. I have therefore removed the requirements.

FScholPer (approved):

PR #1111, opened 2025-05-20
add the template as a starting point

jaakkomaho:


jaakkomaho:
reopen when the concept is more digested

PR #1115, opened 2025-05-20
zorancutura_some_ip_gateway

ZoranCutura:
Draft for Feature Request for SOME/IP Gateway.

This draft needs extensive rework and will be worked on throughout the next weeks.

### DIFF
--- a/docs/features/communication/some-ip-gateway/architecture/index.rst
+++ b/docs/features/communication/some-ip-gateway/architecture/index.rst
@@ -1,0 +1,23 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2024 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+Architecture
+#############
+
+.. note::
+   For now we store the component architecture in the feature tree, because multi-repo docs are not yet supported.
+   Once this support becomes available the component architecture will be moved to the module.
+
+.. toctree::
+   :titlesonly:
--- a/docs/features/communication/some-ip-gateway/index.rst
+++ b/docs/features/communication/some-ip-gateway/index.rst
@@ -12,13 +12,13 @@
    # SPDX-License-Identifier: Apache-2.0
    # *******************************************************************************
 
-.. _ipc_feature:
+.. _some_ip_gateway_feature:
 
 SOME/IP-Gateway
 ###########################
 
 .. document:: SOME/IP-Gateway
-   :id:
+   :id: _some_ip_gateway_feature
    :status: valid
    :safety: ASIL_B
    :tags: contribution_request, feature_request
@@ -42,7 +42,7 @@ To activate this feature, use the following feature flag:
 Abstract
 ========
 
-his contribution describes how data-exchange that is outside the scope of internal communication (IPC) shall be handled in modules that service data-input and data-output to a platform.
+This contribution describes how data-exchange that is outside the scope of internal communication (IPC) shall be handled in modules that service data-input and data-output to a platform.
 Services handling data in this context can be considered as gateways.
 The focus is on a gateway to handle SOME/IP communication with external devices or counterparts, therefore this feature request is called: SOME/IP-Gateway
 
@@ -61,24 +61,23 @@ For data-exchange between applications the IPC feature is providing high-speed c
 the S-CORE platform, services are required that will handle protocols for communication with both side. This is for instance the case when communication with rest of vehicle or sensors
 needs to be realized with the SOME/IP protocol.
 
-For application developers it should be
+For software component developers it should be unrecognized that data is originated from a SOME/IP communication channel, the data should be provided with the same API as in IPC.
+Nonetheless integrators and architects will have to configure the system to receive or send data over SOME/IP, hence provide it as a SOME/IP service.
 
-IPC Security Goals
-------------------
+SOME/IP Gateway Security Goals
+------------------------------
 
-The security approach of the IPC binding achieves the security goals:
+The security approach for SOME/IP shall achieve the following security goals:
 
-- confidentiality (:need:`feat_req__ipc__confidentiality`)
-- integrity (:need:`feat_req__ipc__integrity`)
-- availability (per criticality-level) (:need:`feat_req__ipc__acl_placement`)
+- to be defined
 
 Safety Impact
 =============
 
-The safety impact was already exhaustively covered in :ref:`mot_mixed_criticality` and :ref:`spec_mixed_criticality`.
+  # The safety impact was already exhaustively covered in :ref:`mot_mixed_criticality` and :ref:`spec_mixed_criticality`.
 
-Overall, the communication framework supports use cases up to ASIL-B (:need:`feat_req__ipc__asil`).
-Future extension to ASIL-D use cases is feasible but not in scope for now.
+  # Overall, the communication framework supports use cases up to ASIL-B (:need:`feat_req__ipc__asil`).
+  # Future extension to ASIL-D use cases is feasible but not in scope for now.
 
 License Impact
 ==============
--- a/docs/features/communication/some-ip-gateway/index.rst
+++ b/docs/features/communication/some-ip-gateway/index.rst
@@ -1,0 +1,90 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2024 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+.. _ipc_feature:
+
+SOME/IP-Gateway
+###########################
+
+.. document:: SOME/IP-Gateway
+   :id:
+   :status: valid
+   :safety: ASIL_B
+   :tags: contribution_request, feature_request
+
+
+.. toctree::
+   :hidden:
+
+   architecture/index.rst
+   requirements/index.rst
+   roadmap.rst
+
+
+Feature flag
+============
+
+To activate this feature, use the following feature flag:
+
+``experimental_some-ip-gateway``
+
+Abstract
+========
+
+his contribution describes how data-exchange that is outside the scope of internal communication (IPC) shall be handled in modules that service data-input and data-output to a platform.
+Services handling data in this context can be considered as gateways.
+The focus is on a gateway to handle SOME/IP communication with external devices or counterparts, therefore this feature request is called: SOME/IP-Gateway
+
+This feature request includes:
+- A description of how a SOME/IP gateway service (or data broker) shall be implemented
+- How the SOME/IP gateway services shall integrate with the zero-copy communication from IPC (which might become a general description of how services plug-in to the IPC context)
+
+.. _Motivation:
+
+Motivation
+==========
+
+S-CORE is targeting high-performance automotive systems with safety impact. Applications integrated on S-CORE will be distributed across multiple processes and frameworks (like FEO)
+that schedule software components (i.e. activities) with the need to exchange data.
+For data-exchange between applications the IPC feature is providing high-speed communication capabilities, but when it comes to communication with applications outside the scope of
+the S-CORE platform, services are required that will handle protocols for communication with both side. This is for instance the case when communication with rest of vehicle or sensors
+needs to be realized with the SOME/IP protocol.
+
+For application developers it should be
+
+IPC Security Goals
+------------------
+
+The security approach of the IPC binding achieves the security goals:
+
+- confidentiality (:need:`feat_req__ipc__confidentiality`)
+- integrity (:need:`feat_req__ipc__integrity`)
+- availability (per criticality-level) (:need:`feat_req__ipc__acl_placement`)
+
+Safety Impact
+=============
+
+The safety impact was already exhaustively covered in :ref:`mot_mixed_criticality` and :ref:`spec_mixed_criticality`.
+
+Overall, the communication framework supports use cases up to ASIL-B (:need:`feat_req__ipc__asil`).
+Future extension to ASIL-D use cases is feasible but not in scope for now.
+
+License Impact
+==============
+
+[How could the copyright impacted by the license of the new contribution?]
+
+
+How to Teach This
+=================
--- a/docs/features/communication/some-ip-gateway/requirements/index.rst
+++ b/docs/features/communication/some-ip-gateway/requirements/index.rst
@@ -1,0 +1,31 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2024 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+Requirements
+############
+
+
+
+Safety Impact
+=============
+
+.. feat_req:: SOME/IP Gateway ASIL level
+   :id: feat_req__some_ip_gateway__asil
+   :reqtype: Functional
+   :security: YES
+   :safety: ASIL_B
+   :satisfies:
+   :status: valid
+
+   The SOME/IP Gateway shall support safe communication up to ASIL-B.


HartmannNico:
For starters I recommend to refer to a specific AR specification, e.g. 24/11 or 23/11 for (minimum) SOME/IP compatibility. Otherwise we could also fulfill this FR with an initialo SOME/IP 1.0 implementation which is probably not what we have in mind here.

ZoranCutura:
> For starters I recommend to refer to a specific AR specification, e.g. 24/11 or 23/11 for (minimum) SOME/IP compatibility. Otherwise we could also fulfill this FR with an initialo SOME/IP 1.0 implementation which is probably not what we have in mind here.

Good point. Suppose we should agree on the target version in the architecture community.

For the moment I will refer to AUTOSAR release 24-11. We may change this anytime if other version match the OEM expectations better.

ZoranCutura:
Latest version before the F2F Architecture Workshop (https://github.com/orgs/eclipse-score/discussions/1247) is available for review. Please add your comments for further discussion and improvement.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,136 @@
...
+
+.. _Specification:
+
+Specification
+=============
+
+SOME/IP Gateway protocol implementation
+---------------------------------------
+
+The protocol implementation shall be fully compatible and complying with the SOME/IP specification from AUTOSAR Adaptive.
```
lurtz: We discussed today the possibility to have an internal memory representation for shared memory, which follows the SOME/IP serialization format. The goal is to avoid or speed up the (de)serialization step without performance impact on IPC.

Right now I doubt that will work, if the full feature set of SOME/IP and AUTOSAR Adaptive is being used. These are among the things, which bother me:
- byte order is configurable
- even the size of length fields is configurable
- variable sized structures like vector, string or map can only be supported with scatter / gather IO. But how do we figure out the buffer boundaries?

ZoranCutura:
Need to add functionality for following topics:
- How to expose SOME/IP failures, like checksum, dropped messages to applications
- Service Discovery for services in Apps
- How will the Gateway be handled in lifecycle-management
- Methods in SOME/IP with mw:com / LoLa
- Requirements to Security and Interface to Crypto --> probably not in priority for the moment

Need to discuss if payload-transformation is required or can be avoided, and the concept of how payload-transformation is realized.

dasoho:
* How to expose SOME/IP failures, like checksum, dropped messages to applications
** Is std::expected a possible solution which can be generically used on mw:com user interface level to cover also E2E results?
** E2E attributes to be forwarded to applications: sequence counter, timestamp 
** to cover message loss, message repetition, message delay/stale data
** Problem: std::expected has been introduced with c++23
** on the other hand, std::expected is part of the API datatype feature request
* Service Discovery for services in Apps
** Problem: Service request (actively using "FindService" is not supported by mw:com concept (yet) due to internal realization with inodes for service discovery
** Solution idea 1: Gateway implements FindService
**** Pro: No "SOME/IP specific topics pollute mw:com
**** Con: Clients need to know whether they deal with SOME/IP clients or internal clients
** Solution idea 2: mw:com implements find service
**** It needs to look for corresponding inodes and additionally needs to forward the find service request to the gateway
* How will the Gateway be handled in lifecycle-management
** Do we really need to address lifecycle/health handling already in the feature request?
** It seems to early to identify which entities in the gateway do require which type of supervision
** This comment would potentially apply to all other feature requests?

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: As discussed in the communication cft, the topic might be resolved with the ABI compatible data types feature request, which though is a limitation / constraint. Not all types defined in SOME/IP might be usable in this aproach.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,294 @@
...
+- DTLS (= Datagram Transport Layer Security): Authentication, integrity, confidentiality of UDP packets -> Protection for multicast communication.
+- ACL (= Access Control List): A filter mechanism to ensure that only allowed SOME/IP communication can take place.
+
+Features included `Feature Request for Security & Crypto <https://github.com/eclipse-score/score/issues/905>`_ e.g cryptographic algorithms, symmetric-, asymmetric encryption, Signature functionality, Certificate management, certificate management, entropy generation, data integrity, key management, are out of scope.
+
+
+Access Control List
+-------------------
+
+An access control mechanism is part of a firewall solution, which states that only the traffic defined in the security policy is allowed onto the network and all other traffic must be silently dropped.
```
ZoranCutura, line 165: @kalu-an
Though ACL is explained nicely, I wonder if we actually need this. 
What do we consider needs to be implemented / used? Is there an expectation to use ACL in the implementation of SOME/IP GW?

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,294 @@
...
+There are several E2E (= End-to-End) profiles, which utilize various CRC routines as part of AUTOSAR E2E Protocol Specification:
+
+- CRC8 (SAEJ1850)
+- CRC8H2F (0x2F polynomial)
+- CRC16 (also known as CCITT-FALSE 16-bit CRC)
+- CRC32 (also known as IEEE 802.3 Ethernet 32-bit CRC)
+- CRC32P4 (0xF4ACFB13 polynomial)
+- CRC64 (CRC-64-ECMA)
+- CRC32_J1939 (0x6938392D polynomial) (used by Profile 76).
+
```
ZoranCutura: @kalu-an
On Feature Request level, what is the benefit of listing these profiles instead of referring to AUTOSAR specs?

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,294 @@
...
+License Impact
+==============
+
+[How could the copyright impacted by the license of the new contribution?]
+
+Since SOME/IP is a protocol, including applied E2E protection and the according profile (polynom, etc.),
+defined by AUTOSAR and published under the license of AUTOSAR, the gateway implementation shall carefully distinguish between the SOME/IP communication stack,
+the E2E protection of data, and the integration into S-CORE mw::com. Breach of foreign licenses must be avoided.
+
```
ZoranCutura, line 243: @LittleHuba @lurtz @kalu-an
ETAS internally discussed, if there is a problem with our descriptions.
We don't see any problems, as the autosar documentation is freely accessible and references to it
are no problem.

We need to avoid to describe concepts that are actually defined in AUTOSAR and accidentally not declare them as AUTOSAR concepts (i.e. by referring to the AUTOSAR specs).

We also concluded, that users should be informed, that they must make sure to follow the rules / constraints of AUTOSAR if using the feature in commercial applications.
```suggestion
Anybody using SOME/IP Gateway in a commercial application (i.e. vehicle) needs to make sure to follow the license conditions and rules of AUTOSAR.
```

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+Abstract
+========
+
+This contribution describes how data-exchange that is outside the scope of internal communication (IPC) shall be handled in modules that service data-input and data-output to a platform.
+Services handling data in this context can be considered as gateways.
+The focus is on a gateway to handle SOME/IP communication with external devices or counterparts, therefore this feature request is called: SOME/IP-Gateway
+
+This feature request includes:
+  - A description of how a SOME/IP gateway service (or data broker) shall be implemented
+  - How the SOME/IP gateway services shall integrate with the zero-copy communication from IPC (which might become a general description of how services plug-in to the IPC context)
```
LittleHuba: I would leave the part in the parentheses out. This is something which might come in the future but is irrelevant for this feature.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+==========
+
+S-CORE is targeting high-performance automotive systems with safety impact. Applications integrated on S-CORE will be distributed across multiple processes and frameworks (like FEO)
+that schedule software components (i.e. activities) with the need to exchange data.
+For data-exchange between applications the IPC feature is providing high-speed communication capabilities, but when it comes to communication with applications outside the scope of
+the S-CORE platform, services are required that will handle protocols for communication with both side. This is for instance the case when communication with rest of vehicle or sensors
+needs to be realized with the SOME/IP protocol.
+
+For software component developers it should be unrecognized that data is originated from a SOME/IP communication channel, the data should be provided with the same API as in IPC.
+Nonetheless integrators and architects will have to configure the system to receive or send data over SOME/IP, hence provide it as a SOME/IP service.
```
LittleHuba:
```suggestion
Nonetheless integrators and architects will have to configure the system to receive or send data over SOME/IP.
```

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+
+
+.. _Rationale:
+
+Rationale
+==========
+
+SOME/IP and IPC use different mechanisms to communicate on different channels. Applications integrating on S-CORE platform may have certain requirements to data-input that is not directly supported with SOME/IP and vice versa
+SOME/IP definition may have requirements that cannot directly be supported by the application providing data on IPC. Therefor it's not only a task of this gateway
+to adapt the communication, but also to translate data between the two communication channels and probably even fill data, i.e. when applications require new data that has not be received on SOME/IP or vice versa.
```
LittleHuba:
```suggestion
SOME/IP and IPC use different mechanisms and data representations to communicate. Therefor it's not only a task of this gateway
to adapt the communication, but also to transform from one representation to another. This includes changing data layout, filling in missing data, filtering traffic, error handling, service discovery and further steps.
```

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Hence the gateway on the one side should act as a participant in IPC and fulfill all requirements to this accordingly, whereas on the other side it shall participate in SOME/IP communication
+acting as a SOME/IP service fulfilling all SOME/IP requirements and defined communication. Between these two contexts developers shall be able to create signal or data mappings and
+translate the different data-types and representations of the two contexts.
```
LittleHuba: This wording only covers that we provide SOME/IP services to the network. We may also act as a client.

I would go more into the direction of seeing the gateway as a transparent bridge between IPC and SOME/IP.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+
+The SOME/IP side shall, where possible, use an existing SOME/IP stack that is fully compatible and complying with the SOME/IP specification from AUTOSAR Adaptive.
```
LittleHuba: This should not be part of a FR. It is a up to contributions to decide whether they reuse existing code or implement from scratch.

```suggestion
```

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+
+This module shall fulfill the following requirements:
+ - Multi-Binding support - :need:`feat_req__com__multi_binding_support`
+ - agnostic binding - :need:`feat_req__com__binding_agnostic_api`
```
LittleHuba:
```suggestion
 - binding agnostic API - :need:`feat_req__com__binding_agnostic_api`
```

docs/features/communication/some_ip_gateway/index.rst
LittleHuba: We consider this no longer a path that holds much benefit.
There is often no mapping from network layout of messages to the internal layout that could be traversed without deserialization.

So depicting internal datatypes as SOME/IP in memory will not reduce our deserialization efforts.

docs/features/communication/some_ip_gateway/requirements/index.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. feat_req:: SOME/IP Gateway Secure Communication
+   :id: feat_req__some_ip_gateway__secure_com
+   :reqtype: Functional
+   :security: YES
+   :safety: QM
+   :satisfies: stkh_req__communication__secure
+   :status: valid
+
+   The platform shall support secure communication.
```
LittleHuba: This is not something we can achieve in this feature.
SOME/IP does not define any kind of encryption. E2E is only about safety.

This must be achieved by a system configuration or a separate encryption feature.

docs/features/communication/some_ip_gateway/requirements/index.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. feat_req:: SOME/IP Gateway Access Control
+   :id: feat_req__some_ip_gateway__access_control
+   :reqtype: Functional
+   :security: YES
+   :safety: QM
+   :satisfies: stkh_req__dependability__security_features
+   :status: valid
+
+   The SOME/IP Gateway shall support access control to restrict access to the gateway services.
```
LittleHuba: I would not specify this separately. All requirements to the Communication feature apply also to all its subfeatures.

docs/features/communication/some_ip_gateway/requirements/index.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. feat_req:: SOME/IP Gateway QM network stack
+   :id: feat_req__some_ip_gateway__network_stack
+   :reqtype: Functional
+   :security: YES
+   :safety: ASIL_B
+   :satisfies: stkh_req__communication__safe
+   :status: valid
+
+   If SOME/IP network stacks are available as QM only.
```
LittleHuba: You can just leave this out. The requirement above is the important one. This one is more an implementation detail. If somebody wants to write this in ASIL-B quality, he is free to do so.

docs/features/communication/some_ip_gateway/requirements/index.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. feat_req:: SOME/IP Gateway ASIL level
+   :id: feat_req__some_ip_gateway__asil
+   :reqtype: Functional
+   :security: YES
+   :safety: ASIL_B
+   :satisfies: stkh_req__communication__safe
+   :status: valid
+
+   The SOME/IP Gateway shall support safe communication up to ASIL-B.
```
LittleHuba: Same here. This follows from the Communication Requirements.
Though, we should state somewhere that the Communication requirements apply to the SOME/IP gateway.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+SOME/IP Gateway Security Goals
+------------------------------
+
+As with IPC generally, the security approach for SOME/IP gateway shall achieve the following security goals:
+
+- confidentiality (:need:`feat_req__ipc__confidentiality`)
+- integrity (:need:`feat_req__ipc__integrity`)
+- availability (per criticality-level) (:need:`feat_req__ipc__availability`)
+- secure communication (:need:`feat_req__some_ip_gateway__secure_com`)
+- access control (:need:`feat_req__some_ip_gateway__access_control`)
```
LittleHuba: Half of those are not achievable by a SOME/IP gateway. IMHO, this is something that must be achieved on system level (MACSec, IPsec,...) or by explicit encryption.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+Security Impact
+===============
+
+There are multiple protocols targeting secure communication. In general a holistic security concept for a vehicle will not apply all together.
+
+.. figure:: assets/some_ip_gateway_sec_protocols.drawio.svg
+   :align: center
+   :name: some_ip_gateway_sec_protocols_
+
+   Secure communication protocols
```
LittleHuba: This is more system configuration. I don't think we can tackle this within this feature. System integration must take care of this or the security FR.

Naming these protocols makes sense here, but I would clearly state that this is not something we solve with this FR. We just want to make users aware that SOME/IP is unsecured communication by default.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+Features included `Feature Request for Security & Crypto <https://github.com/eclipse-score/score/issues/905>`_ e.g cryptographic algorithms, symmetric-, asymmetric encryption, Signature functionality, Certificate management, certificate management, entropy generation, data integrity, key management, are out of scope.
+
+
+Access Control List
+-------------------
+
+An access control mechanism is part of a firewall solution, which states that only the traffic defined in the security policy is allowed onto the network and all other traffic must be silently dropped.
+Access Control acts on OSI Layer 5-7. It shall fulfill following:
+
+- Whitelisting of SOME/IP services and methods based on IP addresses and therefore IP address authenticity.
```
LittleHuba: I would not only create an ACL for IP addresses but also for Service-/InstanceIds.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+         :Drop Packet;
+      else (No)
+         :Process Packet;
+      endif
+   endif
+
+   stop
+
+
+E2E and CRC (Informal Notes)
```
LittleHuba: This feels like too much detail. I would keep to the level where we say that E2E-Protection/-Checks based on the AUTOSAR standard is part of the feature. Then discuss that this must happen in ASIL-B context. 

Further, performing checks and protection in a centralized fashion means that all IPC communication channels between the central checker/protector and the end-user must be ASIL-B qualified.

This also means that the IPC channels must be protected against loss of samples. Corruption is not a concern since all participants are ASIL-B.

Finally, I would state that we must support E2E protection for event-based communication as well as methods. And that we need to split E2E checks between CRC and Counter (as seen in the architecture).

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+License Impact
+==============
+
+[How could the copyright impacted by the license of the new contribution?]
+
+Since SOME/IP is a protocol, including applied E2E protection and the according profile (polynom, etc.),
+defined by AUTOSAR and published under the license of AUTOSAR, the gateway implementation shall carefully distinguish between the SOME/IP communication stack,
+the E2E protection of data, and the integration into S-CORE mw::com. Breach of foreign licenses must be avoided.
+
+Anybody using SOME/IP Gateway in a commercial application (i.e. vehicle) needs to make sure to follow the license conditions and rules of AUTOSAR.
```
LittleHuba:
```suggestion
Anybody using SOME/IP Gateway needs to make sure to follow the license conditions and rules of AUTOSAR.
```

Let's make this unconditional. Otherwise, we risk legal issues.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,299 @@
...
+
+.. note::
+   by 2025-08, anybody exploiting AUTOSAR software commercially must be an AUTOSAR member.
+   Please refer to http://autosar.org/ for more information.
```
LittleHuba: I would drop that. We did our job by saying that people must follow the AUTOSAR license.

docs/features/communication/some_ip_gateway/service_discovery/index.rst
```diff
@@ -0,0 +1,180 @@
...
+   participant "Service\nvia IPC" as Service
+
+   Network -> Someipgateway: FindService
+   Someipgateway -> Service: service_discovery()
+   alt Service available
+       Someipgateway -> Network: OfferService
+   end
+   @enduml
+
+If an service is configured to be needed and no OfferService has been received yet, the SOME/IP communication stack may send a SOME/IP-SD message containing a FindService entry.
```
LittleHuba:
```suggestion
If an service is configured to be needed and no OfferService has been received yet, the SOME/IP communication stack shall send a SOME/IP-SD message containing a FindService entry.
```

Based on PRS_SOMEIPSD_00435 this is mandatory.

ZoranCutura (commented):
partially too externsive descriptions for feature request level

references to AUTOSAR should be fine.

docs/features/communication/some_ip_gateway/index.rst
kalu-an: @ZoranCutura , I´ve explicitly added CRC algorithms due to the wish of @LittleHuba. Case on a level of feature request these details are not needed we could remove them since these the ones listed are present in referenced AUTOSAR_FO_PRS_E2EProtocol.pdf specification.

ZoranCutura: Let's discuss in our call on thursday.

docs/features/communication/some_ip_gateway/index.rst
kalu-an: @ZoranCutura , I´m not an expert in AUTOSAR IP and it´s license model but being verbose here to avoid commercial issues if future will not harm. Case there is a common used wording, statement or reference on license model would reduce ambiguity even more. As said I can´t judge it

docs/features/communication/some_ip_gateway/index.rst
kalu-an: @ZoranCutura the ACL acts in on service layer of SOME/IP that´s why I´ve included it feature request for SOME/IP gateway from security perspective. Of course implementation of this feature would be comes with effort.
An iterative approach could also an option. I.e. focus on pure functionality in release 1.0 and as security features later.

LittleHuba: We don't want to gateway all services there are in IPC and SOME/IP. We would only ever want to transfer a specific list. So an ACL based on ServiceId and InstanceId and maybe even version makes a lot of sense.

Extending that with a security focused ACL based on IP addresses is a senseful next step. IMHO, we can list that as a future extension for the moment.

docs/features/communication/some_ip_gateway/index.rst
LittleHuba: Seems like we had a misunderstanding.
I'm all for mentioning that we have CRC and Counter as basic E2E concepts and that there are different profiles. Covering the individual CRCs goes too deep. See also my own review comments.

docs/features/communication/some_ip_gateway/index.rst
LittleHuba: See my own review for this as well. Pointing the license issue out is important. I just would not go into to much detail when somebody needs to check. If we get the conditions wrong, this would be a serious problem. I would just unconditionally point out that users must consider the AUTOSAR license.

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: right, wrote this in April when we where thinking about a general gateway approach.
Will remove it.

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: will change

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: General descriptions, that do not actually request that this functionality is realized. 
Discussed in the call to remove the section, except ACL, which could be realized with the GW.

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: decide to keep the ACL description as an option to secure access to SOME/IP services from the network.
Added description at the beginning as proposed by @LittleHuba

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: proposal from @LittleHuba added

docs/features/communication/some_ip_gateway/index.rst
kalu-an: @LittleHuba I´m sorry for misunderstanding. @ZoranCutura could you, if not already done remove the bullet point list with CRC algo? I´ll will be in vacation from today and back on 2025-08-18. Thanks in advance

LittleHuba: No hard feelings! Misunderstandings always involve two people :wink:
We already reworked the section in our call today. Zoran will push the edited version once he updated the other sections.

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: Adapted text, see my proposal

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/requirements/index.rst
ZoranCutura: removed. These were only placeholders, left over from copying another requirements file.

docs/features/communication/some_ip_gateway/requirements/index.rst
ZoranCutura: @LittleHuba, what would be your proposal on the somewhere?
In the description of the SOME/IP Gateway, I suppose, or did you mean in the Communication requirements?

docs/features/communication/some_ip_gateway/requirements/index.rst
ZoranCutura: removed

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+
+  * Counter, slightly different interpretation depending on profile, n/a in case the profile doesn't support it
+  * Data ID, n/a in case the profile does not support it or if the Data ID is not explicitely transmitted like in profile 22
+  * Profile identification, required to properly interpret E2E attributes like "Counter". (Could be either an "Alive Counter" or a "Sequence Counter" depending on the profile)
+  * Detailed, profile specific error code (enum)
+
+* Proposed Error Enumeration
+
+  * No E2E error
+  * CRC Error
```
LittleHuba, line 115: It might make sense to map CRC errors always to sequence errors, since we might have a corruption of service/instance id and would therefore report the error to the wrong recipient.

docs/features/communication/some_ip_gateway/requirements/index.rst
LittleHuba: Yes, in the description of the SOME/IP Gateway. I would suggest the chapter specification.

docs/features/communication/some_ip_gateway/requirements/index.rst
ZoranCutura: removed, and added a comment in the specification section.

docs/features/communication/some_ip_gateway/requirements/index.rst
ZoranCutura: done.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,251 @@
...
+
+An access control mechanism is part of a firewall solution, which states that only the traffic defined in the security policy is allowed onto the network and all other traffic must be silently dropped.
+Access Control acts on OSI Layer 5-7. It shall fulfill the following:
+
+- What SOME/IP service instances to forward
+- What IP address is allowed to offer a SOME/IP service instance
+- Configuration of ACL is done at deployment
+- Versioning of services
+- ACL shall be able to be switched on/off
+- It shall be possible to persistently log ACL drop actions
```
lurtz: In service discovery I mention the need for configuration files: https://github.com/eclipse-score/score/blob/8e297d2e8b8458b10702dc98a93cf9bc66512a22/docs/features/communication/some_ip_gateway/service_discovery/index.rst?plain=1#L29-L37

Can we merge this with Access Control Lists? But I also see that these are used in a slightly different use case.

My intention was that anything going in or out needs to be present in the configuration. Unless configured no internal IPC service is offered on the network or consumed from the network.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,251 @@
...
+------------------------------
+
+The security approach for SOME/IP gateway shall achieve the following security goals:
+
+- access control (:need:`feat_req__com__acl_per_service_instance`)
+
+The SOME/IP Gateway service instance shall be defined in the deployment configuration.
+
+- :need:`ACL Placement <feat_req__com__acl_placement>`
+- :need:`ACL per Service Instance <feat_req__com__acl_per_service_instance>`
```
lurtz: I may not understand the structure here. The requirement is already linked a few lines above.

docs/features/communication/some_ip_gateway/assets/some_ip_gateway_architecture.drawio.svg
```diff
@@ -0,0 +1,439 @@
...
+        </g>
+        <g>
+            <g transform="translate(-0.5 -0.5)">
+                <switch>
+                    <foreignObject style="overflow: visible; text-align: left;" pointer-events="none" width="100%" height="100%" requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility">
+                        <div xmlns="http://www.w3.org/1999/xhtml" style="display: flex; align-items: unsafe center; justify-content: unsafe center; width: 108px; height: 1px; padding-top: 340px; margin-left: 601px;">
+                            <div style="box-sizing: border-box; font-size: 0; text-align: center; color: #000000; ">
+                                <div style="display: inline-block; font-size: 12px; font-family: &quot;Helvetica&quot;; color: light-dark(#000000, #ffffff); line-height: 1.2; pointer-events: all; white-space: normal; word-wrap: normal; ">
+                                    SOME/IP Gateway is a single process on the operating system and may be part of other frameworks, ie. FEO
```
lurtz: Do not we split it into at least two processes due to different safety levels?

One process acts on the network and one does E2E and payload conversion.

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+---------------------------------------
+As mentioned above, not all E2E checks can be kept hidden within the gateway. For some problems it is up to the application to decide
+whether it is still valuable to access and process the data. This is in particular true for duplication or re-odering issues.
+Therefore, it is required to pass a **tupel** consisting of the **payload data** as well as **supplementary E2E metadata and error information**
+to the IPC clients to enable the client to individually judge on particular E2E issues.
+
+* The API design needs to put the payload information as well as the additional E2E metadata as close as possible together to easily enable and motivate clients to consequently check for potential errors.
+* Additional metadata and error information needs to be incorporated into regular mw::com user interface
+* Error related interface design needs to be highly optimezed for the "good case" to have an optimized support for the common IPC case
+* AoU's need to be provided to force the user to check for the (E2E-) result
```
lurtz:
```suggestion
* Assumptions of Use need to be provided to force the user to check for the (E2E-) result
```

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+
+.. figure:: ../assets/e2e_state_machine_on_client_side.drawio.svg
+   :align: center
+   :name: _e2e_state_machine_on_client_side
+
+   E2E state machine responsibility associated to IPC client
+
+**Considered Alternative**
+
+If we allocate the statemachine responsibility to the gateway the distribution of resposnibilities would look like in the following diagram
```
lurtz:
```suggestion
If we allocate the state machine responsibility to the gateway the distribution of responsibilities would look like in the following diagram
```

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+function invocations over a period. This allows it to determine a more holistic and debounced status of the communication.
+
+Purpose of the E2E State Machine:
+The primary purpose of the E2E state machine is to transform instantaneous "per-cycle" check results into a stable,
+long-term communication channel status. This aggregated status is then provided to the consuming application,
+enabling it to make informed decisions about whether the received data can be trusted and used for safety-related functions.
+
+As mentioned above, the E2E statemachine is associated to one communication channel which is in turn associated to exactly one
+individual IPC client. Therefore it is an obvious consequence, that the individiual state machine handling and state machine
+configuration is responsibility of the client and not a central responsibility of the gateway.
```
lurtz, line 151:
```suggestion
configuration is responsibility of the client and not a central responsibility of the gateway.
Even for the same service different clients may use different state machine configuration.
```

lurtz (changes requested):
only minor findings

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,251 @@
...
+Specification
+=============
+
+The requirements from Communication generally apply to the SOME/IP Gateway.
+
+
+SOME/IP protocol implementation
+-------------------------------
+
+The protocol implementation shall be fully compatible and complying with the SOME/IP specification from AUTOSAR Adaptive.
```
LittleHuba: I think this should be put as an actual requirement.

docs/features/communication/some_ip_gateway/index.rst
LittleHuba: I think merging them is possible. Since the ACL anyway defines what service instances shall be gatewayed, we can directly use that information for the SD as well.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,251 @@
...
+      endif
+   endif
+
+   stop
+
+
+End-to-End (E2E) protection with CRC and counters
+-------------------------------------------------
+
+Applications communicating over the network may have to secure data with end-to-end protection (E2E), which may involve
```
LittleHuba: The word "secure" gives the impression of this being a security mechanism. But this is only about safety and not about security at all.

```suggestion
Applications communicating over the network may have to protect data with end-to-end protection (E2E), which may involve
```

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: You'r right, we can skip this here.
I'll remove it.

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,251 @@
...
+CRC-protection and checks, and message counters.
+
+There are several E2E (= End-to-End) profiles, which utilize various CRC routines as part of AUTOSAR E2E Protocol Specification, that shall be supported with the SOME/IP Gateway.
+
+Though the SOME/IP protocol itself is likely not going to be ASIL-B compliant and have a safety consideration of QM rather,
```
LittleHuba:
```suggestion
Though the implementation of the SOME/IP protocol itself is likely not going to be ASIL-B compliant and have a safety consideration of QM rather,
```

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,251 @@
...
+         :Drop Packet;
+      else (No)
+         :Process Packet;
+      endif
+   endif
+
+   stop
+
+
+End-to-End (E2E) protection with CRC and counters
```
LittleHuba, line 214: This whole section should be moved to Safety Impact. This is not related to security.

docs/features/communication/some_ip_gateway/service_discovery/index.rst
```diff
@@ -0,0 +1,180 @@
...
+
+SOME/IP-Gateway Service Discovery
+#################################
+
+Service discovery deals with discovery and connection setup of services.
+The basic summary of service discovery is as follows:
+
+- provided / required services (including their service elements like events, methods, and fields) must be configured
+- (locally) provided services are discovered (locally) via IPC and then offered on the network by sending a SOME/IP-SD message with an OfferService entry according to configuration
+- (locally) required services may trigger the discovery on the network by sending of a SOME/IP-SD message with a FindService entry
```
LittleHuba: This is mandatory by the AUTOSAR standard.

```suggestion
- (locally) required services trigger the discovery on the network by sending of a SOME/IP-SD message with a FindService entry
```

docs/features/communication/some_ip_gateway/service_discovery/index.rst
```diff
@@ -0,0 +1,180 @@
...
+- when a SOME/IP-SD message containing a FindService entry is received, the SOME/IP gateway checks this service has been already offered locally via IPC
+- for IPC service discovery the features of lola are used by the SOME/IP gateway
```
LittleHuba:
```suggestion
- the IPC service discovery is used by the SOME/IP gateway and no parallel discovery mechanism is introduced
```

docs/features/communication/some_ip_gateway/architecture/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/service_discovery/index.rst
```diff
@@ -0,0 +1,180 @@
...
+
+Configuration
+=============
+
+The configuration tells the SOME/IP communication stack which services it should provide and which services it should require on the network.
+The configuration contains SOME/IP and SOME/IP-SD settings as well as IP interface bindings.
+Events, methods and fields are configured as well.
```
LittleHuba, line 35: Not quite sure yet whether we must configure them on this level or whether we can deduce this based on the GenericProxy concept from IPC.

docs/features/communication/some_ip_gateway/architecture/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/service_discovery/index.rst
```diff
@@ -0,0 +1,180 @@
...
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+.. _some_ip_gateway_service_discovery:
+
+SOME/IP-Gateway Service Discovery
+#################################
+
+Service discovery deals with discovery and connection setup of services.
```
LittleHuba, line 21: I think we should give a statement that the service discovery must follow the AUTOSAR SOME/IP-SD specification. This run down here is more a guide how an implementation fits into our architecture.

I don't want people confusing this as the new specification for SD.

Could you maybe just add three requirements saying that any implementation must follow AUTOSAR SOME/IP, SOME/IP-SD and E2E specification?

docs/features/communication/some_ip_gateway/service_discovery/index.rst
```diff
@@ -0,0 +1,180 @@
...
+   IPC_client -> Service: connect()
+   @enduml
+
+.. note::
+   The SOME/IP communication stack can create the service before receiving an OfferService,
+   but can only start offering it after having received an OfferService message from the network.
+   This behavior may reduce the time until the service is available for consumers, but may increase boot time.
+   Thus the decision is to create the service only after having received an OfferService message from the network.
+
+Upon reception of SOME/IP-SD message containing an StopOfferService entry, the SOME/IP communication stack stops the proxy service offered via IPC as well.
```
LittleHuba:
```suggestion
Upon reception of SOME/IP-SD message containing an StopOfferService entry, the SOME/IP communication stack stops the service offered via IPC as well.
```

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+   # *******************************************************************************
+
+.. _some_ip_gateway_architecture:
+
+SOME/IP Gateway Architecture
+############################
+
+.. note::
+   For now we store the component architecture in the feature tree, because multi-repo docs are not yet supported.
+   Once this support becomes available the component architecture will be moved to the module.
```
LittleHuba: I would argue that this is more a feature architecture.
We don't really care how the component is designed here. We just cut the feature into components.

docs/features/communication/some_ip_gateway/architecture/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/service_discovery/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/service_discovery/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/service_discovery/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: moved

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: I still leave it also here, but add the requirement and refer to it.

docs/features/communication/some_ip_gateway/architecture/index.rst
ZoranCutura: So you're proposing to remove the note, as it doesn't apply to the architecture we are showing for this FR?
Good for me.

docs/features/communication/some_ip_gateway/assets/some_ip_gateway_architecture.drawio.svg
ZoranCutura: That's right, but is shown in the detailed architecture.
Finally it's left to the implementation how to handle that.
Hence, I would not change this here.

docs/features/communication/some_ip_gateway/service_discovery/index.rst
ZoranCutura: Yes, adding requirements to all three specs.

docs/features/communication/some_ip_gateway/architecture/index.rst
LittleHuba: Yes exaclty.

docs/features/communication/some_ip_gateway/service_discovery/index.rst
ZoranCutura: I don't know and understand the GenericProxy feature in detail, but I'd expect that all configuration of Events, Fields, and Methods should happen via deployment
and be rather statical for the SOME/IP Gateway during operations. Would that still hold true with the GenericProxy approach?

docs/features/communication/some_ip_gateway/architecture/index.rst
ZoranCutura: Not sure if I understand this.
Usually one checks CRC first and only if there is no CRC error, sequence counters are checked.
Now if there is a service / instance id corruption and the CRC error is reported to the wrong recipient but as a sequence error, would that be any better?

docs/features/communication/some_ip_gateway/service_discovery/index.rst
ZoranCutura: done

docs/features/communication/some_ip_gateway/architecture/index.rst
LittleHuba: The problem is how do we deliver a CRC error if we do not know whether ServiceID or InstanceID were corrupted? They are also protected by the same CRC after all.
So worst case, we would report the CRC error to somebody completely unrelated. This could cause havoc in a communication that did not have any E2E failures in reality.

The only safe thing we can do in the gateway is to discard the message completely and not forward it at all.
This would result in a sequence error at the client once the next message comes in, or in a timeout error if all following messages fail the CRC check as well.

All in all this is not something we can decide now. I just wanted to raise it. Maybe you can add a note that CRC errors are problematic in the gateway because we cannot safely forward such messages to the correct recipients.

docs/features/communication/some_ip_gateway/assets/some_ip_gateway_architecture.drawio.svg
LittleHuba: Uff I missed that one. Good catch.
I would remove the note since in a lot of other places we argue that we need two processes. This just confuses people.

docs/features/communication/some_ip_gateway/service_discovery/index.rst
LittleHuba: The generic proxy allows at runtime to find out what interface a producer has.
For a gateway this is very powerful since you don't have to compile it with every potential interface that you might want to forward.

Based on the GenericProxy you would then in the gateway only need to configure service and instance ids but no methods or events.
But we likely would need to extend the concept to a GenericSkeleton as well to also cover the sending side in the gateway. That is why I added the "not quite sure yet".

docs/features/communication/some_ip_gateway/architecture/index.rst
ZoranCutura: Ok. Will add a note

docs/features/communication/some_ip_gateway/assets/some_ip_gateway_architecture.drawio.svg
ZoranCutura: Ok, my fault, I overlooked the "one process". Changed it to a less confusing comment.

docs/features/communication/some_ip_gateway/service_discovery/index.rst
ZoranCutura: And wouldn't this also cause a longer startup-time? Not sure though how much longer.

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: to be discussed in our call on thursday 2025-08-14.

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+     - x
+   * - Timeout
+     - Detects if no message is received within a defined time window.
+     -
+     - x
+
+E2E Design and Interface Considerations
+---------------------------------------
+As mentioned above, not all E2E checks can be kept hidden within the gateway. For some problems it is up to the application to decide
+whether it is still valuable to access and process the data. This is in particular true for duplication or re-odering issues.
```
hth9fe:
```suggestion
whether it is still valuable to access and process the data. This is in particular true for duplication or re-ordering issues.
```

typo

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+Therefore, it is required to pass a **tupel** consisting of the **payload data** as well as **supplementary E2E metadata and error information**
+to the IPC clients to enable the client to individually judge on particular E2E issues.
+
+* The API design needs to put the payload information as well as the additional E2E metadata as close as possible together to easily enable and motivate clients to consequently check for potential errors.
+* Additional metadata and error information needs to be incorporated into regular mw::com user interface
+* Error related interface design needs to be highly optimezed for the "good case" to have an optimized support for the common IPC case
```
hth9fe:
```suggestion
* Error related interface design needs to be highly optimized for the "good case" to have an optimized support for the common IPC case
```

typo

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+* AoU's need to be provided to force the user to check for the (E2E-) result
+* Additional E2E metadata to be handed over to the clients:
+
+  * Counter, slightly different interpretation depending on profile, n/a in case the profile doesn't support it
+  * Data ID, n/a in case the profile does not support it or if the Data ID is not explicitely transmitted like in profile 22
```
hth9fe:
```suggestion
  * Data ID, n/a in case the profile does not support it or if the Data ID is not explicitly transmitted like in profile 22
```

typo

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+* Proposed Error Enumeration
+
+  * No E2E error
+  * CRC Error
+  * Sequence error (further subqualification in loss, duplication, reordering is up to the client based on the counter)
+
+.. note::
+   The proposed error enumeration is an abstraction. Deriving detailed errors
+   based on the E2E metadata is task of the client.
+   For reference, this is the error enumeration of the Autosar specification (R24-11):
```
hth9fe:
```suggestion
   For reference, this is the error enumeration of the AUTOSAR specification (R24-11):
```

always capital letters

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+   * INITIAL
+
+E2E State Machine Considerations
+--------------------------------
+The E2E (End-to-End) state machine as defined within AUTOSAR E2E protocol provides a summarized result
+about the overall health and state of a communication channel. Unlike individual E2E Profile Check() functions,
+which assess data validity for a single communication cycle, the state machine aggregates results from multiple Check()
+function invocations over a period. This allows it to determine a more holistic and debounced status of the communication.
+
+Purpose of the E2E State Machine:
```
hth9fe:
```suggestion
Purpose of the E2E State Machine:
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
add formatting, otherwise in the generated documentation it looks weird or additional text doesn't make sense

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+*Diagram: E2E state machine responsibility associated to the gateway*
+
+.. figure:: ../assets/e2e_state_machine_in_gateway.drawio.svg
+   :align: center
+   :name: _e2e_state_machine_in_gateway
+
+   E2E state machine responsibility associated to the gateway
+
+Due to pub/sub nature of mw::com, clients listening on the same topic can not be separately addressed. Therefore, **the state machine results
+can not be selectively distributed according to the particular communication channel they belong to**.
```
hth9fe:
```suggestion
cannot be selectively distributed according to the particular communication channel they belong to**.
```

docs/features/communication/some_ip_gateway/architecture/index.rst
```diff
@@ -0,0 +1,179 @@
...
+
+*Diagram: E2E state machine responsibility associated to the gateway*
+
+.. figure:: ../assets/e2e_state_machine_in_gateway.drawio.svg
+   :align: center
+   :name: _e2e_state_machine_in_gateway
+
+   E2E state machine responsibility associated to the gateway
+
+Due to pub/sub nature of mw::com, clients listening on the same topic can not be separately addressed. Therefore, **the state machine results
```
hth9fe:
```suggestion
Due to pub/sub nature of mw::com, clients listening on the same topic cannot be separately addressed. Therefore, **the state machine results
```

hth9fe (commented):
first review of some-ip gateway architecture document with minor findings

docs/features/communication/some_ip_gateway/service_discovery/index.rst
LittleHuba: Something to evaluate during implementation.

ZoranCutura: Discussed in working session and concluded that SOME/IP events, methods, fields should be configured with mapping to TCP or UDP...

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: conclusion after discussion:
we keep the separate explanations, but an implementation is free to join the concepts, i.e. in one configuration.

LittleHuba (dismissed):

LittleHuba (approved):

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,249 @@
...
+.. _Rationale:
+
+Rationale
+==========
+
+SOME/IP and IPC use different mechanisms and data representations to communicate. Therefor it's not only a task of this gateway
+to adapt the communication, but also to transform from one representation to another. This includes changing data layout, filling in missing data, filtering traffic, error handling, service discovery and further steps.
+
+Hence the gateway shall be a transparent bridge between IPC and SOME/IP. On the one side it should act as a participant in IPC and fulfill all requirements to this accordingly, where providing data or consuming it.
+whereas on the other side it shall participate in SOME/IP communication acting as a SOME/IP service and client fulfilling all SOME/IP requirements and defined communication.
```
hth9fe:
```suggestion
Whereas on the other side it shall participate in SOME/IP communication acting as a SOME/IP service and client fulfilling all SOME/IP requirements and defined communication.
```

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,249 @@
...
+=============
+
+The requirements from Communication generally apply to the SOME/IP Gateway.
+
+
+SOME/IP protocol implementation
+-------------------------------
+
+The protocol implementation shall be fully compatible and complying with the SOME/IP specification from AUTOSAR Adaptive. (:need:`feat_req__some_ip_gateway__someip_protocol`)
+Specifically the SOME/IP specification from AUTOSAR release 24-11 shall be supported by the SOME/IP Gateway. This shall guarantee that systems integrated with the SOME/IP gateway can be used in according
```
hth9fe:
```suggestion
Specifically the SOME/IP specification from AUTOSAR release R24-11 shall be supported by the SOME/IP Gateway. This shall guarantee that systems integrated with the SOME/IP gateway can be used in according
```

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,249 @@
...
+The SOME/IP Gateway shall support SOME/IP Service Discovery. Please refer to the Service Discovery page for detailed discussions on how SD shall be realized with IPC-Clients.
+
+..
+   # fix #1424 SOME/IP Gateway in lifecycle-management
+
+SOME/IP Gateway processes and life cycle management
+---------------------------------------------------
+
+The implementation of a gateway likely requires using one or multiple processes. Startup of processes for the gateway, as with any other
+process in the s-core system, need to be put under the control of lifecycle-management (see feature Lifecycle).
```
hth9fe:
```suggestion
process in the S-CORE system, need to be put under the control of lifecycle-management (see feature Lifecycle).
```

docs/features/communication/some_ip_gateway/index.rst
```diff
@@ -0,0 +1,249 @@
...
+License Impact
+==============
+
+[How could the copyright impacted by the license of the new contribution?]
+
+Since SOME/IP is a protocol, including applied E2E protection and the according profile (polynom, etc.),
+defined by AUTOSAR and published under the license of AUTOSAR, the gateway implementation shall carefully distinguish between the SOME/IP communication stack,
+the E2E protection of data, and the integration into S-CORE mw::com. Breach of foreign licenses must be avoided.
+
+Anybody using SOME/IP Gateway needs to make sure to follow the license conditions and rules of AUTOSAR.
```
hth9fe, line 244: but what does this mean? If you need to follow license conditions and rules of AUTOSAR would mean again, everyone using the SOME/IP Gateway needs to be AUTOSAR member. Really?

hth9fe (commented):
minor findings except one. Really you need to follow AUTOSAR license conditions?

docs/features/communication/some_ip_gateway/index.rst
ZoranCutura: Yes, that's the current situation and that's how we understand the conditions. We decided to formulate it like this, in order to keep this documentation stable even when the license conditions of ARA change.

lurtz (approved):

qor-lb (approved):

ZoranCutura:
@qor-lb the commits have been adapted / squashed, keeping all the commits from individuals of different companies.
Please proceed with merging

qor-lb (approved):

PR #1192, opened 2025-06-03
Use 'process_description' as Process

MaximilianSoerenPollak:
This will delete the Process folder and instead use the [process_description module](https://github.com/eclipse-score/process_description).


MaximilianSoerenPollak:
Blocked until new release of 'docs-as-code'. 

~~Also question regarding some Metamodel options: https://github.com/eclipse-score/process_description/issues/26~~

Edit: 

Question has been answered.

New docs-as-code release in the works currently

MaximilianSoerenPollak:
One issue here is that the Documentation build can not be run as the target for the Documentation build has changed.
This means that only after merge it shows if it's correct.

masc2023 (approved):
works fine for me

AlexanderLanin (approved):

PR #1201, opened 2025-06-05
Enable external needs json & increase versions

MaximilianSoerenPollak:
Currently all needs that are external (like process) will be filtered out in the 'needs.json'. 

This allows the needs.json to contain ALL of the needs imported. 

In a future PR the filter can probably be adapted to make it more efficent, e.g. only get external needs that are actually needed. 

~~## ⚠️  This PR should not be merged until new docs-as-code is out. As that has bugfixes that might be imported for here.~~

New docs-as-code version is available. PR can be reviewed and mered. 

--- 

I pulled the version to a new major one, as the change in needs.json is something that changes how it could be used.
I think that makes sense (at least in my head) .

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -68,3 +68,5 @@ needs_string_links: dict[str, dict[str, Any]] = {
         "options": ["source_code_link"],
     },
 }
+
+needs_builder_filter = ""
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -68,5 +68,5 @@ needs_string_links: dict[str, dict[str, Any]] = {
         "options": ["source_code_link"],
     },
 }
-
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -68,3 +68,5 @@
         "options": ["source_code_link"],
     },
 }
+# This ensures all needs that are imported show up in the build 'needs.json'
+needs_builder_filter = ""
```
AlexanderLanin, line 72: this belongs to wherever we have the external need stuff and not into individual conf.py files

MaximilianSoerenPollak: I put it into the conf.py here right now, as this can be done without a complete new release of docs-as-code. 

In the next release we should add that in there as a default and figure out a better filter yes.

PR #1202, opened 2025-06-05
Added FEO feature architecture

marcmo:
originates from CR #131991


armin-acn:
This is a new proposal of the FEO feature architecture, replacing #1109.

ramceb (approved):
We can approve this for now but we will have to align the component names with the other feature requests. The problem is, that we don't have a nameing scheme yet.

ramceb (approved):

PR #1224, opened 2025-06-11
Summary information about rust coding guidelines

PandaeDo:
Overview of known information's regarding coding guidelines

### DIFF
--- a/docs/contribute/development/rust/coding_guidelines.rst
+++ b/docs/contribute/development/rust/coding_guidelines.rst
@@ -17,6 +17,52 @@ Writing Rust Code
 
 .. document:: Coding Guidelines Rust
    :id: doc__rust_coding_guidelines
-   :status: invalid
+   :status: valid
 
-.. note:: The Rust coding guidelines are not yet defined. This is a placeholder for future content.
+Coding Guidelines
+#################
+
+The Rust consortium has decided that there will be no official coding guidelines provided by the Rust consortium.
+However, the Rust community has a set of guidelines that are widely accepted and used.
+These guidelines can be found in the official Rust documentation and are recommended for all Rust projects.
+
+`Safety Critical Rust Consortium <https://github.com/rustfoundation/safety-critical-rust-consortium/tree/main/subcommittee/coding-guidelines/>`_
+
+`Learn unsafe Rust <https://google.github.io/learn_unsafe_rust/>`_
+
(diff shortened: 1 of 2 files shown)


docs/contribute/development/rust/coding_guidelines.rst
```diff
@@ -17,6 +17,104 @@ Writing Rust Code
...
+consortium is evaluating what further tooling and support are necessary to
+enable certification and safe use of Rust in automotive applications.
+
+`Safety-Critical Rust Consortium <https://rustfoundation.org/safety-critical-rust-consortium>`_
+
+`Safety-Critical Rust Consortium Guidelines <https://github.com/rustfoundation/safety-critical-rust-consortium/tree/main/subcommittee/coding-guidelines/>`_
+
+`Learn unsafe Rust <https://google.github.io/learn_unsafe_rust/>`_
+
+`Rust language <https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html>`_
```
darkwisebear, line 46: Imo these references are - while worth reading - misleading, as unsafe Rust does not necessarily mean bad code, nor does sticking to safe Rust mean perfect code for safe applications.

docs/contribute/development/rust/coding_guidelines.rst
```diff
@@ -17,6 +17,104 @@ Writing Rust Code
...
+to catch issues that the standard compiler might miss, making it an essential
+tool for writing clean, idiomatic, and efficient Rust code.
+
+`Link to Clippy <https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/issues/78/>`_
+
+
+Conclusions for S-CORE
+######################
+
+During the S-CORE project only formatting and clippy checks are enforced. Also
```
darkwisebear: I think we should also strive to use a tool like Miri. This can be a second step, though, and I agree with having formatting and clippy checking as a primary goal for the first step.

docs/contribute/development/rust/coding_guidelines.rst
PandaeDo: Yes it is. But it's the name of the book so I can't change it.

docs/contribute/development/rust/coding_guidelines.rst
PandaeDo: Add Miri

vinodreddy-g (approved):

PR #1259, opened 2025-06-18
update process_description version

masc2023:
Version changed to v1.0.1


pahmann (approved):
These updates reflect the versioning in S-Core, which has been previously modified in process_description repo. See: https://github.com/eclipse-score/process_description/pull/39

PR #1375, opened 2025-07-06
abi compatible data types feature request

### LINKED ISSUE #917: Feature Request for ABI compatible datatypes
Discussed in https://github.com/orgs/eclipse-score/discussions/764

>  - Requirement: ABI compatible data types
>      - FR required: Lead: Qorix

Acceptance Criteria:
- Feature Request written according to the [Change Management](https://eclipse-score.github.io/score/main/process/process_areas/change_management/change_management_concept.html) & [Feature Request Template](https://eclipse-score.github.io/score/main/process/process_areas/change_management/guidance/change_management_feature_template.html)
- Feature requirements written according to the [Requirements Engineering](https://eclipse-score.github.io/score/main/process/process_areas/requirements_engineering/requirements_concept.html)
- If necessary: extend the stakeholder requirements written according to the [Requirements Engineering](https://eclipse-score.github.io/score/main/process/process_areas/requirements_engineering/requirements_concept.html)

qor-lb:
closes #917

### DIFF
--- a/docs/features/communication/abi_compatible_data_types/index.rst
+++ b/docs/features/communication/abi_compatible_data_types/index.rst
@@ -1,0 +1,490 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2025 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+.. _abi_compatible_data_types_feature:
+
+ABI Compatible Data Types
+#########################
+
+.. document:: ABI Compatible Datatypes
+   :id: doc__abi_compatible_data_types
+   :status: valid
+   :safety: ASIL_B
+   :tags: feature_request, change_management, communication, abi_compatible_data_types
+
+
+.. toctree::
+   :hidden:
+
+   requirements.rst
+
+
+Feature flag
+============
+
+To activate this feature, use the following feature flag:
+
+``experimental_abi_compatible_data_types``
+
+
+Abstract
+========
+
+This feature request defines a set of ABI-compatible data types and a runtime type description format to support zero-copy inter-process communication between C++17 and Rust 1.8x processes using the same endianness. It ensures consistent type layouts across languages by requiring fixed-size, statically allocated types without absolute pointers or language-specific metadata.
+
+The specification covers primitive types, structs, enums, arrays, and introduces ABI-stable representations for vectors, options, and results. A runtime-readable type description enables processes to interpret shared memory without compile-time access to type definitions.
+
+
+Motivation
+==========
+
+This feature request addresses specific challenges in achieving type compatibility within our inter-process communication (IPC) framework that leverages zero-copy shared memory mechanisms. Two essential scenarios are under evaluation:
+
+1. **ABI Compatibility**: Processes implemented in different programming languages (C++17 and Rust 1.8x) must interpret a shared memory location consistently as the same native type, provided both have compile-time access to the type definition. This scenario eliminates serialization overhead and allows direct memory access.
+
+2. **Type Description**: It should be possible to record arbitrary data streams, and convert or analyze them at a later time and/or on a different system, without having to recompile the conversion or analysis tools for that particular data format. A machine-readable description of the format, including any user-defined data types, should be available on request during runtime. In addition, this description could potentially be used by gateway processes to perform relatively simple but generic transformations between different data representations.
+
+
+ABI Compatibility
+-----------------
+
+Our communication feature relies on shared memory to transfer data between processes. For effective zero-copy data exchange, processes written in C++17 and Rust 1.8x must inherently understand the data at shared memory locations identically. Achieving this requires ensuring that data types have consistent, fixed-size memory layouts.
+
+This evaluation initially targets the following process configurations:
+
+* Processes running on the same operating system.
+* Processes running on different operating systems but under the same hypervisor.
+
+Supporting different endianness between processes is explicitly out of scope, as it inherently demands bit manipulation, effectively requiring serialization.
+Different bit widths, however, are implicitly supported by specifying the width of all types and excluding word-size integers.
+
+The following data types shall be supported:
+
+* **Primitive Types**:
+
+  * Boolean
+  * Numeric (fixed-size integers 8-128 bits, signed and unsigned; IEEE 754 floating-point numbers)
+
+* **Sequence Types**:
+
+  * Array (fixed-length)
+
+* **User-Defined Types**:
+
+  * Struct
+  * Tuple
+  * Enum (tag-only)
+  * Variant ("tagged union"; *optional*)
+
+* **Fixed-Size, Variable-Length Containers**:
+
+  * Vector
+  * Queue
+  * Hash map (*optional*)
+  * Hash set (*optional*)
+  * Binary tree (*optional*)
+
+* **Specialized Variants**:
+
+  * Result
+  * Option
+
+All provided data types must ensure fixed size and consistent memory layouts.
+
+Type Description
+----------------
+
+A critical scalability feature involves gateway processes, which subscribe to IPC endpoints and translate ABI-compatible data types into external serialization formats. These gateways require the ability to interpret data without compile-time access to type definitions. To address this, an explicit runtime-readable type description format is necessary. This description allows dynamic, runtime interpretation of data structures, enabling the addition of new IPC topics without recompiling gateway processes.
+
+In summary, the motivation behind this feature request is to define and standardize ABI-compatible data types and a runtime-accessible type description mechanism to ensure interoperability and scalability in zero-copy IPC scenarios involving multiple languages and dynamic environments.
+
+
+.. Rationale
+.. ==========
+
+
+Specification
+=============
+
+ABI Compatibility
+-----------------
+
+This specification defines the set of rules and constraints for representing data types in shared memory such that they can be interpreted consistently across processes implemented in C++17 and Rust 1.8x. These types enable zero-copy inter-process communication by enforcing ABI compatibility at the memory layout level. The focus is on data exchange between processes using the same endianness.
+
+Assumptions
+^^^^^^^^^^^
+
+* Shared memory regions are mapped at correctly aligned virtual addresses in both processes.
+* No serialization or runtime copying occurs when interpreting a type from memory.
+* Processes use the same endianness.
+* No synchronization or atomicity guarantees are defined at the data type level; these must be provided by the accessing code.
+* All memory is allocated statically or pre-reserved. Dynamic memory allocation is disallowed.
+
+Type Conformance
+^^^^^^^^^^^^^^^^
+
+Types used in shared memory must meet the following criteria:
+
+1. **Fixed size and alignment**: Every type must have a known, constant size and alignment at compile time.
+2. **Consistent layout across languages**: The layout of a type must be identical in Rust and C++, and on all platforms.
+3. **No absolute pointers or references**: Types must not contain absolute pointers/function pointers/references or any pointers/references that point out of the transferred data.
+4. **No language-specific metadata**: No vtables, slice headers, or implementation-specific type markers are allowed.
+
+Each type definition must clearly indicate whether it conforms to these rules natively or requires a custom definition to do so.
+
+Type Categories
+^^^^^^^^^^^^^^^
+
+Primitive Types
+"""""""""""""""
+
+These types are ABI-compatible when declared using fixed-size standard types:
+
+
+.. list-table:: Native Type Mapping
+   :header-rows: 1
+
+   * - Concept
+     - Rust
+     - C++17
+   * - Boolean
+     - ``bool``
+     - ``bool`` (1 byte, with ``0x00`` and ``0x01`` as the only valid bit patterns)
+   * - Integers (N = 8, 16, 32, 64)
+     - ``uN``, ``iN``
+     - ``std::uintN_t``, ``std::intN_t``
+   * - Floating point
+     - ``f32``, ``f64``
+     - ``float``, ``double`` (compliant with IEEE 754)
+
+All types must avoid trap representations and undefined padding.
+
+Structs and Tuples
+""""""""""""""""""
+
+Structs and tuples are supported using standard layout rules:
+
+* **Rust**: Requires ``#[repr(C)]``
+  (guarantees C-compatible memory layout;
+  `full specification <https://doc.rust-lang.org/reference/type-layout.html#the-c-representation>`__)
+* **C++**: Requires ``standard_layout``
+  (no virtual functions, no virtual inheritance, and only one class in the hierarchy has non-static data members;
+  `full specification <https://en.cppreference.com/w/cpp/language/classes.html#Standard-layout_class>`__)
+
+Field ordering must be preserved and padding must be identical across compilers. Any alignment greater than the default must be explicitly declared.
+
+Enums
+"""""
+
+Only fieldless enums with a defined underlying integer type are supported. These must use:
+
+* ``#[repr(u8)]``, ``#[repr(u16)]``, etc. in Rust
+* ``enum class MyEnum : std::uint8_t`` in C++
+
+*Note:* Enums with payloads ("variants" or "tagged unions") are optionally supported.
+
+Arrays
+""""""
+
+Fixed-size arrays are naturally ABI-compatible and supported in both languages.
+
+* Rust: ``[T; N]``
+* C++: wrapper around ``T[N]`` to enforce bounds-checking for element access
+
+Element types must also conform to this specification. No dynamic length information is allowed.
+
+Vectors
+""""""""
+
+To provide bounded sequence types with familiar APIs, a custom vector implementation must be provided in both languages that matches the memory layout defined below.
+
+.. code-block:: rust
+
+    #[repr(C)]
+    pub struct AbiVec<T> {
+        len: u32,
+        capacity: u32,
+        elements: [T; N],
+    }
+
+.. code-block:: cpp
+
+    template<typename T, std::size_t N>
+    struct AbiVec {
+    private:
+        std::uint32_t len;
+        std::uint32_t capacity;
+        T elements[N];
+    };
+
+* Capacity is fixed and equal to ``N`` at compile time.
+* Overflow beyond capacity must be a checked error.
+* No heap allocation is permitted.
+* Internally, these are ABI-compatible with ``len``, ``capacity`` and ``elements`` accessible from both languages.
+* The public API must match standard vector types in usability (e.g. ``push()``, ``pop()``).
+
+Option Types
+""""""""""""
+
+ABI-compatible optional types must be implemented manually using a one-byte tag followed by a payload.
+
+.. code-block:: rust
+
+    #[repr(C)]
+    pub struct AbiOption<T> {
+        is_some: u8,
+        value: T,
+    }
+
+.. code-block:: cpp
+
+    template<typename T>
+    struct AbiOption {
+    private:
+        std::uint8_t is_some;
+        T value;
+    };
+
+* ``is_some == 0`` indicates absence; ``1`` indicates presence.
+* The value field is always initialized and occupies memory regardless of state.
+* The public API must match standard optional types in usability.
+
+Result Types
+""""""""""""
+
+Result types represent tagged unions with two possible states.
+
+.. code-block:: rust
+
+    #[repr(C)]
+    pub struct AbiResult<T, E> {
+        is_ok: u8,
+        value: AbiResultUnion<T, E>,
+    }
+
+    #[repr(C)]
+    union AbiResultUnion<T, E> {
+        ok: T,
+        err: E,
+    }
+
+.. code-block:: cpp
+
+    template<typename T, typename E>
+    struct AbiResult {
+    private:
+        std::uint8_t is_ok;
+        union {
+            T ok;
+            E err;
+        } value;
+    };
+
+* ``is_ok == 1`` indicates ``ok`` field is valid
+* ``is_ok == 0`` indicates ``err`` field is valid
+* The layout must guarantee correct union member interpretation based on the discriminant
+
+Language Conformance Summary
+""""""""""""""""""""""""""""
+
+.. list-table::
+   :header-rows: 1
+
+   * - Feature
+     - Rust Native Support
+     - C++ Native Support
+     - Specification Status
+   * - Primitives
+     - ✅ Native types
+     - ✅ Native types
+     - Conforming
+   * - Structs
+     - ✅ ``#[repr(C)]``
+     - ✅ ``standard_layout``
+     - Conforming
+   * - Enums (fieldless)
+     - ✅ ``#[repr(C)]``
+     - ✅ With fixed base
+     - Conforming
+   * - Arrays
+     - ✅ ``[T; N]``
+     - ✅ ``T[N]``
+     - Conforming
+   * - Vector
+     - ❌ ``Vec<T>``
+     - ❌ ``std::vector<T>``
+     - ✅ ``AbiVec<T, N>`` required
+   * - Option
+     - ❌ ``Option<T>``
+     - ❌ ``std::optional<T>``
+     - ✅ ``AbiOption<T>`` required
+   * - Result
+     - ❌ ``Result<T, E>``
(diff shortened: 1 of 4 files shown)


docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,404 @@
...
+============
+
+To activate this feature, use the following feature flag:
+
+``experimental_abi_compatible_data_types``
+
+
+Abstract
+========
+This feature request defines a set of ABI-compatible data types and a runtime type description format to support zero-copy inter-process communication between C++17 and Rust 1.8x processes on the same architecture and endianness. It ensures consistent type layouts across languages by requiring fixed-size, statically allocated types with no pointers or language-specific metadata.
```
LittleHuba: No pointers?

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,404 @@
...
+* All memory is allocated statically or pre-reserved. Dynamic memory allocation is disallowed.
+
+Type Conformance
+^^^^^^^^^^^^^^^^
+
+Types used in shared memory must meet the following criteria:
+
+1. **Fixed size and alignment**: Every type must have a known, constant size and alignment at compile time.
+2. **Consistent layout across languages**: The layout of a type must be identical in Rust and C++ on the same platform.
+3. **No pointers or references**: Types must not contain pointers to heap memory, function pointers, or references.
```
LittleHuba: This is immensely restrictive and not absolutely necessary. Yes, absolute pointers are problematic, but offset pointers are absolutely viable and something that is already used nowadays in LoLa.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,404 @@
...
+
+All types must avoid trap representations and undefined padding.
+
+Structs and Tuples
+""""""""""""""""""
+
+Structs and tuples are supported using standard layout rules:
+
+* **Rust**: Requires ``#[repr(C)]``
+* **C++**: Requires ``standard_layout`` types with no inheritance or virtual functions
```
alex-warg-mb: C++: standard_layout types are no virtual functions, and no 'virtual' inheritance, and only one class in the hierarchy has non-static data members.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,463 @@
+..
```
4og, line 1: Some of OSS projects that can potentially fulfill the goals of this FT:

Cap'n Proto https://capnproto.org/
Flatbuffers https://flatbuffers.dev/
Data Interface in Apache Arrow https://arrow.apache.org/docs/format/CDataInterface.html
Cista++ https://github.com/felixguendling/cista
rkyv https://rkyv.org/rkyv.html
epserde-rs https://github.com/vigna/epserde-rs/
musli-zerocopy https://github.com/udoprog/musli
FracPack https://hive.blog/blog/@dan/presenting-fracpack-a-better-zero-copy-binary-serialization-format-for-c-and-rust

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,463 @@
...
+
+  * Vector
+  * Hash map (optional)
+  * Hash set (optional)
+  * Binary tree (optional)
+
+* **Union-like Constructs** (no direct unions):
+
+  * Result
+  * Option
```
4og, line 101: Do we want to support real unions? For example, the SOME/IP spec supports unions.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,463 @@
...
+
+Motivation
+==========
+
+This feature request addresses specific challenges in achieving type compatibility within our inter-process communication (IPC) framework that leverages zero-copy shared memory mechanisms. Two essential scenarios are under evaluation:
+
+1. **ABI Compatibility**: Processes implemented in different programming languages (C++17 and Rust 1.8x) must interpret a shared memory location consistently as the same native type, provided both have compile-time access to the type definition. This scenario eliminates serialization overhead and allows direct memory access.
+
+2. **Type Description**: Processes responsible for translating data between internal ABI-compatible formats and external serialization formats should perform this translation without compile-time knowledge of type definitions. This runtime interpretation capability is critical for scalability, ensuring that gateway processes (or similar entities) can dynamically handle new data types without recompilation.
+
```
4og, line 57: I would explicitly include the record/replay use case. It means if the the IPC data is recorded, we'd need some sort of tooling (e.g. in Python) to process the data further. Additionally, some tooling for debugging would be required.

docs/features/communication/abi_compatible_data_types/index.rst
HartmannNico: Current resolution: No, but we consider a type

* Variant (optional)

to cover that capability. Further discussion needed.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: Fixed in #1451

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,463 @@
...
+.. list-table:: Native Type Mapping
+   :header-rows: 1
+
+   * - Concept
+     - Rust
+     - C++17
+   * - Boolean
+     - ``bool``
+     - ``bool`` (1 byte)
+   * - Integers (N = 8..128)
```
awillenbuecher-xq-tec: C++ doesn't have 128-bit integers, and not all compilers support them as extensions.

```suggestion
   * - Integers (N = 8, 16, 32, 64)
```

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -41,7 +41,8 @@ To activate this feature, use the following feature flag:
 
 Abstract
 ========
-This feature request defines a set of ABI-compatible data types and a runtime type description format to support zero-copy inter-process communication between C++17 and Rust 1.8x processes on the same architecture and endianness. It ensures consistent type layouts across languages by requiring fixed-size, statically allocated types with no pointers or language-specific metadata.
+
+This feature request defines a set of ABI-compatible data types and a runtime type description format to support zero-copy inter-process communication between C++17 and Rust 1.8x processes using the same endianness. It ensures consistent type layouts across languages by requiring fixed-size, statically allocated types with no pointers or language-specific metadata.
```
LittleHuba:
```suggestion
This feature request defines a set of ABI-compatible data types and a runtime type description format to support zero-copy inter-process communication between C++17 and Rust 1.8x processes using the same endianness. It ensures consistent type layouts across languages by requiring fixed-size, statically allocated types with no absolute pointers or language-specific metadata.
```

Completely forbidding pointers will make any complex and performant containers near impossible. Instead, we should strive to use relative offset pointers.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+
+
+Security Impact
+===============
+
+.. note::
+
+   This section does not replace a formal security impact analysis. This only guides the design thoughts behind security related topics and is to be understood as a starting point for later workflows.
+
+No assumed impact on security.
```
4og: IMHO this feature has significant security implications, particularly regarding data reading. Given that the primary use of the ABI types is for communication, they could easily become a threat target. We gonna need at least basic structural validation to prevent issues like out-of-bounds access.

lakmal-score:
@awillenbuecher-xq-tec Could you see the current status of the open review comments and see what else is pending to get the first version of this ready to be reviewed by the architecture round?

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: I've added the *"recording/replaying/inspecting data"* use case.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: Changed to *"8, 16, 32, 64"* bits.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: Changed to *"No absolute pointers or references"*.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: Changed to *"without absolute pointers"*.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: I've changed it to *"without absolute pointers"* to allow relative pointers and offsets.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: Wouldn't the need to validate indices and offsets apply equally to any kind of data structure or protocol format, independent of whether it's an "ABI compatible type" or something else?

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+
+
+Motivation
+==========
+
+This feature request addresses specific challenges in achieving type compatibility within our inter-process communication (IPC) framework that leverages zero-copy shared memory mechanisms. Two essential scenarios are under evaluation:
+
+1. **ABI Compatibility**: Processes implemented in different programming languages (C++17 and Rust 1.8x) must interpret a shared memory location consistently as the same native type, provided both have compile-time access to the type definition. This scenario eliminates serialization overhead and allows direct memory access.
+
+2. **Type Description**: Processes responsible for translating data between internal ABI-compatible formats and external serialization formats should perform this translation without compile-time knowledge of type definitions. This runtime interpretation capability is critical for scalability, ensuring that gateway processes and tools for recording/replaying/inspecting data can dynamically handle new data types without recompilation.
```
LittleHuba: Gateways likely will have to do additional steps like filling in default values, act on errors in specific manners, ... which might very much require recompilation for new data types.

In the context of the SOME/IP gateway we also talk about transformations that might not be straight-forward 1:1 mappings. 
Clients (especially in cooperations) have often a very different picture of how a message should look like, compared to what you get from the network. If you allow for more complex transformations you can avoid an additional client-side transformation. While this may force recompilation, this is a manageable complexity given some good tooling.

All in all, I would not list gateways as a real driver for type description.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+
+Our communication feature relies on shared memory to transfer data between processes. For effective zero-copy data exchange, processes written in C++17 and Rust 1.8x must inherently understand the data at shared memory locations identically. Achieving this requires ensuring that data types have consistent, fixed-size memory layouts.
+
+This evaluation initially targets the following process configurations:
+
+* Processes running on the same operating system.
+* Processes running on different operating systems but under the same hypervisor.
+
+Supporting different endianness between processes is explicitly out of scope, as it inherently demands bit manipulation, effectively requiring serialization.
+Different bit widths, however, are implicitly supported by specifying the width of all types and excluding word-size integers.
```
LittleHuba, line 70: How would this play with alignments? I would imagine we end up in quite a huge mess if we have two parties accessing the same data but with a different bit width.

Maybe you mean that we want to support that our ABI compatible datatypes work on different bit widths, as long as all participants have the same bit width. If so, we should rephrase to make this more explicit.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+
+The following data types shall be supported by the IPC mechanism:
```
LittleHuba: While IPC is a main driver, also crossing the FFI boundary without IPC is a driver. So let's not restrict this to IPC only.

```suggestion
The following data types shall be supported:
```

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+
+  * Result
+  * Option
+
+All provided data types must ensure fixed size and consistent memory layouts.
+
+Type Description
+----------------
+
+A critical scalability feature involves gateway processes, which subscribe to IPC endpoints and translate ABI-compatible data types into external serialization formats. These gateways require the ability to interpret data without compile-time access to type definitions. To address this, an explicit runtime-readable type description format is necessary. This description allows dynamic, runtime interpretation of data structures, enabling the addition of new IPC topics without recompiling gateway processes.
```
LittleHuba, line 108: See comment above on gateways and their transformations. I think reprocessing is a better driver for this.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+
+This specification defines the set of rules and constraints for representing data types in shared memory such that they can be interpreted consistently across processes implemented in C++17 and Rust 1.8x. These types enable zero-copy inter-process communication by enforcing ABI compatibility at the memory layout level. The focus is on data exchange between processes using the same endianness.
+
+Assumptions
+^^^^^^^^^^^
+
+* Shared memory regions are mapped at correctly aligned virtual addresses in both processes.
+* No serialization or runtime copying occurs when interpreting a type from memory.
+* Processes use the same endianness.
+* No synchronization or atomicity guarantees are defined at the data type level; these are provided by the IPC framework.
```
LittleHuba: As mentioned above, FFI boundary crossing is also a use case. There user code must provide the synchronization.

```suggestion
* No synchronization or atomicity guarantees are defined at the data type level; these must be provided by the accessing code.
```

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+* All memory is allocated statically or pre-reserved. Dynamic memory allocation is disallowed.
+
+Type Conformance
+^^^^^^^^^^^^^^^^
+
+Types used in shared memory must meet the following criteria:
+
+1. **Fixed size and alignment**: Every type must have a known, constant size and alignment at compile time.
+2. **Consistent layout across languages**: The layout of a type must be identical in Rust and C++, and on all platforms.
+3. **No absolute pointers or references**: Types must not contain pointers to heap memory, function pointers, or references.
```
LittleHuba: A pointer to stack is equally bad.
Also wording here is tricky. What we really want to say is that pointers and even references are fine, as long as they are implemented with a relative offset and do not point outside of the transferred data.


```suggestion
3. **No absolute pointers or references**: Types must not contain absolute pointers/function pointers/references that point out of the transferred data.
```

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+     - C++17
+   * - Boolean
+     - ``bool``
+     - ``bool`` (1 byte, with ``0x00`` and ``0x01`` as the only valid bit patterns)
+   * - Integers (N = 8, 16, 32, 64)
+     - ``uN``, ``iN``
+     - ``std::uintN_t``, ``std::intN_t``
+   * - Floating point
+     - ``f32``, ``f64``
+     - ``float``, ``double``
```
LittleHuba: In C++ you would have to check that the floating point types adhere to IEEE754:

```c++
static_assert(std::numeric_limits<float>::is_iec559);
static_assert(std::numeric_limits<double>::is_iec559);
```

Furhter for floating point types we need to carefully consider whether it is fine that different FPUs work on the same data. FPUs are not standardized. This could lead to significant deviations in calculations.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+
+Enums
+"""""
+
+Only fieldless enums with a defined underlying integer type are supported. These must use:
+
+* ``#[repr(u8)]``, ``#[repr(u16)]``, etc. in Rust
+* ``enum class MyEnum : std::uint8_t`` in C++
+
+Enums with payloads (discriminated unions) are not supported.
```
LittleHuba: We could mention that a variant can be used to work around this limitation?

Not a must from my side, just an idea.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+
+Arrays
+""""""
+
+Fixed-size arrays are naturally ABI-compatible and supported in both languages.
+
+* Rust: ``[T; N]``
+* C++: ``T[N]``
+
+Element types must also conform to this specification. No dynamic length information is allowed.
```
LittleHuba: At least on C++ side we need something else for this. Using C-style arrays in C++ code is a slippery slope straight to hell.

I would go with something that is similar to a `std::array<>` but defines the ABI.
E.g. internally represented by a size field followed by the C-style array, but externally represented by a `std::array<>`-like interface.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+----------------
+
+To address the scenarios outlined in the motivation, a clearly defined type description mechanism is required. The type description provides sufficient information during runtime, enabling a process without compile-time access to type definitions to correctly interpret a given memory location according to the previously established ABI rules.
+
+The goals are:
+
+* Enable interpretation of shared memory content without compile-time access to type definitions.
+* Support all ABI-compatible data types previously defined.
+* Include versioning to manage schema evolution and compatibility.
+* Allow easy generation and parsing by tooling in both C++ and Rust.
```
LittleHuba, line 349: Just as a side note, I personally think this is very hard to achieve with a type-safe, performant API and a static type system as in C++ and Rust.

docs/features/communication/abi_compatible_data_types/index.rst
LittleHuba: I agree that it applies to all data structures independent of ABI compatibility. But nevertheless, Andrey has a valid point that there is an impact.

For example my comment on the C-style arrays goes into that direction. A C-style array has no safeguards against out-of-bounds reads. Hence, we need some wrapper around that checks accesses.

docs/features/communication/abi_compatible_data_types/index.rst
```diff
@@ -0,0 +1,491 @@
...
+----------
+
+Reflection, in this context, is the ability to inspect data at runtime even if its structure is not or not fully known at compile time.
+Benefits of reflection include being able to translate recorded data into a human-readable format (e.g., JSON or CSV) without having to know the type definitions at compile time; this enables general-purpose data recording and transformation tools.
+
+This ability requires some form of *type description* being available at runtime, so that a sequence of bytes can be interpreted as a data structure.
+There are two primary approaches to achieve this goal:
+
+* *inline type descriptions*, which precede each instance of every type, and
+* *external type descriptions*, which are stored separately from the data.
```
LittleHuba, line 415: I feel like you argue reflection a lot in the type description sections above. But here you reject the idea.
Could you rephrase the sections on type description to clear this up?

docs/features/communication/abi_compatible_data_types/requirements.rst
```diff
@@ -0,0 +1,341 @@
...
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+
+Requirements
```
LittleHuba, line 16: I did not comment the requirements. Please adapt according to my comments in the prosa.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: We would only support integers with explicit width (`uN/iN` in Rust and `uintN_t/intN_t` in C++, for `N=8/16/32/64`, but not `usize/isize` respectively `size_t/ptrdiff_t`). This isn't just helpful for mixed-bitwidth systems (like a 64-bit ortex-A alongside a 32-bit Cortex-R), but also when recording data on a 32-bit MCU and inspecting it on a 64-bit desktop.

You raise a good point on alignment, though: `u64` is 8-byte aligned on a 64-bit system, but might only be 4-byte aligned on a 32-bit system. We can solve this by using `alignas(...)` in C++, and a newtype wrapper in Rust.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: I see, good point. How would you formulate this security aspect for the feature request?

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: You're right, this should be clarified. The idea is to have a `T[N]` memory layout, but a wrapper which checks for out-of-bounds accesses the same way Rust's `[T; N]` does.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: This line is outdated now that variants are (optionally) supported. I'll clarify it.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: I'll add a clarification that floating-point types must be IEEE 754 compliant.

I don't think we have to define the behavior of *operations* on those values; the feature request is concerned with data *exchange and storage*, not with data *processing*.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: I think that *external* type descriptions are really useful and should be added at some point, but shouldn't be included in version 1.0 of the feature. Instead, they should be part of "ABI compatible datatypes v1.1" or 2.0, when we have a better understanding of the whole situation.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: I see. Would you be okay with listing *reprocessing* as the primary motivation, and still including gateways as a secondary one?

LittleHuba: As secondary use case I'm fine. If you add a potentially somewhere in that sentence even better. It might be helpful in some scenarios, but definitely nothing mandatory.

docs/features/communication/abi_compatible_data_types/index.rst
LittleHuba: Sorry my "improved wording" is also wrong again...

```suggestion
3. **No absolute pointers or references**: Types must not contain absolute pointers/function pointers/references or any pointer/reference that point out of the transferred data.
```

docs/features/communication/abi_compatible_data_types/index.rst
LittleHuba: Hmm maybe we should just warn users that while the storage is compatible the operations may differ in the result.
This could lead to some very interesting bugs.

docs/features/communication/abi_compatible_data_types/index.rst
awillenbuecher-xq-tec: How about this:

> The memory underlying an ABI compatible type may potentially be corrupted, e.g., because it originates from an untrusted process, or because of a bug in the producer. As a consequence, any pointers, indices, and offsets need to be validated before they're used to access memory, to ensure they don't address any out-of-bounds locations. In addition, care must be taken to avoid TOC/TOU (*time-of-check to time-of-use*) issues, where the producer might maliciously or inadvertently modify a value after the consumer has already validated it.

awillenbuecher-xq-tec:
> @awillenbuecher-xq-tec Could you see the current status of the open review comments and see what else is pending to get the first version of this ready to be reviewed by the architecture round?

@LakmalBMW All comments have been answered and all requested changes implemented, I think.

johannes-esr:
@qor-lb, @awillenbuecher-xq-tec, can you please resolve the merge conflicts?

docs/features/communication/abi_compatible_data_types/index.rst
FScholPer: @qor-lb We need a central place where we document these things (discussion,issue...)

awillenbuecher-xq-tec: These serialization libraries ultimately fulfill a different use case than the one we're aiming for here. A few of them have some overlapping goals, but none of them satisfy all of our requirements, and they typically come with additional constraints, overhead, or unnecessary features.
I don't think we need to archive this list.

PR #1393, opened 2025-07-09
Fix CfgMgt Plan audit actions

### LINKED ISSUE #1392: Fix Cfg Mgt Plan Audit findings
Configuration Management Plan has to be aligned with audit finding fixes done in the process.

aschemmel-tech:
Ref: Resolves #1392

### DIFF
--- a/docs/platform_management_plan/config_management.rst
+++ b/docs/platform_management_plan/config_management.rst
@@ -116,8 +116,8 @@ Git defines branches as a means of parallel development. In the S-CORE project t
 
 The "remote" branch is not "local" to the developer but resides on the "remote" GitHub server.
 
-In S-CORE all configuration items are kept in GitHub, this means that there only needs to be one baseline for these
-(and not multiple ones for each of the work products which are maintained in seperate tools).
+In S-CORE all configuration items are kept in GitHub, this means that there only needs to be one baseline (per repository) for these
+(and not multiple ones for each of the work products types which are maintained in separate tools).
 Baselines are created by using the GitHub "tag" function. The tag name shall correspond to
 the release branch name the tag is created, adding patch version and pre-release tag.
 See also :need:`doc__platform_release_management_plan`.
@@ -126,7 +126,7 @@ As described in "Identification and Properties" above, there are several reposit
 Baselines are created individually in these repositories, even a different version schema could be adopted.
 In case of dependent repositories, the repository dependet upon on has to be baselined first, to be available
 to refer to this baseline when integrating it. That means that for example a platform baseline also
-documents the versions(baselines) of the modules the platform consists of. This can then also be seen in the platform release note.
+documents the versions (baselines) of the modules the platform consists of. This can then also be seen in the platform release note.
 
 Every change in the release repository is also taken over into the main branch. The module development team
 can decide how to ensure this (e.g. by development in main and cherrypick to release branch).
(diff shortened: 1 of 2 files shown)


docs/platform_management_plan/config_management.rst
```diff
@@ -110,10 +116,18 @@ Git defines branches as a means of parallel development. In the S-CORE project t
 
 The "remote" branch is not "local" to the developer but resides on the "remote" GitHub server.
 
+In S-CORE all configuration items are kept in GitHub, this means that there only needs to be one baseline for these
```
masc2023: Everything is kept in GitHut, but maybe over several repos, so the baseline is only per repo, and is there a need to baseline over multiple repos for a S-CORE release?

docs/platform_management_plan/config_management.rst
```diff
@@ -110,10 +116,18 @@ Git defines branches as a means of parallel development. In the S-CORE project t
...
+(and not multiple ones for each of the work products which are maintained in seperate tools).
 Baselines are created by using the GitHub "tag" function. The tag name shall correspond to
 the release branch name the tag is created, adding patch version and pre-release tag.
 See also :need:`doc__platform_release_management_plan`.
 
+As described in "Identification and Properties" above, there are several repositories for the modules and the platform integration.
+Baselines are created individually in these repositories, even a different version schema could be adopted.
+In case of dependent repositories, the repository dependet upon on has to be baselined first, to be available
+to refer to this baseline when integrating it. That means that for example a platform baseline also
+documents the versions(baselines) of the modules the platform consists of. This can then also be seen in the platform release note.
```
masc2023: add blank between versions and (baseline)

docs/platform_management_plan/config_management.rst
```diff
@@ -123,3 +137,27 @@ Backup and Recovery
...
+
+Status and Reporting
+^^^^^^^^^^^^^^^^^^^^
+
+Every work product defined in our proceses has a "status" attribute. These are used to communicate to all the stakeholders.
+The main communication means is a document list containing all documents and workproducts including their status.
+This list is typically part of the Documentation Management Plan :need:`doc__documentation_mgt_plan` as part of the Platform Management Plan,
+as defined in :need:`PROCESS_gd_guidl__documentation`.
+Completeness of the configuration items (within a baseline) is checked at least for every release
+against the list of planned documents and workproducts, which is also part of the Documentation Management Plan.
```
masc2023: so the planned safety work products will be also there and not in the safety plan?

masc2023: what is the difference between work products and documents? Tools are also config items, may link here to tool management.

docs/platform_management_plan/config_management.rst
aschemmel-tech: see below starting from line 125, but will add clarifying half sentence

docs/platform_management_plan/config_management.rst
aschemmel-tech: ok

docs/platform_management_plan/config_management.rst
aschemmel-tech: I agreed with Volker that we would like to have one Document list which contains all documents safety/security/QM and refer to this from the safety/security/QM plans. But this is actually a pending change and not in place.

aschemmel-tech: Currently the tool process defines doc__tool__* needs. So currently we would see these also in the Documentation Management Plan. Do you want to change this to an own need definition? And work products in this context I meant all the needs realizing a wp_ which are not a doc_ (e.g. feat_req_) - added some explanation

antonkri (dismissed):

docs/platform_management_plan/config_management.rst
masc2023: For me, we need also add there a doc_ header, as it fulfils wp__requirements__feat, so I see only documents

masc2023: But we need to define document, for documents are need: doc_   , realizes wp____, and see comment below, including tools

aschemmel-tech: So the proposal is to have for example the requirements in a file and this file has also a doc_ header so we have this in the Doc list instead of the requirements needs and maybe the status of this document generated automatically from the requirements it contains?

masc2023: Yes, e.g. also feature requests have this header, so we can just collect every ..document, all should be linked to wp_.... somehow, including tool, planes, etc.

docs/platform_management_plan/config_management.rst
```diff
@@ -123,3 +137,35 @@ Backup and Recovery
...
+Configuration Management Tooling
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Almost all requirements of the standards towards configuration management can be covered by
+standard versioning tooling of the Eclipse Foundation and of the S-CORE project
+("structured text" identification of work products).
+The respective tools used in the project are:
+
+* versioning: GitHub
+* structured text: sphinx-needs
```
masc2023, line 170: is sphinx feature, not needs specific

masc2023 (approved):
Fine for now

PR #1519, opened 2025-07-28
docs: Add module architecture diagram

ramceb:


arsibo (approved):
fine. Maybe we should mentioned that it is in draft state

PR #1529, opened 2025-07-30
Adapt Score to Docs-As-Code 1.0 Release

MaximilianSoerenPollak:


MaximilianSoerenPollak:
## The documentation does not build due to the building command changing. Therefore we need to change the workflow. In the next build / PR's it should work again.
## All commands are working and it builds without issue. 

bazel run //:docs 
image

bazel run //:live_preview
image

bazel build //:needs_json
image

bazel run //:ide_support
image

PR #1531, opened 2025-07-31
upgrade module versions

masc2023:
doc-as-code
process description
python basics


MaximilianSoerenPollak:
#1529 will bring this change, as it requires some more changes.

PR #1539, opened 2025-08-01
docs: add Result Library requirements

fbaeuerle:
Add requirements for the result library.

### DIFF
--- a/docs/modules/baselibs/result/docs/requirements.rst
+++ b/docs/modules/baselibs/result/docs/requirements.rst
@@ -1,0 +1,128 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2025 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+Requirements
+############
+
+.. document:: Result Library Requirements
+   :id: doc__result_lib_requirements
+   :status: draft
+   :safety: ASIL_B
+   :realizes: PROCESS_wp__requirements__comp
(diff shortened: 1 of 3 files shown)


docs/modules/baselibs/result/docs/requirements.rst
```diff
@@ -0,0 +1,128 @@
...
+
+.. comp_req:: Exception-Free Operation
+   :id: comp_req__result__exception_free_operation
+   :reqtype: Non-Functional
+   :security: NO
+   :safety: ASIL_B
+   :satisfies: feat_req__baselibs__core_utilities, feat_req__baselibs__safety
+   :status: valid
+
+   The Result library shall operate without throwing C++ exceptions.
```
hth9fe, line 89: I guess you mean, that no C++ exception explicitly should be thrown by the library. I guess in addition the code might still need to take measures to "catch" C++ exception which might occur cause of implementation, right?

hth9fe:
Do we really need a separate library for "Results"? Or would this be more likely be part of a utilities/core library?

fbaeuerle:
> Do we really need a separate library for "Results"? Or would this be more likely be part of a utilities/core library?

It is part of [eclipse-score/baselibs](https://github.com/eclipse-score/baselibs/tree/main/score/result). The requirements are a documentation of the current state of implementation.

hth9fe:
> > Do we really need a separate library for "Results"? Or would this be more likely be part of a utilities/core library?
> 
> It is part of [eclipse-score/baselibs](https://github.com/eclipse-score/baselibs/tree/main/score/result). The requirements are a documentation of the current state of implementation.

but then isn't it more a Result class and not a library?

fbaeuerle:
> > > Do we really need a separate library for "Results"? Or would this be more likely be part of a utilities/core library?
> > 
> > 
> > It is part of [eclipse-score/baselibs](https://github.com/eclipse-score/baselibs/tree/main/score/result). The requirements are a documentation of the current state of implementation.
> 
> but then isn't it more a Result class and not a library?

The result class needs to interact with a few more classes that reside in [baselibs/score/result](https://github.com/eclipse-score/baselibs/tree/main/score/result), so i would argue it is an (admittedly small) library.

PR #1606, opened 2025-08-11
Feature Request: Record and Replay

### LINKED ISSUE #1167: Feature Request for Record & Replay
Discussed in https://github.com/orgs/eclipse-score/discussions/472#discussioncomment-13271395

Acceptance Criteria:
- Feature Request written according to the [Change Management](https://eclipse-score.github.io/score/main/process/process_areas/change_management/change_management_concept.html) & [Feature Request Template](https://eclipse-score.github.io/score/main/process/process_areas/change_management/guidance/change_management_feature_template.html)
- Feature requirements written according to the [Requirements Engineering](https://eclipse-score.github.io/score/main/process/process_areas/requirements_engineering/requirements_concept.html)
- If necessary: extend the stakeholder requirements written according to the [Requirements Engineering](https://eclipse-score.github.io/score/main/process/process_areas/requirements_engineering/requirements_concept.html)

johannes-esr:


docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+To activate this feature, use the following feature flag:
+
+``experimental_record_and_replay``
+
+
+Abstract
+========
+
+Recording is a function allowing to record a trace, called a *Trace Record*, of
+the activity of the system while it is executing. Replay is a function which
```
a-zw, line 38: I think we can be more precise here.

Instead of "system", I would propose "constrained set of FEO activities". This implies we can ignore the process and task-chain stuff for Replay/Simulation except for some scheduling constraints (like order and triggers).

Instead of "activity", we could write "in- and outputs". While this still ignores the internal state aspect, I'd argue any solution turns the internal state into "in- and output" anyways.

docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+|reprocessing|
+
+To reproduce the internal behavior of the SUR when reprocessing, its original
+internal state at the beginning of the recording must be available to properly
+initialize the SUR. Consequently, the original internal state must be part of
+the Trace Record.
+
+A trace can be recorded from Reprocessing in a similar way as from a SUR. That
+Trace Record can be used for both use-cases, replay and reprocessing.
+
```
a-zw, line 122: As an additional aspect: This allows to verify if an activity behaves deterministically. Reprocess it two or more times and check if the outputs are identical.

docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+     precise requirement, is that the recorded initialization information and the
+     recordings from mw/com shall *fit together*. That, however, is very context dependent
+     and might be an ill-posed problem for our scope.
+* Activities (replace with the correct term from FEO denoting an **execution
+  unit**) shall be able and required to register (with a recording daemon) their
+  internal state needed to reestablish its state for initialization for
+  reprocessing.
+* The Trace Record shall resemble the time-order of messages as seen by FEO at
+  execution time. (Do we need that requirement?)
+
```
a-zw, line 160: It can be useful to trim recordings. Examples:

- Remove outputs to save space (assuming you are sufficiently confident about the determinism of the SUR)
- Remove confidential parts and give recordings to third-parties
- Cut out pieces by time. An application engineer records 3h of driving data, then you only use minute 73 for your regression test suite as it contains an interesting situation.

docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+     recording from mw/com start with execution step n. A more relaxed, but less
+     precise requirement, is that the recorded initialization information and the
+     recordings from mw/com shall *fit together*. That, however, is very context dependent
+     and might be an ill-posed problem for our scope.
+* Activities (replace with the correct term from FEO denoting an **execution
+  unit**) shall be able and required to register (with a recording daemon) their
+  internal state needed to reestablish its state for initialization for
+  reprocessing.
+* The Trace Record shall resemble the time-order of messages as seen by FEO at
+  execution time. (Do we need that requirement?)
```
a-zw, line 159: It seems to be a pointless restriction to me. How does it derive from the Rationale?

Maybe something else is behind this requirement: We need good performance such that the data format matters? Then we should rather require that specifically.

docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+  initialize the SUR for reprocessing to the exact state it has when recording
+  started.
+
+  .. note::
+     A possible design would be to have a header part of the Trace Record which stores the
+     initialization information, followed by the trace recording.
+
+     |trace|
+
+* Initialization information shall only be written once to the Trace Record.
```
a-zw, line 144: It is useful to have more than a single initialization info just like movie formats have regular key frames. It allows to skip forwards and backwards efficiently. With hours of recordings that can save a lot of time (or compute power).

It is possible to work with a single init info though but we should not restrict ourselves at the requirements level here. At least, I don't see how it follows from the Rationale.

docs/features/recording-and-replay/index.rst
a-zw, line 160: It is useful to store all kinds of meta information. Some once for the whole file (version info, comments from test drivers, etc). Some as time series additional to FEO in- and outputs (additional ground truth info like GPS info and extra sensors).

a-zw, line 160: For compatibility reasons a _self-descriptive_ file format seems useful to me. For example, ProtoBuf is not self-descriptive because you always need to know the schema from somewhere else. In my experience it is tough to establish the necessary discipline to ensure that in large projects.

docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+* The S-CORE stack shall provide the possibility to execute in an artificial
+  environment in which inputs are provided from a Trace Record. This execution
+  mode is called **Reprocessing**.
+* Reprocessing shall use already established communication interfaces. A
+  special reprocessing input path for data is to be avoided.
+* On initialization of Reprocessing all activities, for which initialization
+  information is available in the Trace Record, shall be restored to that state.
+* All output interfaces shall publish in reprocessing such that a **reprocessed
+  trace** can be recorded.
+
```
a-zw, line 174: Some visualization tools support "live" use. For example, feeding virtual data into the SUR. Currently, I don't see any danger that we violate it but maybe making that an explicit requirement could become useful.

docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+  .. note::
+     Ideally, the initialization information is recorded from execution step n-1 and
+     recording from mw/com start with execution step n. A more relaxed, but less
+     precise requirement, is that the recorded initialization information and the
+     recordings from mw/com shall *fit together*. That, however, is very context dependent
+     and might be an ill-posed problem for our scope.
+* Activities (replace with the correct term from FEO denoting an **execution
+  unit**) shall be able and required to register (with a recording daemon) their
+  internal state needed to reestablish its state for initialization for
+  reprocessing.
```
a-zw, line 157: Is it ok to leave the serialization format unspecified?

Somehow it should be ensured that IPC communication and recording format will stay consistent. Otherwise trace recording might eventually require a separate serialization and that would probably a prohibitive performance cost at runtime.

docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+
+Recording
+---------
+Recording is the activity of capturing the input, output, and possibly the
+internal state of a defined system, the System Under Recording (SUR).
+
+|record|
+
+* **Input:** The blue box depicts all parts of the larger system which generates
+  inputs consumed by the SUR via its input interfaces.
```
a-zw, line 74: What exactly are the "input interfaces"? Naively it would just be communication / IPC.

However, there are other inputs which are relevant for determinism: What about time, file systems, ConfigProvider, floating point accuracy, shared memory, GPU scheduling, CPU info, GPIO, POSIX signal handlers, scheduling jitter, build configuration, Key-Value-Storage?

docs/features/recording-and-replay/index.rst
```diff
@@ -0,0 +1,274 @@
...
+* **System under Record:** The System Under Record (SUR), green box, is the
+  system of which the runtime behavior shall be recorded. It consumes its
+  inputs, executes on them, and generates its outputs. While executing, the SUR
+  might acquire internal state.
+* **Recorder:** The Recorder is a function able to read inputs, outputs, and
+  possibly the internal state of the SUR and writes them time-ordered to a file,
```
a-zw, line 80: While we want to record internal state, in practice it might mean "not really all" internal state. Developers often need the ability to control what internal state is recorded and what is not. Of course, that requires a careful assessment to stay deterministic.

The essential question is the default. Record everything by default (risking performance) or only record explicitly marked things (risking replay correctness).

docs/features/recording-and-replay/index.rst
wmader23: Ad "system"
I am not sure if we can ignore process and task-chains. As soon as you have more than one FEO activity, you need to know how these are chained, don't you. You can replicate that for the replay/reprocessing situation "out-of-line" with the actual deployment, but somehow things need to be connected. But I agree, system might to too lazy terminology.

Ad "activity vs in- and outputs"
I agree to that. In replay you can only observe inputs and outputs. Only in reprocessing, you can also observe all execution-time aspects.

docs/features/recording-and-replay/index.rst
wmader23: Yes, assuming the entire reprocessing machinery behaves deterministically in relevant aspects.

docs/features/recording-and-replay/index.rst
wmader23: I take that as a yes, we want that requirement.

docs/features/recording-and-replay/index.rst
wmader23: Sorry, which restriction are you referring to?

docs/features/recording-and-replay/index.rst
wmader23: Valid. My motivation was to keep the trace files smaller. We can request the condition/frequency at which internal state is written to the recording to be a configuration option of the recording system.

docs/features/recording-and-replay/index.rst
wmader23: I will add that, but I don't know if that is technically possible.

docs/features/recording-and-replay/index.rst
wmader23: You refer to tooling visualizing the live output of the SUR, right? Since the requirements states that the usual output interfaces shall be active also in reprocessing, I would conclude, that any for-deployment tooling (i.e., in-car tooling) also can be connected to a reprocessing session. Does that work for you?

docs/features/recording-and-replay/index.rst
wmader23: Additional serialization costs are to be avoided. Maybe we need to further discuss how to streamline that in the best possible way.

docs/features/recording-and-replay/index.rst
wmader23: Yes, there are trade-offs to be made. Let's discuss.

docs/features/recording-and-replay/index.rst
wmader23: Further down there is the requirement

> Activities shall be able and required to register (with a recording daemon) their internal state needed to reestablish its state for initialization for reprocessing.

As you say, it's specific to the activity what needs recording. I would leave that up to the developers of that activity.

docs/features/recording-and-replay/index.rst
a-zw: This comment was not related to the "Do we need that requirement?" question. It is a proposal for an additional bullet point. Like this:

```suggestion
* The Trace Record format shall support efficient cutting by time, removal of individual signals, and removal of individual ports.
```

docs/features/recording-and-replay/index.rst
a-zw: The restriction is "resemble the time-order of messages as seen by FEO". At least, I understand this is a restriction on the trace record data format.

Functionally, we only need to ability to _recreate_ the time-order but there is no need to _store_ it in that order.

docs/features/recording-and-replay/index.rst
a-zw: It certainly does conflict with other needs like keeping the trace files small.

JSON and XML are the most well-known self-descriptive data formats but they are not suitable here. There are [binary alternatives](https://en.wikipedia.org/wiki/Comparison_of_data-serialization_formats#Comparison_of_binary_formats).

docs/features/recording-and-replay/index.rst
a-zw: Well, we could do _no_ serialization at all at runtime and shift all of that to some post-processing.

docs/features/recording-and-replay/index.rst
wmader23: OK, understood. I will clarify.

a-zw:
I'm not really familiar with S-CORE bureaucracy, but since this pull request isn't merged yet, it missed some kind of deadline for feature proposals, didn't it?

masc2023:
> I'm not really familiar with S-CORE bureaucracy, but since this pull request isn't merged yet, it missed some kind of deadline for feature proposals, didn't it?

Has this PR something todo with the Feature Request here: https://github.com/eclipse-score/score/issues/1167?
Then you may link it and ask for review, but first you should solve your documentation build error

FScholPer:
Added to the feature request. @johannes-esr  can you please take a look

qor-lb:
@wmader23 / @johannes-esr / @LittleHuba is this Feature Request still in progress or ready for review in the architecture community?

LittleHuba:
I'm only in a support role and do not actively drive this.

FScholPer:
@johannes-esr please update this pull request that it can be merged

FScholPer:
@johannes-esr please update or close

PR #1622, opened 2025-08-18
Unify devcontainers

lurtz:
Linux / WSL and Docker on Windows use different scripting languages and this shows in the initializeCommand. With this solution the same devcontainer.json file can be used on all platforms, which reduces maintenance costs.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,7 @@
 {
     "name": "eclipse-s-core",
     "image": "ghcr.io/eclipse-score/devcontainer:latest",
+    // This will run on Linux `initialize_command` and on Windows `initialize_command.bat`
     "initializeCommand": [
         ".devcontainer/initialize_command",
         "${localEnv:HOME}"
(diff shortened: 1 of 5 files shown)


lurtz:
This approach was tested at https://github.com/eclipse-score/baselibs/pull/15#issuecomment-3197506987

.devcontainer/devcontainer.json
```diff
@@ -1,5 +1,8 @@
 {
     "name": "eclipse-s-core",
     "image": "ghcr.io/eclipse-score/devcontainer:latest",
-    "initializeCommand": "mkdir -p ${localEnv:HOME}/.cache/bazel"
+    "initializeCommand": [
+        ".devcontainer/initialize_command",
```
AlexanderLanin, line 6: As we cannot add comments here?!

.devcontainer/initialize_command
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
AlexanderLanin, line 1: Maybe add a comment and link to to something like `eclipse-score/devcontainer/initialize_command.md`? For explanation of file selection magic.

AlexanderLanin (approved):

.devcontainer/devcontainer.json
lurtz: I added a comment in the json file. This seems to be supported by the devcontainer

PR #1628, opened 2025-08-19
WSL2 settings recommendation added

bl81hi:
## Description

Added a short recommendation how to put settings for WSL2.

### DIFF
--- a/docs/contribute/development/development_environment.rst
+++ b/docs/contribute/development/development_environment.rst
@@ -148,3 +148,13 @@ IDE Guides
 
    ide/vscode
    ide/clion
+
+WSL2 Settings
+=============
+If you are using a Linux box running on WSL2, the following minimal settings in your .wslconfig file are recommended:
+
+.. code-block:: shell
+
+   [wsl2]
+   swap=8GB
+   memory=16GB


MaximilianSoerenPollak:
We probably should make a comment somewhere (although I don't think this belongs in the RST) that there is a very long standing bug in WSL2 for Windows 10&11 that enduces a memory leak and cpu hogging. 
It can be that WSL2 takes up to 20-30GB (if allowed) of Memory and quiet a bit of CPU as well. 

This happens when you have WSL open and lock your PC. If you log back in and resume your session the leak starts to manifest (as far  as I can remember)

The fix is to shut it down and restart `wsl --shutdown` & `wsl` inside of powershell. 

I will search the coming days if I can find the discussion / issue that I'm refering to.

bl81hi:
> We probably should make a comment somewhere (although I don't think this belongs in the RST) that there is a very long standing bug in WSL2 for Windows 10&11 that enduces a memory leak and cpu hogging. It can be that WSL2 takes up to 20-30GB (if allowed) of Memory and quiet a bit of CPU as well.

What level of detail is targeted in this documentation? Is it intended to be short, or can it be more comprehensive? If it is more comprehensive, we could add hints on bugs or add additional information (e.g. how to connect to WSL from the IDE). That could feel too chatty for experienced developers, but might be helpful for less experienced people working on the project. As we use X-as-Code, we could expect that people who are actually used to other, "non-code tools" would work here, too.

What do you think?

docs/contribute/development/development_environment.rst
```diff
@@ -148,3 +148,13 @@ IDE Guides
 
    ide/vscode
    ide/clion
+
+WSL2 Settings
+=============
+If you are using a Linux box running on WSL2, the following minimal settings in your .wslconfig file are recommended:
```
AlexanderLanin: Maybe we can rephrase this and merge, so we don't need to wait for @MaximilianSoerenPollak? Currently it's difficult to tell when out-of-memory happens, as no-one has complained yet. I assume memory becomes an issue when you run an (old) C++ linker?!

```suggestion
If you encounter memory issues on WSL2, the following minimal settings in your .wslconfig file are recommended:
```

MaximilianSoerenPollak: The settings do not solve this issue. So to speak, once the leak happens and stays for a while the performance goes into the tank, even if it stays inside the parameters. 

But I do think we should have some sort of short info that one can set such settings?

MaximilianSoerenPollak:
> > We probably should make a comment somewhere (although I don't think this belongs in the RST) that there is a very long standing bug in WSL2 for Windows 10&11 that enduces a memory leak and cpu hogging. It can be that WSL2 takes up to 20-30GB (if allowed) of Memory and quiet a bit of CPU as well.
> 
> What level of detail is targeted in this documentation? Is it intended to be short, or can it be more comprehensive? If it is more comprehensive, we could add hints on bugs or add additional information (e.g. how to connect to WSL from the IDE). That could feel too chatty for experienced developers, but might be helpful for less experienced people working on the project. As we use X-as-Code, we could expect that people who are actually used to other, "non-code tools" would work here, too.
> 
> What do you think?

I think the detail you have is good, I wouldn't go too into detail imo.

docs/contribute/development/development_environment.rst
AlexanderLanin: We have the info about settings here. Info about the leak we can publish once people start complaining :)

AlexanderLanin (approved):

PR #1408, opened 2025-07-14
doc: added adr template (#1383)

FScholPer:
see [#1383](https://github.com/eclipse-score/score/issues/1383)

Frank Scholter Peres <frank.scholter_peres@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

PR #1644, opened 2025-08-27
Setup area for infrastructure documentation

AlexanderLanin:
As this involves a crazy amount of reviewers, I'm creating a separate PR without actual content just to setup an area for infrastructure documentation.

The goal here is for the infrastructure community to have a place where we can document stuff. Once we have a better understanding of the actual content, it will be easier to decide how exactly we want to structure it.


docs/infrastructure/index.rst
```diff
@@ -0,0 +1,33 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2024 Contributors to the Eclipse Foundation
```
masc2023: copyright 2025?

FScholPer:
I would prefer to locate DRs in a central place like in my pull request which is currently blocked: https://github.com/eclipse-score/score/pull/1408

Is it only documentation or is it also decisions?

masc2023:
> I would prefer to locate DRs in a central place like in my pull request which is currently blocked: #1408
> 
> Is it only documentation or is it also decisions?

docs/requirements or docs/infrastructure seems for me same central places, then I would propose to add it here

/docs/requirements/stakeholder/       
/docs/requirements/tool/ 
**/docs/requirements/design_decisions**

AlexanderLanin:
For now my need is to document decision records. Some are about design, but not all.

I don't have the text at hand, so I'll just copy a screenshot as an example :)

example decision records

AlexanderLanin:
> We have two people announced to leave the project, why keep them in?

Leaving a team within their company does not affect committer status within open source.

masc2023:
> > We have two people announced to leave the project, why keep them in?
> 
> Leaving a team within their company does not affect committer status within open source

But it does not help, if people are not any more in project are gating people and they will not show-up any more, I think we need discussion here
We are using here the CODEOWNER file to implement ROLE definitions as defined in Process Description workflow, their commiter status has nothing todo with that here

masc2023 (approved):
Fine as initial version for Design Decision, concerning Codeowner, Tech Leads must decide

PR #1687, opened 2025-09-01
DR-001-Infra: Integration Strategy for External Development Tools

AlexanderLanin:
Rendered: https://eclipse-score.github.io/score/pr-1687/design_decisions/DR-001-infra.html


docs/design_decisions/DR-001-infra.md
```diff
@@ -0,0 +1,176 @@
...
+
+We use a broad set of tools (formatters, linters, static analyzers, code generators, license/SBOM and security scanners, doc builders) across multiple repositories. Today, tools are invoked via Bazel, IDEs, and CI with varying levels of pinning and caching. We need a standard approach that runs **identically** in CI, local CLI, and IDE/editor integrations, and that fits ISO 26262/ASPICE constraints (version pinning, determinism, long-term retention, offline path).
+
+## 2. Requirements
+
+1. **Version Pinning**: Identical tool versions across CLI/CI/IDE; no hidden network during execution; artifacts hash-pinned.
+2. **Config Consistency**: Central baseline config; explicit, reviewable local overrides with drift detection.
+3. **Performance**: Fast full and incremental runs; leverage remote cache/execution for heavy, Bazel-config-dependent tools.
+4. **Offline & Retention**: Prefetch & store artifacts for ≥10 years; fully offline execution path.
+5. **Platforms**: Ubuntu x86_64 + additional to be defined, e.g. Alpine, arm64 etc.
```
opajonk, line 33: This should refer to the platform definition document once it is established. This may or may not happen before this PR is complete ;-)

docs/design_decisions/DR-001-infra.md
```diff
@@ -0,0 +1,176 @@
...
+
+More interestingly there are cases where the exact version is less important (python) or
+is managed outside bazel anyway (python requirements.txt files). In such cases native
+installs have no disadvantages and are significantly easier (development) and faster (runtime).
+
+
+## 5. Conclusion
+
+- **Devcontainer** is the primary distribution for all tools, ensuring consistency and ease of onboarding.
+- The same devcontainer (possibly stripped down) is used in CI for reproducibility.
```
opajonk, line 135:
```suggestion
- The same devcontainer (possibly stripped down) is used in CI for reproducibility.
  However, it must be ensured that the devcontainer strictly builds _on top_ of the CI container to avoid that CI issues cannot be reproduced locally.
```

docs/design_decisions/DR-001-infra.md
```diff
@@ -0,0 +1,176 @@
...
+- **Bazel aspects** are used for tools that benefit from Bazel-config-dependent analysis or caching. Custom Bazel rules are possible but not actively pursued.
+- For small/fast CI jobs, **native installs** are preferred due to speed, especially when version pinning is less critical (see Deep Dive 4.3).
```
opajonk, line 137: Could be, but I would first try containers. Maybe CI + devcontainer on top. One problem, one solution. Otherwise, I can already see the "oops we forgot to bump this tool here - **that** is why it broke!!" issues... after many hours of investigation.

docs/design_decisions/DR-001-infra.md
AlexanderLanin: Yeah sure! We can add a reference afterwards. But it's not critical for the decision here.

docs/design_decisions/DR-001-infra.md
AlexanderLanin: Versioning etc I would like to pull to the next level of detail, when we talk about how to implement this DR.

docs/design_decisions/DR-001-infra.md
AlexanderLanin: I would like to pull details to the next level of detail, when we talk about how to implement this DR. I'll add this without explicitly specifying *on top*.

PR #1689, opened 2025-09-01
DR-002-Infra: Integration Testing in a Distributed Monolith

AlexanderLanin:
# DR-002 (Integration)

Large systems often span multiple repositories. Each repository can look “green” on its own, yet problems only show up when everything is combined. These late surprises slow down development and make debugging painful.  

DR-002 turns a collection of separate repositories into a system that behaves like a single, continuously tested whole — ensuring the main line is always integrable across all components.

## Proposed Approach
- Every change in any repository is tested **in combination with the rest of the system**, not just in isolation.  
- There are **two testing layers**:  
  - a **fast feedback loop** (lightweight tests that run on every pull request),  
  - and a **deeper validation** (heavier tests run after merges or on a schedule).  
- This setup guarantees that developers can trust the system as a whole to consistently work.

## Benefits
- Problems across repositories are caught early.  
- Developers spend less time coordinating merges (“merge after me” scenarios disappear).  
- The project always has a “known good” baseline to fall back on, enabling stability while still moving fast.  

---

Note: this concept is easily extendable to support multiple versions of S-CORE. But that's currently not required.

---

Rendered: https://eclipse-score.github.io/score/pr-1689/design_decisions/DR-002-infra.html

### DIFF
--- a/docs/design_decisions/DR-002-infra.md
+++ b/docs/design_decisions/DR-002-infra.md
@@ -1,0 +1,330 @@
+<!--
+Copyright (c) 2025 Contributors to the Eclipse Foundation
+
+See the NOTICE file(s) distributed with this work for additional
+information regarding copyright ownership.
+
+This program and the accompanying materials are made available under the
+terms of the Apache License Version 2.0 which is available at
+https://www.apache.org/licenses/LICENSE-2.0
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# DR-002-Infra: Integration Testing in a Distributed Monolith
+
+* **Status:** Draft. Bazel integration missing.
+* **Owner:** Infrastructure Community
+* **Date:** 2025-09-01
+
+---
+Teams often split what is functionally a single system across many repositories. Each
+repository can show a green build while the assembled system is already broken. This
+article looks at how to bring system-level feedback earlier when you work that way. This
+article does not argue for pull requests, trunk-based development, or continuous
+integration itself. Those are well covered elsewhere. It also does not look into any
+specific tools or implementations for achieving these practices - except for providing a
+GitHub based example.
+
+The context here assumes three things: you develop through pull requests with required
+checks; you have multiple interdependent repositories that ship together; and you either
+have or will create a central integration repository used only for orchestration. If any
+of those are absent you will need to establish them first; the rest of the discussion
+builds on them.
+
+## Where Problems Usually Appear
+An interface change (for example a renamed field in a shared schema) is updated in two
+direct consumers. Their pull requests pass. Another consumer several repositories away
+still depends on the old interface and only fails once the whole set of changes reaches
+main and a later integration run executes. The defect was present early but only visible
+late. Investigation now needs cross-repo log hunting instead of a quick fix while the
+change was still in flight.
+
+Running full end-to-end environments on every pull request is rarely affordable.
+Coordinated multi-repository changes are then handled informally through ad-hoc
+ordering: “merge yours after mine”. Late detection raises cost and makes regression
+origins harder to locate.
+
+## Core Concepts
+We model the integrated system as an explicit set of (component, commit) pairs captured
+in a manifest. Manifests are derived deterministically from events: a single pull
+request, a coordinated group of pull requests, or a post-merge refresh. A curated fast
+subset of integration tests provides pre-merge feedback; a deeper suite runs after
+merge. Passing suites produce a recorded manifest (“known good”). Coordinated
+multi-repository change is treated as a first-class case—we validate the set as a unit
+rather than relying on merge ordering.
+
+Terminology (brief):
+* Component - repository that participates in the assembled product (e.g. service API
+  repo, shared library).
+* Fast subset - curated integration tests finishing in single-digit minutes (protocol
+  seams, migration boundaries, adapters).
+* Tuple - mapping of component names to commit SHAs for one integrated build (e.g. {
+  users: a1c3f9d, billing: 9e02b4c }).
+* Known good - tuple + metadata (timestamp, suite, manifest hash) stored for later
+  reproduction.
+
+History & context: classic continuous integration assumed a single codebase; splitting
+one system across repositories reintroduces coordination issues CI was intended to
+remove. This adapts familiar CI principles (frequent integration, fast feedback,
+reproducibility) to a multi-repository boundary. The central integration repository is a
+neutral place to define participating components, build manifests, hold
+integration-specific helpers (overrides, fixtures, seam tests), and persist known-good
+records. It should not contain business logic; keeping it lean reduces accidental
+coupling and simplifies review.
+
+## Integration Workflows
+We use three recurring workflows: a single pull request, a coordinated subset when
+multiple pull requests must land together, and a post-merge fuller suite. Each produces
+a manifest, runs an appropriate depth of tests, and may record the tuple if successful.
+
+### Visual Overview
+```mermaid
+flowchart TB
+  subgraph COMP[Component Repos]
+    pr[PR opened / updated<br/>&lt;event&gt;]:::event --> comp_ci[Component tests]:::step
+
+    trigger1[Merge to main<br/>&lt;event&gt;]:::event
+  end
+
+  subgraph INT[Integration Repo]
+    comp_ci --> |dispatch|detect_changeset[Detect multi repository PRs]:::step
+    knownGood[(Known good store)]:::artifact
+
+    %% PR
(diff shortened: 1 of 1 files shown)


docs/design_decisions/DR-002-infra.md
```diff
@@ -0,0 +1,330 @@
...
+  repository_dispatch:
+    types: [component-merged]
+jobs:
+  full-suite:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Generate new last_known_good.yaml
+        run: update_last_known_good.py last_known_good.yaml > last_known_good.yaml
```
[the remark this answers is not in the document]
AlexanderLanin: it's pseudocode

docs/design_decisions/DR-002-infra.md
```diff
@@ -0,0 +1,342 @@
...
+specific tools or implementations for achieving these practices - except for providing a
+GitHub based example.
+
+The context here assumes three things: you develop through pull requests with required
+checks; you have multiple interdependent repositories that ship together; and you either
+have or will create a central integration repository used only for orchestration. If any
+of those are absent you will need to establish them first; the rest of the discussion
+builds on them.
+
+## Where Problems Usually Appear
```
AlexanderLanin:
```suggestion
## Motivation (Where Problems Usually Appear)
```

AlexanderLanin:
Basically agreed in infrastructure planning session. Feel free to document your opinion here formally with a PR review!

thilo-schmitt:
Overall, this reads very well. I'm in support of this approach.

However, what I still don’t understand is how “coordinated multi-repo” works. Yes, I understand that the individual PRs shall be tagged with a common label. But how does the integration CI know that all PRs with that label are now present — more could still be added (or removed again, e.g. if I accidentally applied the wrong label and change it then); when is the changeset considered complete and the integration pipeline can run? Does it need a manual trigger? Do you use a time window of x minutes? Does the pipeline simply run for each individual PR in the coordinated set and simply fails n-1 times until the whole set (of n PRs) is there?

And then it would be nice if one could merge the entire coordinated set automatically with "one click" (if everything is green in each component as well as in the integration). In any case, we should strive for a way that ensures everything belonging to the coordinated set gets merged and nothing is forgotten, so that inconsistencies and non-"integrability" don’t creep back in. Maybe if you merge in one repo, then all PRs in all other repos that belong to the coordinated set are merged too?!

opajonk:
> Overall, this reads very well. I'm in support of this approach.
> 
> However, what I still don’t understand is how “coordinated multi-repo” works. Yes, I understand that the individual PRs shall be tagged with a common label. But how does the integration CI know that all PRs with that label are now present — more could still be added (or removed again, e.g. if I accidentally applied the wrong label and change it then); when is the changeset considered complete and the integration pipeline can run? Does it need a manual trigger? Do you use a time window of x minutes? Does the pipeline simply run for each individual PR in the coordinated set and simply fails n-1 times until the whole set (of n PRs) is there?
> 

Good points. There are some ways to do this technically in GitHub. I know it for "enterprise", I suspect it for "public". I would use "PR with head branch & base branch the same" as unique identifier. It can be the same on all participating repositories, can even be created automatically. Example:

- I open a PR on communication with `HEAD` `foo` and `BASE` `main` (draft as a start) and do some work.
- automation creates a draft PR on `reference_integration`. Due to that PR, CI runs whatever is required there. Automation creates feedback to my PR on communication. That feedback is made a required check to merge to main.
- It can now be that e.g. I need to open another PR with `HEAD` `foo` and `BASE` `main` on "baselibs" to fix a compilation / test failure on `reference_integration`. OK, exactly what I wanted.

(very rough sketch, just to convey the idea)

> And then it would be nice if one could merge the entire coordinated set automatically with "one click" (if everything is green in each component as well as in the integration). In any case, we should strive for a way that ensures everything belonging to the coordinated set gets merged and nothing is forgotten, so that inconsistencies and non-"integrability" don’t creep back in. Maybe if you merge in one repo, then all PRs in all other repos that belong to the coordinated set are merged too?!

This is very tricky. In fact, we are right now trying to implement something like that internally on GitHub Enterprise - using only GitHub methods. I cannot yet say that this works; it's like an **atomic merge** which involves a set of "the same series of PRs". Ideally a merge queue, so that long-running CI is not a problem. Maybe it requires a third-party tool (some GitHub Bot) or some additional requirement (e.g. you must use a **singular** integration repository). I can report when I have results here.

(What I wrote here IMHO only expands on the proposal, should in no way contradict it. If it does, that is my lacking ability to explain my thoughts ;-))

docs/design_decisions/DR-002-infra.md
```diff
@@ -0,0 +1,372 @@
...
+# DR-002-Infra: Integration Testing in a Distributed Monolith
+
+* **Status:** Agreed within Community
+* **Owner:** Infrastructure Community
+* **Date:** 2025-09-01
+
+---
+## Executive Summary
+
+Large systems often span multiple repositories. Each repository can look “green” on its own, yet problems only show up when everything is combined. These late surprises slow down development and make debugging painful.
```
opajonk, line 23:
```suggestion
Large systems often span multiple repositories. Each repository can look “green” on its own, yet problems only show up when everything is combined. These late surprises slow down development and make debugging painful. They can even block releases.
```

docs/design_decisions/DR-002-infra.md
```diff
@@ -0,0 +1,372 @@
...
+* Known good - tuple + metadata (timestamp, suite, manifest hash) stored for later
+  reproduction.
+
+History & context: classic continuous integration assumed a single codebase; splitting
+one system across repositories reintroduces coordination issues CI was intended to
+remove. This adapts familiar CI principles (frequent integration, fast feedback,
+reproducibility) to a multi-repository boundary. The central integration repository is a
+neutral place to define participating components, build manifests, hold
+integration-specific helpers (overrides, fixtures, seam tests), and persist known-good
+records. It should not contain business logic; keeping it lean reduces accidental
```
opajonk, line 98: I would argue that the integration repository can contain e.g. integration tests or other checks which make sense only in that specific integration context. Developing "integration tests" in another repository would be very confusing.

opajonk (approved):
I am in favor of this. I have added minor remarks, but these are not critical.

Technicality: for reviews and other reasons, the [one sentence per line rule](https://jeffkreeftmeijer.com/one-sentence-per-line/) is a very helpful one. But this is clearly only a _suggestion_.

thilo-schmitt:
> 
>     * I open a PR on communication with `HEAD` `foo` and `BASE` `main` (draft as a start) and do some work.
> 
>     * automation creates a draft PR on `reference_integration`. Due to that PR, CI runs whatever is required there. Automation creates feedback to my PR on communication. That feedback is made a required check to merge to main.
> 
>     * It can now be that e.g. I need to open another PR with `HEAD` `foo` and `BASE` `main` on "baselibs" to fix a compilation / test failure on `reference_integration`. OK, exactly what I wanted.
> 
> (very rough sketch, just to convey the idea)

So, if I understand you correctly, basically what you are saying is: It will fail n-1 times until the n-th required PR is in place. I mean I can somehow live with it, but appears to me to be a waste of resources. 

Maybe we can think of something that signals "this set of PRs is now complete and stable, now the integration pipeline will yield a sensible outcome". Ideally _not (entirely)_ manual, because I still want automation as much as possible.

> it's like an **atomic merge** which involves a set of "the same series of PRs".

Exactly. I had this term in mind but hesitated to write it down, since I don't think it's truly achievable (being truly atomic). But yes, that's probably the best term to use to convey the idea.

opajonk:
> Maybe we can think of something that signals "this set of PRs is now complete and stable, now the integration pipeline will yield a sensible outcome". Ideally _not (entirely)_ manual, because I still want automation as much as possible.

I think as long as the merge requires a passing CI (and you cannot "forget" to request this integration run) - should be fine!

> > it's like an **atomic merge** which involves a set of "the same series of PRs".
> 
> Exactly. I had this term in mind but hesitated to write it down, since I don't think it's truly achievable (being truly atomic). But yes, that's probably the best term to use to convey the idea.

Yes, I also doubt that **truly** atomic is possible. Best that is possible is something like "serializing the integration of groups of PRs via some queue", with "queue halts if there is a problem integrating one of the groups". And hoping that will not happen often. Maybe something is possible around "backing out failed integrations automatically" and "continuing with the next group" - but I can see already the complexity of that code exploding...

docs/design_decisions/DR-002-infra.md
AlexanderLanin: I'll rename "seam tests" to "integration tests"

dcalavrezo-qorix (approved):

PR #1797, opened 2025-09-24
update CODEOWNER

marcmo:
originates from CR #135618


AlexanderLanin (approved):

AlexanderLanin:
Agreed with @dcalavrezo-qorix as well. Templates should not be governed by infrastructure community in our opinion.

Issue #1707, opened 2025-09-08
Improvement: Automate versioning of tools used in score


nicu1989:
### What

Currently the docs-as-code tool version is incorrect because it has to be manually updated: https://github.com/eclipse-score/score/blob/f8ee55add5618a169058752bf25e167f88d38c00/docs/score_tools/doc_as_code.rst?plain=1#L18

### How

Use either existing sphinx extensions or implement the auto update mechanism so that the version is taken from MODULE.bazel: `bazel_dep(name = "score_docs_as_code", version = "1.2.0") `.

### Estimates for realization

Keeps the docs accurate by implementing a small script to auto-update the docs-as-code version.

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

AlexanderLanin:
We cannot simply update the version via script, as such an increment (re-)triggers tool verification process.

nicu1989:
> We cannot simply update the version via script, as such an increment (re-)triggers tool verification process.

Hmm, good point, then shall we create a CI workflow or step in existing workflow that verifies the version from `bazel_dep(name = "score_docs_as_code", version = "1.2.0")` against the version from `docs/score_tools/doc_as_code.rst` ? This way we remind the docs-as-code uplifter to update the version in .rst as well?

PR #1802, opened 2025-09-25
ci: Check Docs-as-Code version consistency

nicu1989:
Verify that the version defined in MODULE.bazel
matches the one in doc-as-code.rst and warn in PRs.

Related: https://github.com/eclipse-score/score/issues/1707

### DIFF
--- a/.github/scripts/check_doc_tool_version.py
+++ b/.github/scripts/check_doc_tool_version.py
@@ -1,0 +1,79 @@
+# *******************************************************************************
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
+#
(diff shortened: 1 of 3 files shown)


nicu1989:
A warn comment is added in all PR indicating a version mismatch in docs. No comment is added when docs-as-code is the same in MODULE.bazel and the doc-as-code.rst.

image

AlexanderLanin:
the problem is that the score MODULE file does not decide which docs-as-code version is used. That's decided dynamically in integration repo. But it's a start :)

AlexanderLanin (approved):

Issue #1461, opened 2025-07-21
Improvement: Reduce the size of gh-pages


nicu1989:
Our current(21.07.2025) gh-pages branch is close to github limit of 1GB as stated [here](https://docs.github.com/en/pages/getting-started-with-github-pages/github-pages-limits): 

_Published GitHub Pages sites may be no larger than 1 GB._

While we have stale PR clean-up we need to make sure that PRs wont push `gh-pages` over this limit. Also lets consider the other limitations from the link above.

Current snapshot:
```
score$ du -sh .
860M    .
```
Tree:

```
├── index.html
├── main
├── pr-1069
├── pr-1115
├── pr-1125
├── pr-1170
├── pr-1202
├── pr-1203
├── pr-1209
├── pr-1246
├── pr-1258
├── pr-1281
├── pr-1293
├── pr-1371
├── pr-1374
├── pr-1375
├── pr-1388
├── pr-1395
├── pr-1408
├── pr-1416
├── pr-1435
├── pr-1437
├── pr-1438
├── pr-1451
├── pr-1457
├── pr-1458
├── pr-822
├── pr-890
└── versions.json
```

a-zw:
I once optimized a sphinx theme to not write a duplicate of the navigation to each and every html file. That reduced like 90% of the html storage.

I doubt this is the problem here though. The navigation on the score website seems rather small.

PR #1834, opened 2025-10-02
ci: Warn on gh-pages branch size

nicu1989:
Display a warning message in PR that docs
might not be displayed properly if gh-pages size > 950MB.

Related: https://github.com/eclipse-score/score/issues/1461

### DIFF
--- a/.github/actions/deploy-versioned-pages/action.yml
+++ b/.github/actions/deploy-versioned-pages/action.yml
@@ -108,6 +108,22 @@ runs:
         folder: version_root
         clean: false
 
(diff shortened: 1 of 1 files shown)


nicu1989:
Tested with 500MB to trigger the warning: 
image
In case the branch size is smaller than the limit no warning/text is inserted in the docs comment.

AlexanderLanin (approved):

AlexanderLanin:
@nicu1989 this repo is not using cicd-workflows... we should fix that anyway... then your warning will disappear again :(

PR #1845, opened 2025-10-06
docs: add DR-003-infra for devcontainer strategy

olivembo:
## Summary

This PR adds Design Decision Record DR-003-infra that outlines the strategy for standardizing S-CORE development environments using devcontainers.

## Changes

- Added `docs/design_decisions/DR-003-infra.md` with comprehensive analysis of devcontainer approaches
- Proposes hybrid approach with base container + service-specific extensions
- Addresses both local development experience and CI/CD performance needs

## Key Decisions

- **Base devcontainer**: Shared tooling foundation for consistency
- **Service extensions**: Rich local development environments  
- **CI/CD optimization**: Lean containers for performance
- **Layered approach**: Balance between simplicity and flexibility

## Benefits

- Consistent development environment across contributors
- Improved onboarding experience
- Optimized CI/CD performance
- Flexible local development workflows
- Clear separation of concerns

This addresses the infrastructure standardization efforts and provides a clear path forward for S-CORE development environment consistency.

## Testing

- [x] Document follows existing DR format and structure
- [x] All requirements and options clearly documented
- [x] Decision rationale provided with clear consequences
- [x] Next steps defined for implementation

### DIFF
--- a/docs/design_decisions/DR-003-infra.md
+++ b/docs/design_decisions/DR-003-infra.md
@@ -29,7 +29,7 @@ Additionally, we must ensure that devcontainers are usable both **locally** and 
 
 ---
 
-## 2. Requirements
+## 2. Goals and Requirements
 
 1. **Performance**: Acceptable build/startup times and reasonable resource usage, both locally and in CI/CD.
 1a. **Startup Performance**: Fast container initialization for productive development workflows.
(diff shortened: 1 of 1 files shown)


nick-hildebrant-etas:
It's not really in the scope of this document, but as a door for layer local side effects, maybe it should be mentioned that additional layers should not be required for CI/CD. The CI layer is wholly responsible for builds and required tooling, which should be brought in via Bazel whenever possible?

FScholPer:
@olivembo Is that something you want to drive forward?

olivembo:
> I agree with the general "Hybrid Approach" idea. However, I would suggest to use a base layer (built in the [devcontainer](https://github.com/eclipse-score/devcontainer) repository), which contains all tools that also CI requires.
> 
> If there are then tools which are required only for development of certain repositories, and which really add a significant amount of data (!! I would not do this for just a few MB, since it adds overhead and complexity !!), we can create (or re-use existing) devcontainer features.
> 
> An example can be seen [in the devcontainer repository itself](https://github.com/eclipse-score/devcontainer/blob/main/.devcontainer/devcontainer.json). The `devcontainer.json` there constructs the local devcontainer (used for developing the s-core devcontainer) from "default" building blocks:
> 
>     * `javascript-node:0-18` acts as base image
> 
>     * `ghcr.io/devcontainers/features/git` etc. are additional "layers"
> 
>     * These additional layers are then on-the-fly added on top of the base image.
> 
> 
> However, we must keep in mind:
> 
>     * Everything that is downloaded from "the internet" may stop existing _at any point in time_, killing our development environment and builds. We already had this, with the downtime of the Arm GitLab.
> 
>     * Everything that is downloaded from "the internet" may be used for injecting "unwanted things" into the supply chain. This applies to dynamic features, but also for Bazel, its modules and additional files, Visual Studio Code plugins, etc. It is _much less risky_ to use/audit/archive pre-built images, than an on-the-fly created development & build environment. Food for thought...

I just adapted the DR-003 document. Agree to the points.

The "downloaded from the internet" problem, you have imho in any case. Maybe in the dynamic features more than with the usage of prebuild images, but that should be treated independent of this DR.

a-zw (approved):

olivembo:
> @olivembo Is that something you want to drive forward?

We would drive it from ETAS side

PR #1886, opened 2025-10-09
upgrade process description and doc-as-code

masc2023:
process_description v1.2.1
doc-as-code v1.4.0


masc2023:
> as I understand the warning from the CI, also https://github.com/eclipse-score/score/blob/main/docs/score_tools/doc_as_code.rst should be updated.

Warning was before force push, there was unresolved links, which are not there any more. please check again and please approve

PR #1913, opened 2025-10-17
Orchestration architecture update

prabakaranklst:
Updated files to fix review comments, CI failure and appearance.


prabakaranklst:
This is will be addressed through another PR https://github.com/eclipse-score/score/pull/1914

PR #1927, opened 2025-10-22
Implement checks

invincibleXiaobao:


FScholPer:
trlc only allowed in module repos

PR #1930, opened 2025-10-22
Decision Record 004

a-zw:
Rendered HTML: https://eclipse-score.github.io/score/pr-1930/design_decisions/DR-004-infra.html

### DIFF
--- a/docs/design_decisions/DR-004-infra.rst
+++ b/docs/design_decisions/DR-004-infra.rst
@@ -41,17 +41,17 @@ Goals and Requirements
 ^^^^^^^^^^^^^^^^^^^^^^
 
 1. We want to enable each module to work independently and efficiently, i.e. build their own documentation quickly.
-2. We want a complete and consistent documention for a release, i.e. everything with backlinks.
+2. We want a complete and consistent documentation for a release, i.e. everything with backlinks.
    Regulations only require that *eventually* though and during development slight inconsistencies are acceptable.
 3. Some modules want backlinks even if it slows down the doc build.
    For example, Platform wants backlinks for the feature requirements.
-   For contrast, Baselibs might not care about backlinks because they simply assume that everybody uses them.
+   For contrast, Baselibs may not require backlinks because they simply assume that everybody uses them.
 
 Non-Goals
 ~~~~~~~~~
 
 * `#258 <https://github.com/eclipse-score/docs-as-code/issues/258>`_: The repository layout for documentation is not relevant here.
-* `#259 <https://github.com/eclipse-score/docs-as-code/issues/259>`_: Whether only certain rst files should contain needs element is not relevant here.
+* `#259 <https://github.com/eclipse-score/docs-as-code/issues/259>`_: Whether only certain rst files should contain needs elements is not relevant here.
 * `#260 <https://github.com/eclipse-score/docs-as-code/issues/260>`_: Whether Bazel should handle the creation of some documentation artifacts is not relevant here.
 * `#261 <https://github.com/eclipse-score/docs-as-code/issues/261>`_: Whether Bazel should configure documentation variants is not relevant here.
 * Referencing Sphinx elements apart from Needs is not relevant here because such links are never bi-directional.
@@ -86,6 +86,8 @@ A prototype is available in `a-zw/score_full-docs <https://github.com/a-zw/score
 Current implementation for external needs means modules reference external ids with a prefix.
 This prefixed id does not exist in a mega-build which results in Sphinx warnings.
 
+💚  Effort: Proof-of-concepts exist, so the basic implementation should be quick.
+
 😡  Bureaucracy: Modules must not use prefixes for external needs because it breaks the integration build.
 The docs-as-code repo must be refactored accordingly.
 Since the id schemas already contain namespacing rules, there is barely any risk of id clashes.
@@ -113,6 +115,8 @@ While it might seem redundant if we already have the independent module's docs, 
 Since modules might reference older versions of other modules, integration may be forced to override versions.
 This is an additional reason for a rebuild.
 
+😡  Effort: We have no proof-of-concept yet. Especially one which provides a live-preview with multiple modules.
+
 💚  Bureaucracy: Modules decide about prefixes for themselves.
 
 💚  Speed: This should build faster because we can rebuild modules in parallel.
@@ -126,8 +130,9 @@ Evaluation
    :header: Criteria, Option S, Option M
    :widths: 20, 10, 10
 
+   Effort, 💚, 😡
    UX, 💚, 😡
    Bureaucracy, 😡, 💚
    Speed, 😡, 💚
 
-Since UX is more important than speed and bureaucracy, option S is best.
+There is no particularly strong criterium, thus we pick option S because it can be done quickly.


docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,112 @@
...
+Module vs Target dependencies
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To understand how backlinks do not imply circular dependencies we must understand the difference between module and target dependencies.
+
+*If* Baselibs wants to see backlinks (i.e. COM references) in their documentation, they need to introduce the blue dependencies below:
+
+.. image:: img/DR-004-backlinking.png
+
+There is a additional module dependency to the Integration module.
```
dcalavrezo-qorix: an additional

dcalavrezo-qorix:
also, not a big fan of emojis in documentation ( but that's just my taste)

a-zw:
> also, not a big fan of emojis in documentation ( but that's just my taste)

I essentially want something clearly saying "good" or "bad" and red vs green does that. In Confluence there is an easy way to make it "green plus" or "red minus":

image

Unfortunately, there are no corresponding emojis afaik. Hence the smiley and heart thingies.

a-zw:
> How do you handle different versions for Option S ?

Version as in "0.5" or "1.0"? That would simply be separate builds. Nothing changes from the status quo.

Do you see a need to do something @dcalavrezo-qorix ?

ubmarco (commented):
I want to discuss this!
There are concepts out there for bidirectional links.
https://github.com/useblocks/sphinx-needs/discussions/1166
https://github.com/useblocks/sphinx-needs/discussions/1220
It's not required to put all documentation into a single project.

This comes down to a multi stage build in which stage 1 builds needs.json and ignores broken links to generate the project inventory.
Then needs.json files are exchanged for a full stage 2 build of the projects.
Stage 1 is commonly fast as HTML is not written.
This leads to a non-cyclic Bazel build graph.
It's some effort however to model this in Bazel.

Is bidirectional linking a hard requirement or just user convenience? Because not requiring this makes our lives much easier.
If you wish we can chat about this tomorrow.

ubmarco:
So these are independent topics:

1. Bi-Directional linkage of needs
2. Bi-Directional linkage of Sphinx references (intersphinx)
3. Split traceability from docs generation

This DR is about 1.

For 3. there are other DR issues raised by @AlexanderLanin that we should follow up.
https://github.com/eclipse-score/docs-as-code/issues/258
https://github.com/eclipse-score/docs-as-code/issues/259
https://github.com/eclipse-score/docs-as-code/issues/260
https://github.com/eclipse-score/docs-as-code/issues/261

The bi-directional linking actually depends on these DR, so starting there is a reasonable choice for me.

a-zw:
> I want to discuss this! There are concepts out there for bidirectional links. [useblocks/sphinx-needs#1166](https://github.com/useblocks/sphinx-needs/discussions/1166) [useblocks/sphinx-needs#1220](https://github.com/useblocks/sphinx-needs/discussions/1220) It's not required to put all documentation into a single project.

The concepts you reference should be Option M in the decision record. I can add these references to the DR.
 
> Is bidirectional linking a hard requirement or just user convenience? Because not requiring this makes our lives much easier. If you wish we can chat about this tomorrow.

Yes, that is precisely the essential question. What you describe as "makes our lives much easier" is probably the "bureaucracy" criterium in the DR? The tradeoff is the UX for people who just want nice documentation.

In other words, depending on _whose_ lives we want to make easier, we must pick either option S or M.

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,117 @@
...
+.. image:: img/DR-004-example.png
+
+
+Goals and Requirements
+^^^^^^^^^^^^^^^^^^^^^^
+
+1. We want to enable each module to work independently and efficiently, i.e. build their own documentation quickly.
+2. We want a complete and consistent documention for a release, i.e. everything with backlinks.
+3. Some modules want backlinks even if the slows down the doc build.
+   For example, platform wants backlinks for the feature requirements.
```
AlexanderLanin: Feature requirements are in the platform? Assuming you mean score repo

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,117 @@
...
+   Baselibs does not care about backlinks because they assume everybody anyways.
```
AlexanderLanin: Check Sentence

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+For discussion, this subset of S-CORE modules shall serve as an example:
+
+.. image:: img/DR-004-example.png
+
+
+Goals and Requirements
+^^^^^^^^^^^^^^^^^^^^^^
+
+1. We want to enable each module to work independently and efficiently, i.e. build their own documentation quickly.
+   (Sphinx builds are annoyingly slow in any case but that is not the core point of this decision record.)
```
AlexanderLanin:
```suggestion
```

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+2. We want a complete and consistent documention for a release, i.e. everything with backlinks.
+   Regulations only require that *eventually* though and during development slight inconsistencies are acceptable.
```
AlexanderLanin, line 45: Inconsistencies will lead to build failures during development. I would address this more to convinience: we do not have a hosted html version of a every specific dependency/version.

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+3. Some modules want backlinks even if the slows down the doc build.
+   For example, Platform wants backlinks for the feature requirements.
```
AlexanderLanin, line 47: platform=score, and feature requirements are in score. Backlinks to what?

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+Module vs Target dependencies
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To understand how backlinks do not imply circular dependencies we must understand the difference between module and target dependencies.
+
+*If* Baselibs wants to see backlinks (i.e. COM references) in their documentation, they need to introduce the blue dependencies below:
+
+.. image:: img/DR-004-backlinking.png
+
+There is an additional module dependency to the Integration module.
```
AlexanderLanin, line 68: I don't understand this paragraph sorry 😆 

I guess what you want to say, is that baselibs in general is not aware of its users and therefore cannot show backlinks.

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Collect the documentation from all repositories and build a single huge html website.
+
+A prototype is available in `a-zw/score_full-docs <https://github.com/a-zw/score_full-docs/tree/61205c616a99e395e0f63fb995bd828271433b44>`_.
+
+Current implementation for external needs means modules reference external ids with a prefix.
+This prefixed id does not exist in a mega-build which results in Sphinx warnings.
+
+😡  Bureaucracy: Modules must not use prefixes for external needs because it breaks the integration build.
```
AlexanderLanin, line 91: Modules do use such prefixes. So the overhead is, that we need to remove them on the fly when building such a full build. Or we need to prefix the needs on the fly to avoid name collisions.

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+The docs-as-code repo must be refactored accordingly.
+Since the id schemas already contain namespacing rules, there is barely any risk of id clashes.
+
+😡  Speed: This will result a long-running build step unsuitable for pull requests in general.
+This only affects the final integration build.
+Everybody else can independently chose their sweet spot between link consistency and build speed.
```
AlexanderLanin:
```suggestion
Everybody else can independently chose their sweet spot between bi-directional link convenience and build speed.
```

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+for linking between different documentations.
+By using them on both sides, bi-directional traceability can be achieved without circular dependencies.
+
+The documentation of a single module may be outdated with respect to external needs, i.e. backlinks.
+Just imagine a new module is created which uses Baselibs but Baselibs makes no changes, so the documentation is never rebuilt.
+
+This implies that the integration module must build all documentation themselves for a release.
+While it might seem redundant if we already have the independent module's docs, this is necessary to ensure consistency.
+
+Since modules might reference older versions of other modules, integration may be forced to override versions.
```
AlexanderLanin, line 115: integration will need to host the html pages of modules itself.

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+This just makes all their targets visible to each other.
+
+There is an additional target dependency from @Baselibs:html to @Integration:needs_json which contains *all* needs-elements of the S-CORE project.
+There is no circular dependency among target dependencies.
+Bazel would show an error if you try.
+
+Options Considered
+------------------
+
+Option S: Single documentation build
```
AlexanderLanin, line 79: Option S will work better when multiple repositories are local with local_module_override?!

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+
+.. csv-table::
+   :header: Criteria, Option S, Option M
+   :widths: 20, 10, 10
+
+   UX, 💚, 😡
+   Bureaucracy, 😡, 💚
+   Speed, 😡, 💚
+
+Since UX is more important than speed and bureaucracy, option S is best.
```
aschemmel-tech: Discussion result from the processes community - see https://github.com/orgs/eclipse-score/discussions/407#discussioncomment-14760276 : we would be in favour of option M, as the "UX" advantage between S and M we consider as small.

docs/design_decisions/DR-004-infra.rst
a-zw: Component requirements should be in the modules but currently they are in platform (=score). They derive from feature requirements which should be in platform, so requirements in modules reference requirements in platform.

One probably wants to know which modules are involved for certain features, thus you want to see the backlinks in the feature requirements in the platform documentation.

docs/design_decisions/DR-004-infra.rst
a-zw: The paragraph only describes the image. In particular, the dependencies (fat dashed lines) from Baselibs to Integration and back.

(Why doesn't Github allow me to insert an image here? 😡)

docs/design_decisions/DR-004-infra.rst
a-zw: We can just remove the prefixes. That should work without conflicts because id schemas already involve namespaces via process. It just means a little bit of refactoring work... s/PROCESS_//g

docs/design_decisions/DR-004-infra.rst
a-zw: There can be inconsistencies without build failures.

For example, let's assume Platform has backlinks to component requirements in modules, such as Com. Now Com adds a new component requirement derived from an existing feature requirement in Platform. Platform documentation does not _yet_ show the backlink for this simply because it has not been rebuilt yet. Thus the github.io documentation of Platform and Com are inconsistent.

docs/design_decisions/DR-004-infra.rst
a-zw: I don't understand. What do you mean by "local_module_override"? Also, "better" in what way?

docs/design_decisions/DR-004-infra.rst
```diff
@@ -0,0 +1,134 @@
...
+Option M: Module documentation rebuilt
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Sphinx provides `intersphinx <https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html>`_
+and Sphinx-Needs provides `needs-external-needs <https://sphinx-needs.readthedocs.io/en/stable/configuration.html#needs-external-needs>`_
+for linking between different documentations.
+By using them on both sides, bi-directional traceability can be achieved without circular dependencies.
+
+The documentation of a single module may be outdated with respect to external needs, i.e. backlinks.
+Just imagine a new module is created which uses Baselibs but Baselibs makes no changes, so the documentation is never rebuilt.
```
ubmarco, line 110: How can this fail? If there is a dependency to baselibs, then this is modeled in Bazel and the needs.json / objects.inv artifacts are provided by the build system to the new module.

a-zw: It doesn't fail. The new module is fine. 

The inconsistency is that the baselibs documentation hosted on github.io does not show any backlinks. It is "outdated". This is until someone triggers a rebuild. (All of this assumes baselibs is configured to consider all modules for backlinks)

ubmarco: Ok got it, this can only be solved by having the 2-stage build I was talking about earlier.
1. Generate needs.json / intersphinx inventories for all dependent projects while ignoring broken links.
2. Exchange the files between projects
3. Run the full builds of all participating projects
4. Deploy the docs

Depending on the amount of project relations, that leads to more frequent full builds of all projects.

Btw, the stage 1 build is also related to [namespace support](https://github.com/useblocks/sphinx-needs/discussions/1088) and [edge properties](https://github.com/useblocks/sphinx-needs/discussions/1481).
If Sphinx-Needs knows whether a link goes internal or external, a safe stage 1 build becomes possible.
We would not have to suppress dead link warnings.

Another thought: How would the integration repository know in which order to build the projects for case M? You would only get an acyclic Bazel build graph when introducing 2-stage builds.

AlexanderLanin:
> Can we please divide into traceability tests and doc generation. I still would see that before we dive deeper.

@FScholPer note that you can use different commands for exactly those two use cases... which we did not document.... hups.

Commands:
* `bazel run //:docs` to generate docs.
* `bazel run //:docs_check` to quickly check traceabily and all our custom rules.

docs/design_decisions/DR-004-infra.rst
AlexanderLanin: Platform will never have backlinks to com :)
Platform would be generic, and is not aware of who exactly implements those requirements.
That information is always only available at integration time.

That's why I have 4 platform websites in this drawing: https://github.com/AlexanderLanin/score/blob/docs-multirepo/docs/_tooling/decision-records/_assets/multirepo_bidirectional.drawio.svg

docs/design_decisions/DR-004-infra.rst
AlexanderLanin: Wow that's hard to read 😆 
I guess this is answered by the previous comment/image?

docs/design_decisions/DR-004-infra.rst
AlexanderLanin: > Why doesn't Github allow me to insert an image here? 😡

Corporate bureaucracy. If you were allowed to upload an image, you could leak internal data.
(Creating public commits is a different process/responsible)

docs/design_decisions/DR-004-infra.rst
AlexanderLanin: I thought we have an issue for that, but I cannot find it. So I'll try to explain the use case here:

When I have two repos cloned. Both with docs-as-code, and one depends on the other.
e.g. docs-as-code repo and process repo.

When I need to edit them in sync, I would create a local_module_override in docs-as-code (or use some fancy bazel command line flag). However, now when I build the html pages, or use the live_preview feature. The generated pages do not link to my local process html pages.

As far as I understand that would be implicitly solved by option S, since I get a single big website. While option M is quite difficult, as it needs to know the other url. Which is probably on localhost, but with a random port number.

docs/design_decisions/DR-004-infra.rst
AlexanderLanin: Can't integration just build stage 1 on all documentations, exchange json files, build stage 2 on all documentations?

By the way, we can easily differentiate external-need-links as they start with an uppercase letter (module prefix).

AlexanderLanin:
Option S is the old option 1.1?
Option M is the old option 1.2?

Those old options mention some pros and cons that are missing in this DR?
And the generic use cases and constrains should be taken over?
And keep `Multiple Platform Websites Model` in mind!

I guess that old DR is actually valid. As it states we'll implement option 2 (done), and then evaluate switching to option 1 (now) 😆

docs/design_decisions/DR-004-infra.rst
ubmarco: Yes, that's possible. Somes notes:
- You can also have links between other documentations, not just between integration and its children
- What's the relation between integration and docs? Is that the repo we are talking about https://github.com/eclipse-score/reference_integration? I do not see docs in there. If integration just means, we are rebuilding and deploying all documentations, to be sure the backlinks are up-to-date, then it's clear. But I would assume a safe build system, which, on change, knows what to rebuild. Backlink dependencies are part of this. If a needs.json changes, then all dependent projects on this file have to be rebuild.
- We need to make sure the stage 1 needs.json outputs are reproducible.
- Sphinx-Needs does not know about the module prefixes after the import, so you cannot suppress warnings based on this. That means we cannot rely on the `-W` flag (which turns warnings into != 0 exit code).

a-zw: I only learned about the [bazel-drives-sphinx](https://github.com/useblocks/bazel-drives-sphinx) proof-of-concept yesterday. It seems to overlap very well with this decision record. I see no conflicts and the diagram in the readme is essentially the same diagram as in this record (just set project 1 as baselibs, project 2 as com, and project 3 as integration).

Correct, reference_integration has no documentation at all currently. [Here is a commit to add it](https://github.com/etas-contrib/score_reference_integration/commit/80aadc17b9e73537014ce167bdc300dc05dffe5e) but it assumes we decide for option S here.

Now, I don't see any disagreement in this thread, so I believe we can resolve it.

docs/design_decisions/DR-004-infra.rst
ubmarco: I do not like this catch-all statement in the context of this DR. Sphinx is a highly extensible documentation generator written in Python. The performance significantly depends on the usage and extensions people add. If you start rendering plantumls, no documentation generator can make this magically faster. I think it's good enough for the use cases we have.
I posted a message in Slack the other day that I was writing a needs.json for 15k needs in 8s.
On my poor Linux laptop. Quick traceability becomes possible then. Documentation generation takes longer, sure, but also more work to do there translating docutils nodes to HTML.

docs/design_decisions/DR-004-infra.rst
a-zw: First, may you never have the need to "edit them in sync" because you already got into a mess then. 🤞

Yes, technically an advantage for S but I don't think it is a relevant one because this sync-editing will be painful in either case.

docs/design_decisions/DR-004-infra.rst
a-zw: Wait, now you confused me.

"Platform" is the eclipse-score/score repository for me. It does not reference the modules. However, linking happens the other way round because modules reference the higher-level requirements in Platform. This implies **back**links in Platform if we want bi-directional traceability.

If these backlinks are visible on the eclipse-score.github.io/score/ website or just on eclipse-score.github.io/reference_integration/ is intentionally left open by this decision record. The Platform repository can be configured either way.

docs/design_decisions/DR-004-infra.rst
a-zw: Since you are from Useblocks, I understand your dislike. You are right that words like "annoyingly" are rather emotional.

However, speed seems to be the primary concern for everybody. Whenever I discuss Sphinx concepts like this DR with developers, speed concerns come up. We use Sphinx a lot but usually _despite_ its speed.

a-zw:
> Option S is the old option 1.1? Option M is the old option 1.2?

Yes, probably.
 
> Those old options mention some pros and cons that are missing in this DR?

The "Shared navigation menu and search functionality" is "UX" here.

The "No customization via conf.py, limiting module-specific templates" I'm not sure. Some customization is certainly possible. Some things are not but our process and docs-as-code toolchain already introduce limitations too and I'm not sure if that would limit it further.

>And the generic use cases and constrains should be taken over? 

The use cases are implicitly mentioned here.

I don't think the constraints are necessary for the decision making here because they are too fundamental. Conciseness is important.

> And keep `Multiple Platform Websites Model` in mind!

Neither option S nor M prevents anything about this.

docs/design_decisions/DR-004-infra.rst
ubmarco: I would say these are 2 different things. This DR states
> The S-CORE project adopted a multi-repo strategy. Since [S-CORE aims for compliance with regulations like ASPICE](https://eclipse-score.github.io/process_description/main/standards/index.html#standards), bi-directional traceability is necessary (see gd_req__req_traceability) and should be made convenient. How to provide this in S-CORE documentation?

The wanted bi-directional traceability in this context means `backlinks`. So the dependency is still unidirectional, but we want the HTML output to be aware of the opposite direction of links. I do not really like the term `bi-directional traceability` as it implies the dependency direction is bidirectional. It's not. I think `backlinks` is clearer. From build system point of view, if we have a 2-stage build, the bidirectionality is resolved which leads to a proper build DAG.

The [bazel-drives-sphinx](https://github.com/useblocks/bazel-drives-sphinx) approach on the other end cares for file based traceability with Sphinx/Sphinx-Needs and a better integration to build systems (dependency management, caching). The goal is quick traceability that integrates well with possibly slower documentation builds. The approach can be adopted with or without backlinks.

The main question is whether we are accepting that some deployed versions of the build has incomplete backlinks, then we can run a softer version and not regenerate docs constantly. It also means any schema validation / traceability statement that must exist *fast*, cannot rely on backlinks being available.

a-zw: The term "bi-directional traceability" is used by ASPICE and an important keyphrase for many people in automotive.

This decision simply assumes that we accept "that some deployed versions of the build has incomplete backlinks". It says so in the goals section: "during development slight inconsistencies are acceptable." So far, nobody has questioned this assumption. Most importantly, our colleagues from the process community did not question it.

AlexanderLanin:
Pro S: simpler to implement (IMHO)
Pro S: local multi-repo changes are naively supported via live_preview

-> Would it make sense to start with S and investigate M once we hit performance problems? Or offer both possibilities?

AlexanderLanin:
We did not cover images correctly in CODEWNER. Can someone approve this please?

PR #1996, opened 2025-11-13
Switch to tagged devcontainer

opajonk:
WIth this change, a semver-tagged devcontainer version is pinned. Thus, updates to the "latest" tag cannot unexpectedly cause harm to the repository anymore.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,9 +1,4 @@
 {
     "name": "eclipse-s-core",
-    "image": "ghcr.io/eclipse-score/devcontainer:latest",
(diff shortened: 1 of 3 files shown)


opajonk:
I checked the following to test the container:

```
bazel test //...
bazel run //:docs
bazel run //:ide_support
```

Those passed fine.

In the README, a target `bazel coverage --combined_report=lcov //docs:score_metamodel_test` is mentioned. However, this target does not exist (anymore?). The target `bazel run //:format.check` does not work since it expands to three targets. Running those individually worked.

AlexanderLanin (approved):

Issue #2020, opened 2025-11-14
Certified libcore


pawelrutkaq:

dcalavrezo-qorix:
Hi @pawelrutkaq,

Is this issue still relevant? If not, please add the `trash` label so it can be marked for deletion.

attifunel:
Sentence "`libcore` certified as ASIL-B in quite big scope already" at https://eclipse-score.github.io/score/main/design_decisions/DR-001-arch.html unclear for me. Does it mean that the library is **NOT** fully certified? If so, is there a guide for its non-certified parts? What about libstd?

PR #2249, opened 2025-11-26
rust: add libcore to certification doc

pawelrutkaq:
Fulfills: https://github.com/eclipse-score/score/issues/2020

WIP: Still looking into empty internal modules at the certification part.

### DIFF
--- a/docs/contribute/development/rust/certification/toolchain/libcore.rst
+++ b/docs/contribute/development/rust/certification/toolchain/libcore.rst
@@ -14,3 +14,120 @@
 
 Certified libcore
 #################
+
+.. tool_req:: Rust libcore
+   :id: tool_req__rust_libcore
+   :safety: ASIL_B
+   :security: NO
+   :status: valid
+
+   Rust `libcore` shall be certified at `ASIL-B` or higher for below list of modules:
+
+   .. list-table::
+      :header-rows: 1
+      :widths: 15 10 60
+
+      * - Module
+        - Certification Available (Ferrocene)
+        - Comment
+      * - alloc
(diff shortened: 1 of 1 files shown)


pawelrutkaq:
before this goes out of draft, I will try to use feroceen compiler to recheck if there are more gaps hidden inside libcore specific modules. Currently our codebase support cargo, so it may be that I will use cargo for that.

docs/contribute/development/rust/certification/toolchain/libcore.rst
```diff
@@ -14,3 +14,120 @@
 
 Certified libcore
 #################
+
+.. tool_req:: Rust libcore
```
masc2023, line 18: Thanks for starting here. Can you add also something here, https://eclipse-score.github.io/score/main/score_tools/index.html, which is more or less the evaluation result to motivate, that we need the qualification here for the different elements

pawelrutkaq: I would do it in seperate PR once I obtain more info from Feroceen about the compiler itself. The libcore is just the library part. I will also propose to tak eover Feroceen aproach for TCL and all tooling assesment for other tools we need in rust scope

masc2023: Fine for me, thank you

FScholPer:
@pawelrutkaq can you please update

pawelrutkaq:
Yes, soon i will do.

FScholPer:
@pawelrutkaq close or ready?

PR #2262, opened 2025-11-28
Revise DR-005-infra design document

nradakovic:
Updated the design document to reflect the proposed directory structure and rationale for the S-CORE Bazel C++ Toolchain Configuration Repository.

### DIFF
--- a/docs/design_decisions/DR-005-infra.rst
+++ b/docs/design_decisions/DR-005-infra.rst
@@ -1,0 +1,242 @@
+DR-005-infra: Directory Structure for S-CORE Bazel C++ Toolchain Configuration Repository
+====================================================================================
+
+Overview
+--------
+
+This repository acts as a **pure configuration layer** for C++ toolchains used in Bazel builds.
+It contains **only configuration files** (``BUILD.bazel``, ``.bzl``), while all **toolchain
+binaries** are downloaded from external repositories via ``http_archive``.
+
+Platform definitions are **not** part of this repository—they are provided separately by the
+S-CORE Bazel Platforms module.
+
+This structure supports:
+
+- Multiple compiler versions
+- Multiple OS/CPU combinations
+- Clean layering following Bazel best practices
+- Bzlmod-compatible module extensions
+- Reusability and long-term maintainability
+
+----
+
+High-Level Goals
+----------------
+
+- Provide a **central place** for all C++ toolchain configuration logic.
+- Separate toolchain **binaries** from **configurations**.
+- Provide scalable support for:
+
+  - Linux (x86_64, arm64)
+  - QNX (x86_64, arm64)
+  - Multiple compiler variants (GCC, Clang, QCC)
+  - Multiple versions per compiler
+
+- Allow all consuming repositories to pull toolchains via **Bzlmod** with simple configuration.
+
+----
+
+Proposed Directory Structure
+----------------------------
+
+.. code-block:: text
+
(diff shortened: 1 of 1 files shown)


docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,242 @@
+DR-005-infra: Directory Structure for S-CORE Bazel C++ Toolchain Configuration Repository
+====================================================================================
```
AlexanderLanin, line 2:
```suggestion
=================================================================================================
```

docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,242 @@
...
+
+High-Level Goals
+----------------
+
+- Provide a **central place** for all C++ toolchain configuration logic.
+- Separate toolchain **binaries** from **configurations**.
+- Provide scalable support for:
+
+  - Linux (x86_64, arm64)
+  - QNX (x86_64, arm64)
```
evinoth1206, line 32: Given that Linux has multiple distributions and S-core currently uses three different ones (Ubuntu for dev containers, EB-Linux and Red Hat Linux for reference integration), should we consider creating separate platform and tool chain definitions for each (e.g., linux-ubuntu, linux-eb, linux-redhat) or as the toolchain contains only rules/config it should be compatible for different Linux flavours?

docs/design_decisions/DR-005-infra.rst
LittleHuba, line 1: @nradakovic am I right that this would go into the direction of implementing all the toolchains on our own? What about available opensource toolchains for GCC and LLVM?
Or would it be viable to use these available toolchains as part of this setup?

I would like an evaluation of the different options. This helps in the future if we ever want to reevaluate or understand why we decided this way.

docs/design_decisions/DR-005-infra.rst
nradakovic: These are distributions and as such should not depend on platform constraints (OS and CPU). We can add flags to platform if we want to have some separation between them, but I would like first to know what is delta between these 3 in terms of configuration.

docs/design_decisions/DR-005-infra.rst
nradakovic: GCC doesn't exist in OS world (at least I'm not aware) and LLVM seems to be a bit heavy on resource usage (not optimized for CI). That's why BMW has it's own implementation of LLVM.

LittleHuba: It would still be great to have this as part of the document.

> GCC doesn't exist in OS world (at least I'm not aware)

GCC does exist here: https://github.com/f0rmiga/gcc-toolchain
It works as advertised, but I can't judge the quality of the implementation.

> LLVM seems to be a bit heavy on resource usage (not optimized for CI)

My proposal is to then upstream the necessary changes. I'm sure that other people also use the toolchain in a CI environment. So there should be interest in having this in the OS world.

nradakovic: Well, it's not my call when LLVM will be outsourced from SWH end, I cannot enforce this. For now we use the one that is available. Once it's available we can add it into repo without too much effort.
The other GCC solution doesn't really fit our needs. The interface is quite close, we cannot inject additional features (my implementation will support that) and there is no way of redefining existing features. Also I cannot point to different package. This config only support certain versions of GCC.

PR #2290, opened 2025-12-04
update safety manager roles

masc2023:

PR #2299, opened 2025-12-05
bugfix for

masc2023:
https://github.com/eclipse-score/score/pull/2290


masc2023:
Seems now my PR was properly build, so can close that here

PR #2351, opened 2025-12-16
Update CODEOWNERS for PMP part

pahmann:
@PandaeDo is safety manager and quality manager, but misses the right to submit approving of the PMP, which should be changed by an update of the CODEOWNER file.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -33,7 +33,7 @@ verification/                         @eclipse-score/automotive-score-committers
 /docs/manuals/                        @eclipse-score/automotive-score-committers
 /docs/modules/                        @eclipse-score/automotive-score-committers
 # /docs/platform_management_plan/     @eclipse-score/community-process
-/docs/platform_management_plan/       @pahmann @PhilipPartsch @masc2023 @aschemmel-tech
+/docs/platform_management_plan/       @pahmann @PhilipPartsch @masc2023 @aschemmel-tech @PandaeDo
 # /docs/quality/                      @eclipse-score/quality-managers
 /docs/quality/                        @masc2023 @pahmann @PandaeDo
 # /docs/requirements/stakeholder/     @eclipse-score/automotive-score-technical-leads


masc2023:
> @PandaeDo is safety manager and quality manager, but misses the right to submit approving of the PMP, which should be changed by an update of the CODEOWNER file.

What is here the reference, only the selected roles in meritocratic way counts, as done here?
https://eclipse-score.github.io/score/main/platform_management_plan/role_assignment/platform_safety_manager.html

If this is agreed, then first other roles should be elected, and properly updated.

.github/CODEOWNERS
masc2023, line 1: Needs more updates, Safety Manager are responsible also for other work products, e.g. Safety Managent Plan, see here https://github.com/eclipse-score/score/pull/2352, PanDoe not added as reviewer

anmittag: @pahmann general question: why do we use user-names here and not work with [Teams](https://github.com/orgs/eclipse-score/teams) and is it an "AND" or and "OR" relation?

masc2023: Teams does not work, it is "OR"

masc2023:
> PLs or TLs should also be in that list

No, as defined here, https://eclipse-score.github.io/process_description/main/general_concepts/score_review_concept.html
PLs are responsible, but approval by Process Community, which is reflected in the CODEOWNER File

TLs as discussed Role does not exists any more, at least in Process Description

pahmann:
I see that this will become a lengthy discussion potentially. I spotted that @PandaeDo was not listed as reviewer when I opened PMP related PR in this repo. I expected he will be eliable reviewer based on his role and position within the process community. I had no intention to check the whole CODEOWNERS file for missing or wrong assignment, rather perform an atomic fix in a single line, where I spotted a potential flaw. (Change sets should be small, so that they can be understood and easily merged. This was my further rational.
Based on the feedback from @masc2023 and @FScholPer, I will close this PR with this comment. It can be reopened and merged if someone likes to.

.github/CODEOWNERS
anmittag: @masc2023 well 

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

tells also Teams are working if you ensure:

_"The people you choose as code owners must have write permissions for the repository. When the code owner is a team, that **team must be visible and it must have write permissions, even if all the individual members of the team already have write permissions directly**, through organization membership, or through another team membership."_

masc2023: It is a technical topics i am referring to, lets discuss that next year with @aschemmel-tech , as he tried it out and did not work as expected

Issue #1091, opened 2025-05-15
Bugfix: Practical execution of verificaiton activities


pahmann:
> [!IMPORTANT]
> Make sure to link this issue with the PR for your bugfix.

Container issue to cover verification audit findings, which are on platform level and related to actual implementation/execution:

- [ ] https://github.com/eclipse-score/score/issues/608
- [ ]

PR #2364, opened 2025-12-19
add manual test execution guidelines to software verification plan

pahmann:
Resolves: #1091

### DIFF
--- a/docs/platform_management_plan/software_verification.rst
+++ b/docs/platform_management_plan/software_verification.rst
@@ -361,6 +361,25 @@ Test execution and result analysis
 The execution of the tests is based on a full automation defined by build pipelines. The analysis of the test results
 needs to be performed by the contributor.
 
+Manual test execution
+^^^^^^^^^^^^^^^^^^^^^
+
+The automation rate for test case execution is expected to be above 99%.
+
+Automated tests can be executed locally by the contributor before pushing code to the repository.
+To support local test execution respective `bazel test ...` commands are provided as part of the module and feature documentation.
+
+When a manual test is considered to verify a requirement or architectural element, the respective links need to be established as for any other automated test case.
+The script for manual testing uses the same templates as any other automated test including the Record Properties.
+Therefore, these test cases require a script to be provided which describes how to execute the test case manually step by step.
+The script and its execution will be part of the test reporting and traceability and need to be reviewed in the PR as part of the verification activities.
+Within the review an approver confirms that automation is not feasible for the specific test case.
+
+In contrast to automated result comparison of an automated test case, the feedback from the test executor is logged for manual tests.
+The script will wait for the user input to proceed to the next step and describe the expected result after each step.
+The user input can be judged as confirmation that the expected result is achieved.
+The final result will be `PASSED` or `FAILED`, based on the test executors confirmation.
+
 Test selection and regression testing
 -------------------------------------
 


docs/platform_management_plan/software_verification.rst
```diff
@@ -361,6 +361,25 @@ Test execution and result analysis
...
+To support local test execution respective `bazel test ...` commands are provided as part of the module and feature documentation.
+
+When a manual test is considered to verify a requirement or architectural element, the respective links need to be established as for any other automated test case.
+The script for manual testing uses the same templates as any other automated test including the Record Properties.
+Therefore, these test cases require a script to be provided which describes how to execute the test case manually step by step.
+The script and its execution will be part of the test reporting and traceability and need to be reviewed in the PR as part of the verification activities.
+Within the review an approver confirms that automation is not feasible for the specific test case.
+
+In contrast to automated result comparison of an automated test case, the feedback from the test executor is logged for manual tests.
+The script will wait for the user input to proceed to the next step and describe the expected result after each step.
```
PiotrKorkus, line 379: what script? I don't fully understand this paragraph

PiotrKorkus (commented):
I think for manual test cases we shall have also a documented reference setup / container that should be used for verification. Just to prevent "it works on my machine".

Once a manual test case is present in a feature repo do we require every PR to have it executed and confirmed before merging? How do we ensure that?

docs/platform_management_plan/software_verification.rst
pahmann: Okay. The idea is to put your test specification in an e.g. host executed python script which will say something like "Given that you do this and that, when we are in this and that setup, then the output should be a message of kind..." This would be confirmed with a "pass" or a "y/n" response. In this way you would manually execute a test, but you have the script as guidance. When the script is called by a framework you have proper logs and a reproduction scenario. You can easily achieve it via a helper function which prints out the steps on a terminal and waits for user interaction as a response. When this is host based test you can even ignore an ssh connection to the hardware or so (depending on test case). 

Does it make sense and becomes clear?

PiotrKorkus: I understand now but what's the point of creating a script to interact with a user? Isn't proper documentation with test steps and expected outcome enough?

pahmann: I would see two points. It is properly documented, but people still do something different or jump over a step and add something in between. This can also not be avoided with a script, but it can provide additional logs which relate to the execution. e.g. when something is typed in a terminal or order or execution has changed. (This happened in one of the projects when I was leading a team of test developers.)
My 2nd part would be to have an xml file after execution which also includes the Record Properties and so on. Whenever a test case will be executed for a release and it links to a requirement the traceability to execution should be there and a log/report of the result should be visible. With a manual execution script we make sure it has the proper format and it can be automatically processed after execution for report collection and generation.

pahmann: > I understand now but what's the point of creating a script to interact with a user? Isn't proper documentation with test steps and expected outcome enough?

One more point to consider. So far all our test specification is done within the repository. So basically it is all documented in the test code. How would you do the documentation of the manual tests? Would you use rst files for them? And where will you document the results? Via a Pull Request? I may be driven by the past where we have specified all test cases in robot framework and manual test cases felt a lot like automated test cases.

PiotrKorkus: yeah, that would require writing dedicated tests in .rst files. Makes sense to have everything same way, in the same place. Maybe writing a pytest (we decided we won't use robot) is a good idea.
But we shall consider how we want to ensure fail/pass doesn't just happen on a single machine. And who will be responsible on checking those manual tests on PR/main/release?

pahmann: pytest is what I was also thinking about.
I would try to find an argument that manual test cases are only executed for specific releases not for every PR. It could be an idea to execute once a month on main, so we do not get surprises during a release, but still keep the effort low. The actual execution may need to be done by someone from the Testing Community. Maybe we take a release candidate (e.g. when we are closer to a release) and execute the tests then. The maintainer of the component/feature/platform test will confirm the proper execution. We can also swap and let the maintainer execute the test and testing community reviews. But at least this should overcome some single person effects. Best would be two people execute, but this would even increase efforts. 
The execution would in best case be on the machine where the reference hardware will be to avoid local machine execution. This would also be the place where the automated tests got executed and where the logs are stored. In this way, we guarantee that the right software was deployed and the right version of the test is executed and so on.

One point though is, that manual execution is always effort, so we should limit the number to the least possible, but should no get unexpected surprises when it comes to the formal release testing. For one project we executed the tests once a month (depending on changes in the component) where we had a formal deliverable which had to be fully tested once a quarter.

pahmann:
As overall there is agreement that manual test cases as pytest scripts can be a good idea to reduce infrastructure effects (e.g. by writing test specs in rst), we need to follow up on the execution to be as close to the automated test execution as possible. This is tracked in this ticket. #2368

pahmann:
@PiotrKorkus if you are fine with the latest suggestion, please approve this PR.

pahmann:
@masc2023 anything to add from your side or are you okay with approving this PR?

masc2023 (approved):

PR #2346, opened 2025-12-16
process:  add DR-001 for independent feature delivery

arsibo:
Add design decision to enable SCORE features as independent SEooC delivery units in dedicated repositories.

PR #2400, opened 2026-01-12
DR-005: Release strategy

olivembo:
Establish a polyrepo release process with a manifest repository and relaxed SemVer for coordinated product releases.

### DIFF
--- a/docs/design_decisions/DR-005-infra.rst
+++ b/docs/design_decisions/DR-005-infra.rst
@@ -117,7 +117,7 @@ product versions.
      arise when deviating from strict trunk-based development and introducing long-lived
      release branches. This is no longer considered state-of-the-art trunk-based development.
 
-.. code-block:: none
+.. mermaid::
 
      gitGraph
              commit id: "1.2.3"


docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,158 @@
...
+- Module developers must know how to name their released versions
+- It must be clear how to do the integration, means what to reference (e.g. „extended“ semver as 1.2.3-etas-r1.0)
+
+---
+
+## 3. Options Considered
+
+### 3.1 Trunk-Based Development with SemVer
+
+Assumes continuous deployment on a longlived branch, e.g. `main` and does not define coordinated product releases.
```
a-zw: This describes "trunk-based" with "does not define coordinated product releases". This implies that either

1. This is not valid option as it does not achieve the goals. It should not even be considered but removed from this decision record.
2. This option needs to describe how coordinate product releases given trunk-based development. The description must be extended accordingly.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,158 @@
...
+
+    checkout main
+    merge bugfix
+```
+> *Explaination:* Using SemVer quickly reaches its limits.
+If branching off from a version e.g. 1.2.3 and needing a bugfix while the development on main goes on where other versions e.g. 1.2.4 are created, now on a release branch you would need to use the version 1.2.5. The development on main for the next version needs to use 1.2.6 although the logical predecessor is 1.2.4. This violates SemVer since backward compatibility in that case is (or could be) broken.
+
+### 3.2 Gitflow Across Repositories
+
+Introduces coordinated release branching across repositories but lacks a central integration manifest.
```
a-zw: How does "coordinated release branching across repositories" work? This option is not described sufficiently.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,158 @@
...
+- Reproducible and auditable release snapshots
+- Explicit stabilization phases
+- Continuous integration before formal releases
+- Independent module lifecycles without blocking development
+- A process that scales across many repositories and teams
+- Working on the main branch should be always possible
+- Working on a release branch should be always possible and not harm the development on the main branch (resp. Vice versa)
+- Working on a bugfix should always be possible (for any old release)
+- Module developers must know how to name their released versions
+- It must be clear how to do the integration, means what to reference (e.g. „extended“ semver as 1.2.3-etas-r1.0)
```
a-zw: Option 3.3 lists "integration governance" as a Con. However, some people would consider that lack of a such governance as undesirable lack of control.

Add "minimal integration governance" as a goal to document our assumption that we want that.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,158 @@
...
+**Pros**:
+- Single source of truth for product integration.
+- Supports continuous verification.
+- Provides reproducible releases.
+- Scales with module count and team autonomy.
+- Clear separation between development, integration, and stabilization.
+
+**Cons**:
+- Requires explicit integration governance.
+- Introduces additional coordination effort compared to single-repo workflows.
```
a-zw:
```suggestion
```

This is not a Con but a goal. The goals section says "A process that scales across many repositories".

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,158 @@
...
+manifest-based integration and release trains (e.g. Android/AOSP, Chromium-style roll-ups),
+while remaining explicit, and flexible.
+
+It provides a single source of truth for integration, supports both continuous verification and reproducible releases, and scales with module count and team autonomy.
+
+Option 3.1 (Trunk-Based Development Only) has been rejected because it does not address the need for coordinated product releases or explicit stabilization phases in a poly repo environment.
+
+Option 3.2 (Gitflow Across Repositories) has been rejected because it requires coordinating release branches across all repositories and lacks a central integration manifest, which does not scale well.
+
+---
```
a-zw:
```suggestion
```

To fix build error.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,158 @@
...
+
+**Cons**:
+- Requires explicit integration governance.
+- Introduces additional coordination effort compared to single-repo workflows.
+
+---
+
+## 4. Decision
+
+We decided for **Option 3.3**.
```
a-zw: As of now this decision record is pointless. It puts up two strawmen options (3.1 and 3.2) and dismisses them for not achieving the goals. If there is only a single valid option, then there is no decision to be made.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,158 @@
...
+
+Commonly referenced workflows do not fully address this problem space:
+- **Trunk-Based Development** assumes continuous deployment and does not define coordinated product releases.
+- **Gitflow** introduces coordinated release branching across repositories and lacks a central integration manifest.
+
+This ADR defines a release process explicitly designed for **polyrepo systems with coordinated integration releases and continuous verification**.
+
+---
+
+## 2. Goals and Requirements
```
a-zw: Can we distinguish between requirements (which must be satisfied) and optimization goals (where we try to find the best option)?

There is no point in invest time looking into option which don't satisfy requirements. Also, goals are usually few and that simplifies the decision evaluation.

docs/design_decisions/DR-005-infra.md
olivembo: I will adapt the document according to 2.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+
+- Reproducible and auditable release snapshots
+- Explicit stabilization phases
+- Continuous integration before formal releases
+- Independent module lifecycles without blocking development
+- A process that scales across many repositories and teams
+- Working on the main branch should be always possible
```
opajonk: What does "working on the main branch" mean here? I understand the main branch like this: "latest and greatest, always in a working state". That means you cannot actually "work on" the main branch: you must always branch off, then work, make sure the work is "fully done", integrate.

I would also use e.g. "codeline" instead of "branch", and define codeline as "The set of branches of the same name across all S-CORE repositories.". If useful, one can go more into details and write "Each of the branches of a codeline must have been created from the same head branch, in a state consistent with a head codeline."

docs/design_decisions/DR-005-infra.md
opajonk: This would be helpful. The amount of strict requirements are probably much lower than the goals.

docs/design_decisions/DR-005-infra.md
olivembo: Will describe in further detail

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+```{mermaid}
+gitGraph
+    commit id: "1.2.3"
+
+    branch release
+    checkout release
+    commit id: "1.2.3-r1.0"
+
+    checkout main
+    branch bugfix
```
opajonk: I think this is likely not correct. You need to create a support branch for 1.2.x, which starts at 1.2.0, where you can create 1.2.1, 1.2.2, 1.2.3, 1. 1.2.4 etc. from. These cannot come from a main line, where development for 1.3.x (or even 2.x.x) continues. The patch releases are by defintion only bugfix releases and cannot contain new features.

(This assumes that on main you continue with new features, i.e. minor updates in the sense of SemVer)

I think GitFlow actually does a good job to capture all of this (main, release, support, bugix, how to connect branches to SemVer).

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+### 3.2 Gitflow Across Repositories
+
+Introduces coordinated release branching across repositories but lacks a central integration manifest.
+
+**Pros**:
+- Well-known branching model.
+- Provides release branches for stabilization.
+
+**Cons**:
+- Requires coordination of release branches across all repositories.
```
opajonk: I don't get this point: the branching model simply deals with a single repository. If you want to enhance it for a multi-repo, you have to do something in addition. I do not see this as a drawback really... a branching model alone can **never** do this. You need something like the reference_integration, with its lockfile method (this json file we create is nothing else).

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+- Lacks a single source of truth for integration state.
+- Does not scale well with increasing module count.
```
opajonk: Again, this is comparing apples and oranges...

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+
+    checkout release
+    merge bugfix
+    commit id: "1.2.5"
+
+    checkout main
+    merge bugfix
+```
+> *Explaination:* Using SemVer quickly reaches its limits.
+If branching off from a version e.g. 1.2.3 and needing a bugfix while the development on main goes on where other versions e.g. 1.2.4 are created, now on a release branch you would need to use the version 1.2.5. The development on main for the next version needs to use 1.2.6 although the logical predecessor is 1.2.4. This violates SemVer since backward compatibility in that case is (or could be) broken.
```
opajonk: As stated above I think this is not correct. SemVer has no issues with this, but the connection to the branching model must be done correctly.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+- Provides release branches for stabilization.
+
+**Cons**:
+- Requires coordination of release branches across all repositories.
+- Lacks a single source of truth for integration state.
+- Does not scale well with increasing module count.
+
+### 3.3 Polyrepo Release Process with Manifest Repository and relaxed version of SemVer
+
+As described in [DR-002 (Integration Testing in a Distributed Monolith)](./DR-002-infra.md) there is a dedicated manifest repository containing the "known goods sets". There is a known good set for the latest version, but also known good sets for release(d) versions. Because of the earlier described limitations of SemVer, the correct module versions should not be referenced in the `MODULE.bazel` (in the manifest repository) as e.g., `1.2.3` but either by referencing the git commit hash directly or using a relaxed SemVer string, e.g. as `1.2.3-v1.0` where the string after the hyphen represents the respective S-CORE release.
```
opajonk: I do not agree here. We can simply cut a release branch of all modules, which "captures" a minor version of the module (i.e. all 1.1.x versions). The versions 1.1.0, 1.1.1 etc are the "line of module versions agreed for a specific S-CORE release". S-CORE-bugfix releases then take bugfix releases of the modules. etc.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+
+With that approach releases are possible, e.g. by creating a release branch in the integration repository as well as in the affected module repositories. Also bugfixes of "old" releases are possible by checking out the repective release branch in the reference integration and if necessary to also created bugfixes in the affected modules.
```
opajonk: This is correct, but as outlined above it does not really require a "modified SemVer". The standard one would do :-)

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+---
+
+## 4. Decision
+
+We decided for **Option 3.3**.
+
+**Rationale**
+
+This approach reflects established industry practice for large-scale polyrepo systems using
+manifest-based integration and release trains (e.g. Android/AOSP, Chromium-style roll-ups),
```
opajonk: In the end I agree  with the conclusion: we need a manifest / lock file / call-it-what-you-want. But we do not need special branching methods or a new mapping to SemVer. The standard stuff does the job, when thinking about "codelines" (= sets of branches with same name/purpose) instead of "single-repository-branches".

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,156 @@
...
+while remaining explicit, and flexible.
+
+It provides a single source of truth for integration, supports both continuous verification and reproducible releases, and scales with module count and team autonomy.
+
+Option 3.1 (Trunk-Based Development Only) has been rejected because it does not address the need for coordinated product releases or explicit stabilization phases in a poly repo environment.
+
+Option 3.2 (Gitflow Across Repositories) has been rejected because it requires coordinating release branches across all repositories and lacks a central integration manifest, which does not scale well.
```
opajonk: With this I am a bit confused: of course we will need coordinated release branches across all repositories - simply for consistency reasons during stabilization. There is **no way** that a module introduces new features into a release when stabilization is already happening.

However, release branch cutting (i.e. "code freeze") is **trivially scriptable**, and a matter of 3 minutes across all repositories - given that you use a workspace like available via gitea on reference-integration. All you have to do is to announce "code freeze day & time" and fire the script. This is absolutely no problem if your mainline is always stable and integrates well. And this we have already outlined in DR-002.

docs/design_decisions/DR-005-infra.md
olivembo: Will add this point to the goals section

docs/design_decisions/DR-005-infra.md
olivembo: The problems with SemVer are described here: https://beza1e1.tuxen.de/patching_old_versions.html

Do you agree with the limitations described in there?

docs/design_decisions/DR-005-infra.md
a-zw: The point is that module devs should be able to push new features on their module main branch while the integration is still busy integrating an older version. No announcements like "only merge bugfixes until the 0.5 release final is done".

In other words, it implies that modules must have some kind of release branches for hotfixes.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,162 @@
...
+---
+
+## 2. Requirements and Goals
+
+### 2.1 Requirements
+
+Options that do not satisfy these requirements are not viable and will be rejected:
+
+- Reproducible and auditable release snapshots
+- Working on the main branch should be always possible
```
AlexanderLanin: Within DR-002 we agreed that integration can block breaking changes / issues.
```suggestion
- Working on the module main branch should be always possible (except when integration PR checks block due to issues)
```

Alternatively more on the point @a-zw was mentioning:
```suggestion
- Working on the module main branch should be always possible (no "no new features now, we are releasing soon")
```

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,162 @@
...
+- Working on a release branch should be always possible and not harm the development on the main branch (resp. Vice versa)
+- Working on a bugfix should always be possible (for any old release)
+- Module developers must know how to name their released versions
+- The versioning scheme must clearly indicate which product release a module version belongs to, enabling parallel release streams
```
AlexanderLanin: That sounds good, but actually I don't understand it ;-)

Version of the module should indicate which module release the version belongs to?? Is there any difference between those two terms?

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,162 @@
...
+
+### 2.2 Optimization Goals
+
+Among viable options, we optimize for:
+
+- Explicit stabilization phases
```
AlexanderLanin: Stabilization should not be required. It often is, due to bad automation etc... but it shouldn't be. So let's not advertise that we will have stabilization phases.

```suggestion
- Enabling stabilization phases if required
```

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,162 @@
...
+- A process that scales across many repositories and teams
+- Minimal integration governance to coordinate releases across repositories
+
+---
+
+## 3. Options Considered
+
+### 3.1 Trunk-Based Development with SemVer
+
+Uses trunk-based development where modules develop on `main` branches and release with standard SemVer tags. The manifest repository (per DR-002) continuously tracks the latest module versions. When a product release is needed, release branches are created in the manifest and affected module repositories. Module bugfixes on release branches are tagged with standard SemVer, and the manifest's release branch references these versions.
```
AlexanderLanin: We should not affect module repositories by product releases.

docs/design_decisions/DR-005-infra.md
AlexanderLanin: * In SemVer there is no branching off 1.2.3 after 1.2.4 has been created. So maybe the limit you mean is not a complex diagram but simply "patching an old patch release is not possible in semver"?
* The diagram shows that there is a gap between 1.2.4 and 1.2.6, which is later filled by 1.2.5. Is it meant that 1.2.6 is made after 1.2.5? Can you adjust the timeline? (but see point 1)
* Can you copy the relevant explanations here from external sites? (but see point 1)

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,162 @@
...
+
+**Pros**:
+- Simplifies development workflow.
+- Encourages frequent integration.
+- Well-understood versioning convention.
+
+**Cons**:
+- Standard SemVer cannot represent parallel release streams. When a bugfix is needed on a release branch while `main` has advanced, version numbering becomes ambiguous (see diagram below).
```
AlexanderLanin: SemVer can represent parallel release streams for different major versions. What you have below is a branching scheme, and the problem is that you need to define whether you work on main or on the branch for 1.2.x. It's also possible to work on main up to 1.2.3, and then switch to a branch for 1.2.4, while main goes on to 1.3 or 2.0.

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,162 @@
...
+
+**Pros**:
+- Well-known branching model.
+- Explicit `develop` branch separates ongoing work from release stabilization.
+- Release branches provide clear stabilization phases.
+
+**Cons**:
+- Standard SemVer suffers from the same parallel release stream problem as Option 3.1 (version numbering conflicts).
+- Additional overhead of maintaining separate `develop` branches across all repositories.
+- More complex branching model increases coordination complexity in a polyrepo setup.
```
FScholPer: This is the most important topic. Are we willing and are we able to handle the complexity with that branching model. Copied from https://docs.aws.amazon.com/prescriptive-guidance/latest/choosing-git-branch-approach/advantages-and-disadvantages-of-the-gitflow-strategy.html

**Gitflow-based development does have some disadvantages that can impact the development process and the team dynamics. The following are a few notable drawbacks:**


Complexity 
– Gitflow is a complex pattern for new teams to learn, and you must adhere to the rules of Gitflow to use this successfully.

Continuous deployment 
– Gitflow doesn't fit a model where many deployments are released to production in a rapid fashion. This is because Gitflow requires the use of multiple branches and a strict workflow governing the release branch.

Branch management 
 – Gitflow uses many branches, which can become burdensome to maintain. It can be challenging to track the various branches and merge released code in order to keep the branches properly aligned with each other.

Technical debt 
 – Because Gitflow releases are typically slower than the other branching models, more features can accumulate for release, which can cause technical debt to accumulate.

Teams should carefully consider these drawbacks when deciding whether Gitflow-based development is the right approach for their project.

docs/design_decisions/DR-005-infra.md
opajonk: OK so the "normal codefreeze concept", right?

"On the 31st January, 15:00 we will cut the release branches for 1.0.0" --> on all modules, of course. Release branch creation takes a minute, and just after that everyone can continue to work on the mainline.

docs/design_decisions/DR-005-infra.md
opajonk: Yes, sure - but that description is purely about "What to do if you need to do a **breaking change** to an old version?" - so strictly about **major versions**.

Why would you do a **breaking change** to an already released major version? You can fix bugs, you can add **compatible** features, but **breaking change** simply means: new major version. And you cannot break an existing major release. I think we've picked the development model that is described below "Instead I want to hint at a solution on a higher level." in the text?

opajonk: (Just to be clear: I do not see SemVer religiously as the secret sauce that solves everything :smile: - for example, when distributions come into play and distributors are patching / backporting / ... it stops working and gets really messy. Cf. the ["versioning scheme"](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version) of Debian, which is more or less a collection of necessities created over years of experience...

However, S-CORE is (here, during regular development at least!) not a distribution, right?)

arsibo:
Please rename to DR-006. I use DR-005 https://github.com/eclipse-score/score/pull/2346

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,162 @@
...
+- Does not scale well with increasing module count.
+
+### 3.3 Polyrepo Release Process with Manifest Repository and relaxed version of SemVer
+
+As described in [DR-002 (Integration Testing in a Distributed Monolith)](./DR-002-infra.md) there is a dedicated manifest repository containing the "known goods sets". There is a known good set for the latest version, but also known good sets for release(d) versions. Because of the earlier described limitations of SemVer, the correct module versions should not be referenced in the `MODULE.bazel` (in the manifest repository) as e.g., `1.2.3` but either by referencing the git commit hash directly or using a relaxed SemVer string, e.g. as `1.2.3-v1.0` where the string after the hyphen represents the respective S-CORE release.
+
+With that approach releases are possible, e.g. by creating a release branch in the integration repository as well as in the affected module repositories. Also bugfixes of "old" releases are possible by checking out the repective release branch in the reference integration and if necessary to also created bugfixes in the affected modules.
```
guejan:
```suggestion
With that approach releases are possible, e.g. by creating a release branch in the integration repository as well as in the affected module repositories. Also bugfixes of "old" releases are possible by checking out the respective release branch in the reference integration and if necessary to also created bugfixes in the affected modules.
```

docs/design_decisions/DR-005-infra.md
```diff
@@ -0,0 +1,162 @@
...
+
+**Pros**:
+- Single source of truth for product integration.
+- Supports continuous verification.
+- Provides reproducible releases.
+- Scales with module count and team autonomy.
+- Clear separation between development, integration, and stabilization.
+
+**Cons**:
+<<<<<<< HEAD
```
guejan:
```suggestion

```
Unresolved merge conflict.

docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,212 @@
...
+^^^^^^^^^^^^
+
+Options that do not satisfy these requirements are not viable and will be rejected:
+
+- Reproducible and auditable release snapshots
+- During working on a release, working on the main branch should still be possible
+- Working on a release branch should be always possible and not harm the development on the main branch (resp. vice versa)
+- Working on a bugfix should always be possible (for any old release)
+- Module developers must know how to name their released versions
+- The versioning scheme must clearly indicate which product release a module version
```
AlexanderLanin, line 68: Sorry, I don't understand the sentence. How can a module version indicate which product release it belongs to? 

Is this something like: Product releases must have a defined, clear and reproducible manifest of the contained module versions?

docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,212 @@
...
+- Maximizes integration and fast feedback through continuous integration.
+- Minimizes merge conflicts and versioning issues.
+
+**Cons**
+
+- Parallel maintenance of multiple product versions (e.g., for long-term support) is
+    not feasible.
+- Not suitable for polyrepo setups that require parallel support of several releases,
+    as parallel release streams are not supported by design.
+- Requires explicit integration governance.
```
AlexanderLanin, line 113: What do you mean here with integration governance?

docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,212 @@
...
+
+.. note::
+     The versioning and merge conflicts described in the diagram below typically
+     arise when deviating from strict trunk-based development and introducing long-lived
+     release branches. This is no longer considered state-of-the-art trunk-based development.
+
+.. code-block:: none
```
AlexanderLanin:
```suggestion
.. mermaid::
```

docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,212 @@
...
+**Cons**
+
+- Parallel maintenance of multiple product versions (e.g., for long-term support) is
+    not feasible.
+- Not suitable for polyrepo setups that require parallel support of several releases,
+    as parallel release streams are not supported by design.
+- Requires explicit integration governance.
+
+.. note::
+     The versioning and merge conflicts described in the diagram below typically
```
AlexanderLanin, line 116: The diagram block is not quite correct: Once you start a 1.2 branch you cannot release 1.2.x from main.

I don't know what you wanted to show here, but here is a trunk-based release branching variant:
https://mermaid.ai/play?utm_medium=toggle&utm_source=mermaid_live_editor#pako:eNqrVkrOT0lVslJSAIH0zBL3osSCjJg8BWSQnJ-bm1mikJlipRCjZKhnpGcco0RIiQkOJTF5aMJJRYl5yRkKQD3o6jNSk7PzS0uwSWE3Cq4jNzEzD4cWfK421jMAuZpUdyD52hSkX6kWAP8yU74

AlexanderLanin: Or is this more about simply showing that you need branching discipline? Or limitations of SemVer?

docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,212 @@
...
+
+- Well-known branching model.
+- Explicit ``develop`` branch separates ongoing work from release stabilization.
+- Release branches provide clear stabilization phases.
+
+**Cons**
+
+- Parallel maintenance of multiple product versions (e.g., for long-term support) is
+    not feasible.
+- Standard SemVer suffers from the same parallel release stream problem as Option 3.1 (version numbering conflicts).
```
AlexanderLanin, line 167: Ah ok, so that clarifies my questions above 😆 

But that would be a SemVer limitation, and not branching.

docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Bugfixes for previous releases are handled by checking out the corresponding commits
+in the manifest and the affected modules, applying the fix, and updating the
+manifest to reference the new module version.
+
+**Pros**
+
+- Single source of truth for product integration.
+- Supports continuous verification.
+- Provides reproducible releases.
```
AlexanderLanin, line 192: thats manifest in general as implemented and applies to all 3 options?

docs/design_decisions/DR-005-infra.rst
```diff
@@ -0,0 +1,212 @@
...
+
+**Cons**
+
+- Requires explicit integration governance.
+
+
+Decision
+--------
+
+We decided for **Option 3.3**.
```
AlexanderLanin, line 204: I'm sorry, I'm lost. Option 3.3 does not describe branching. It's only about a manifest.

AlexanderLanin:
Maybe its easier to focus on a single branching model instead of alternatives?

olivembo:
> Maybe its easier to focus on a single branching model instead of alternatives?

The single branching model does not work for scenarios where you have to maintain old releases.

docs/design_decisions/DR-005-infra.rst
olivembo: That somebody must have the governance of the integration process and the repository

docs/design_decisions/DR-005-infra.rst
olivembo: It should show that SemVer fails in some cases, see https://beza1e1.tuxen.de/patching_old_versions.html

docs/design_decisions/DR-005-infra.rst
olivembo: see comment above (https://beza1e1.tuxen.de/patching_old_versions.html)

docs/design_decisions/DR-005-infra.rst
olivembo: no, only to 3.3

docs/design_decisions/DR-005-infra.rst
olivembo: > Bugfixes for previous releases are handled by checking out the corresponding commits
> in the manifest and the affected modules, applying the fix, and updating the
> manifest to reference the new module version.

This implicitly means branching.

AlexanderLanin:
> > Maybe its easier to focus on a single branching model instead of alternatives?
> 
> The single branching model does not work for scenarios where you have to maintain old releases.

I meant focus on one model, instead of describing 3 alternatives.

FScholPer:
@olivembo can you please update that so that we can merge this to main?

docs/design_decisions/DR-005-infra.rst
olivembo: e.g. v1.2.3_score_v1.0_beta

PR #2431, opened 2026-01-14
Update score to new docs_as_code and process version

pahmann:
This requires also to update the wp__tailoring of work products as this has been modified in the process_description repo to differntiate beween generic work product tailoring and safety tailoring.


AlexanderLanin (approved):
bazel_dep changes LGTM

PR #2475, opened 2026-01-20
 execution contract

dcalavrezo-qorix:


docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+
+---
+
+## Purpose
+
+This document defines the execution contract for S-CORE builds across developer machines
+and CI infrastructure.
+Its goal is to ensure **long-term reproducibility (≥10 years)**, **traceability** and
+**practical hermeticity**, despite changes in underlying infrastructure such as
+GitHub-hosted runners.
```
AlexanderLanin, line 36:
```suggestion
GitHub-hosted runners.

It builds on [DR-001], which was concerned about the same topics, but was focused on tools only. This adds details where the original description was too fuzzy.
```

(not sure about the syntax)

For me it makes all the difference, that this is perfectly aligned, and not something different.

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+---
+
+### R3 — Hermeticity (Practical)
+Build actions must not depend on **undeclared inputs**.
+
+In practice:
+- Tools affecting build outputs must either be:
+  - managed by Bazel, or
+  - explicitly injected as Bazel action inputs, or
+  - reflected in cache partitioning
```
AlexanderLanin: not sure what you mean here. Theoretically its enough if they are documented as in R2, but of course we want more.

What about "mirrorable"... no idea how to describe it. I'm talking about pypi for example.

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+- Reliance on host state must be minimized and documented where unavoidable.
```
AlexanderLanin, line 92: This is not the same indentation as above. So "Tools affecting build outputs must either be...documented" is missing?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+This information must be recorded in a **build manifest** and stored alongside the build
+outputs.
+
+---
+
+### R3 — Hermeticity (Practical)
+Build actions must not depend on **undeclared inputs**.
+
+In practice:
+- Tools affecting build outputs must either be:
```
AlexanderLanin: Let's explain "Tools affecting build outputs" to be very clear. With example of such tools. And tools that are less relevant.

e.g. I'm currently not sure whether pytest affects build outputs. Are test results build output?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+  - reflected in cache partitioning
+- Reliance on host state must be minimized and documented where unavoidable.
+
+Perfect hermeticity is not required, but **undeclared variability is not acceptable**.
+
+---
+
+## Three-Layer Execution Contract
+
+### Layer 1 — Host Platform Contract
```
AlexanderLanin, line 100: x86, arm for Macs would also belong to host?!

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+- Linux kernel version and configuration
+- Security mechanisms (AppArmor / LSM)
+- Filesystems, networking, namespaces
+- Support for:
+  - Bazel `linux-sandbox`
+  - QEMU / `binfmt`
+  - privileged operations where required
+
+#### Requirements
+- Linux host OS (Ubuntu LTS for reference integration)
```
AlexanderLanin: There is an endless discussion going on for host platforms. Let's not get involved.

```suggestion
- Linux host OS (For now Ubuntu LTS for reference integration)
```

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+- Ubuntu 24.04+ AppArmor may block Bazel sandbox mount operations
+- Containers **cannot** mitigate host-kernel restrictions
+
+#### Policy
+- A **Reference Integration Host Baseline** must be defined (e.g. Ubuntu 22.04).
+- Deviations (e.g. privileged runners, sandbox disabled) must be explicit and isolated.
+
+---
+
+### Layer 2 — Execution Context Contract (Devcontainer)
```
AlexanderLanin: Do we already assume builds are executed in a devcontainer? So far we only agreed that the development environment is in devcontainers? Maybe this layer is optional?

```suggestion
### Layer 2 — Execution Context Contract (e.g. native or devcontainer)
```

AlexanderLanin: Ah, the last point says its optional. Thats quite late ;-)

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+
+This layer defines the **default user-space environment** in which builds are executed.
+
+#### Purpose
+- Provide consistent runtime ABI (`glibc`, `libstdc++`)
+- Ensure tool binaries (e.g. rustc) can execute reliably
+- Eliminate “works on my machine” discrepancies
+- Enable local reproduction of CI builds
```
AlexanderLanin, line 149: We need to figure out what exactly has to be identical. Or is it this list?
e.g. devcontainer is based on Ubuntu 24. GitHub Runners are based on Ubuntu 24. Is it now enough to ensure the same python version for python scripts?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+
+This program and the accompanying materials are made available under the
+terms of the Apache License Version 2.0 which is available at
+https://www.apache.org/licenses/LICENSE-2.0
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+
+# DR-001-Infra-Extension: S-CORE Build Execution Contract
```
AlexanderLanin: 001-B? 006?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+- Builds should still be possible on compatible bare-metal hosts.
+
+---
+
+### Layer 3 — Bazel Contract
+
+This layer defines **what Bazel controls and guarantees**.
+
+#### Bazel Versioning
+- Each repository must contain `.bazelversion`.
```
AlexanderLanin: not every repository is using bazel... well, again it depends on which repositories are relevant here.

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+  - host security configuration
+
+These constraints must be handled in Layer 1 and Layer 2.
+
+---
+
+## Minimum Supported Baselines
+
+### OS and Runtime Baseline
+- Minimum supported baseline: **Ubuntu 20.04 LTS** (subject to revision)
```
AlexanderLanin: Baseline is defined in layer 2. So layer 3 should not mention an exact version?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,222 @@
...
+- Toolchains must be built against this baseline
+- Older environments are **not supported**
+
+### Rationale
+We explicitly do **not** support all historical `glibc` or kernel versions.
+Portability is achieved by choosing and documenting a baseline, not by unlimited
+backward compatibility.
```
AlexanderLanin:
```suggestion
backward compatibility. Layer 2 can easily be virtualized as needed, for future reproducibility.
```

AlexanderLanin (commented):
Lots of comments, but this was a great read! Good content IMHO!

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: Test results are not build outputs, but they do affect CI decisions ( I guess quality gates) so the tools that produce them must still be Bazel-visible if you want correctness, reproducibility and traceability.

Theoretically  we should use Bazel Pythong rules ( rules_python). This would ensure:

- Reproducible test outcomes
- Correct test caching

But I guess it is an acceptable fallback to have Pytest installed in the devcontainer.  

Do we have a file like a fingerprint.txt which  contains the versions of the tools?
```
pytest==7.4.2
python==3.11.6
container=sha256:deadbeeeeeeeeef.......

```
 That one can be used as an input for the test action. That way Bazel sees when the fingeprint changes and it can invalidate the caches accordingly

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: Good point

dcalavrezo-qorix: But I've no idea what's the correct list of supported `archs` for hosts. Who could help clarify this?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,224 @@
...
+outputs.
+
+---
+
+### R3 — Hermeticity (Practical)
+Build actions must not depend on **undeclared inputs**.
+
+In practice:
+- Tools affecting build outputs must either be:
+  - managed by Bazel, or
```
MaximilianSoerenPollak, line 87: The writing here suggests that if it's managed by bazel it then automatically is solved and we don't need to worry about it anymore. 
Is that the case, or am I miss-reading here?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### Responsibilities
+- User-space runtime libraries
+- Bootstrap tooling (git, bash, coreutils, python, etc.)
+- Bazel entrypoint (preferably Bazelisk)
+- Development UX tooling (optional)
+
+#### Non-Goals
+- The devcontainer must **not silently override** repository-declared Bazel versions.
+- The devcontainer must **not be the only place** where critical tool versions are defined.
```
MaximilianSoerenPollak, line 187: Where else would you define it then? In a global ledger or so?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,224 @@
...
+- Each repository must contain `.bazelversion`.
+- S-CORE uses a **single Bazel version** across repositories.
+- CI enforces version consistency.
+
+#### Toolchains and Tools
+- Toolchains (e.g. Rust/Ferrocene, C/C++) must be:
+  - versioned
+  - immutable
+  - built against a documented baseline
+- Tools affecting outputs must be known to Bazel or reflected in action inputs.
```
MaximilianSoerenPollak: Outputs in which way? 
Does this mean for example if something saves a .json that is used as cache it should only work via Bazel actions?
Or what about the test frameworks that can affect the output xml?

MaximilianSoerenPollak:
Thanks for the write up, looks over all quite good and a great baseline we can work out the small other stuff.

docs/design_decisions/DR-001-infra-extension.md
lurtz: you can specify at Github Action Workflows that a specific container image and version shall be used: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idcontainerimage

Then there is no fingers crossed anymore if CI and devcontainer have the same tools and versions installed.

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,224 @@
...
+- A **versioned devcontainer image** is the default execution context.
+- The container image must be:
+  - built from a known OS baseline (Ubuntu LTS)
+  - referenced by immutable digest
+  - archived for long-term reproducibility
+
+#### Responsibilities
+- User-space runtime libraries
+- Bootstrap tooling (git, bash, coreutils, python, etc.)
+- Bazel entrypoint (preferably Bazelisk)
```
lurtz, line 176: I am a bit puzzled. IIRC Bazelisk transparently downloads whatever bazel version has been specified via `.bazelversion` from the internet. How will this ensure stable builds in like 10+ years? We have no guarantee that the server address is still the same and the version still available.

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,224 @@
...
+S-CORE builds must be reproducible for **at least 10 years** after creation, given:
+- the source revision
+- archived execution context
+- archived toolchains
+- recorded build metadata
+
+This must remain possible even if:
+- GitHub runner images change or are retired
+- upstream toolchains are no longer available
+- external services are unavailable
```
lurtz, line 57: should we have a test build, which has no access to the internet? This way we can be sure all needed tools and code are present.

docs/design_decisions/DR-001-infra-extension.md
lurtz: I have doubts about that. bazelisk downloads even bazel from the Internet, which then might download more dependencies.

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,224 @@
...
+#### Purpose
+- Provide consistent runtime ABI (`glibc`, `libstdc++`)
+- Ensure tool binaries (e.g. rustc) can execute reliably
+- Eliminate “works on my machine” discrepancies
+- Enable local reproduction of CI builds
+
+#### Definition
+- A **versioned devcontainer image** is the default execution context.
+- The container image must be:
+  - built from a known OS baseline (Ubuntu LTS)
```
lurtz: we should keep in mind that Microsoft removed support for older Ubuntu versions running in devcontainers from Visual Studio Code. Thus if your devcontainer image is based on Ubuntu 18.04, you are at the moment only able to run it via command line.

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: @lurtz correct
I guess it's time we start converting our workflows into using the `devcontainers`
Not sure if there's something in particular needed ( in terms of configuration of the container) for the Bazel local-cache.

@AlexanderLanin  if we go ahead with using `devcontainers`, we can make the clean-up action even more aggresive

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: solved

docs/design_decisions/DR-001-infra-extension.md
MaximilianSoerenPollak: So we for sure would need to have a test build for official releases that checks if the archiving is complete by building it without access to the internet?

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: I think this is relevant for the underlying OS/ kernel version/ glibc

docs/design_decisions/DR-001-infra-extension.md
opajonk: At least devcontainers we build for both `arm64` and `x86_64`. I would say at the moment that pretty much covers all hosts. I successfully ran devcontainer on an Apple M3 (`arm64`) MacOS, Windows WSL2 (`arm64` (Snapdragon Laptop) and `x86_64`), Linux (`arm64` (Snapdragon Laptop) and `x86_64`). I cannot imagine what else we need.

opajonk: Note: ran the devcontainer, started the tools - **not**: successfully built S-CORE. The issue here is: the Bazelized tools are not all available for `arm64`.

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: I think that would only work *IF* we would have a remote repository cached that is populated for Bazel fetching or we configure  our own local mirrors.

Or?

How would you see that? For the `devcontainer` part I think it is straightforward, for Bazel, I'm not too sure

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: very good point !

So we might loose the "works in VS Code" and just be stuck with CLI-only

@lurtz  can you propose a better wording, please?

```
#### Definition
- A **versioned devcontainer image** is the default execution context for CI and local builds.
- The container image must be:
  - built from a **defined Ubuntu LTS baseline**
  - compatible with common developer tooling (e.g. VS Code Dev Containers)
  - referenced by an **immutable image digest**
  - archived for **long-term reproducibility**
  ```

and maybe add a section like

```
#### Baseline Preservation and Reproducibility
- Once a devcontainer image is used in CI, its image digest becomes part of the build provenance.
- All such images must be archived and retrievable for **at least 10 years**.
- Reproducing historical builds may rely on legacy container runtimes or CLI-only execution,
  and does not require continued IDE support.
```

File: docs/design_decisions/DR-001-infra-extension.md (hunk 17 above)
Comment (lurtz) [reply docs/design_decisions/DR-001-infra-extension.md:57]: The bazel mirror would also need to act like an archive, because it should not remove anything needed to build a release for +10 years. And I guess all S-CORE builds should ideally make use of that mirror.

btw. in addition to bazel dependencies being downloaded from the internet, I expect that Rust code might also download some crates from crates.io.

If the needed dependencies are stored within the devcontainer, it would be rather easy, because you need the devcontainer image from the release and the S-CORE code and you should be able to build. No internet access or mirror needed.

File: docs/design_decisions/DR-001-infra-extension.md (hunk 16 above)
Comment (dcalavrezo-qorix) [reply docs/design_decisions/DR-001-infra-extension.md:176]: Good point again!

`Bazelisk` is fine as an entrypoint, but only if we pair it with an archived, controller source of Bazel binaries.

Would that mean to create our own internal mirror for Bazel releases (within eclipse).

What do you think?

File: docs/design_decisions/DR-001-infra-extension.md [hunk 19]
```diff
@@ -0,0 +1,224 @@
...
+  - referenced by immutable digest
+  - archived for long-term reproducibility
```
lurtz: original discussion: https://github.com/eclipse-score/score/pull/2475/changes/BASE..d8505443571c72e377d9d440da6d48884e27f80d#r2712981170

Main point was that newer devcontainer runtimes / VS Code might refuse to run quite old images which contain old operating systems.

```suggestion
#### Definition
- A **versioned devcontainer image** is the default execution context for CI and local builds.
- The container image must be:
  - built from a **defined Ubuntu LTS baseline**
  - compatible with common developer tooling (e.g. by following the [specification](https://containers.dev/))
  - referenced by an **immutable image digest**
  - archived for **long-term reproducibility**

#### Baseline Preservation and Reproducibility
- Once a devcontainer image is used in CI, its image digest becomes part of the build provenance
- All such images must be archived and retrievable for **at least 10 years**
- Reproducing historical builds may rely on legacy container runtimes or CLI-only execution,
  and does not require continued IDE support
```

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: "Outputs” should mean any build- or CI-relevant result that we rely on, not just binaries.

1) Build artifacts (most important)

Examples:
- binaries, libraries, containers, packages
- generated source code checked-in or shipped
- compiled outputs used downstream

2) Decision artifacts (CI gating) - I know that we don't have them, but I guess we need to add them at some point 😄 

Examples:

- test pass/fail outcome
- coverage percentage used as a gate
- lint results used to pass/fail

If a tool can change these → Bazel must track the tool version / inputs (or we risk wrong decisions / wrong cached results). Yeah, yeah, I know, where's the cache ?

docs/design_decisions/DR-001-infra-extension.md
lurtz: lets continue at https://github.com/eclipse-score/score/pull/2475#pullrequestreview-3692291515

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: Not a global ledger — the source of truth should be the repo (ex., .bazelversion, MODULE.bazel/lockfile, pinned toolchain deps). The devcontainer may provide binaries but it must not be the only place where versions are defined, otherwise changes to the container silently change builds. Guess this is part of the 3-layered approach.

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: @MaximilianSoerenPollak  @lurtz You are both correct

Maybe we should explicitly mention

 “Bazel-managed” improves cache correctness and traceability, but **does not by itself
  guarantee long-term reproducibility**; artifact availability must also be ensured
  via pinning, checksums, internal mirroring/archiving and offline verification.
  
  
  Ok, we do have our own Bazel Container Registry, but it will be probably a good idea to mirror the other artefacts that Bazel/bazeliks downloads ( even Bazel itsself, it can be that in 10 years, version `7.1.0 ` (random example) won't be available anymore for download).

docs/design_decisions/DR-001-infra-extension.md
opajonk: Weeeelll one can see this the other way around: the devcontainer defines the environment, also Bazelversion.

**BUT**

Changing the devcontainer in a repository is **not** silent. It is an explicit PR, with a version change of the container. That change must do a build & test of the repository content using the updated container. If that builds - all good, right? If not --> PR fails, investigation required. No blocking of development or surprises at any point in time.

docs/design_decisions/DR-001-infra-extension.md
lurtz: I have to be open, that I do not like all of bazels concepts. It is amazing for tracking dependencies, caching and its best effort sandboxing when building and running tests. But I would rather not let it download itself, tools or toolchains and store these within the devcontainer image. However this is against the point you made about having the devcontainer optional.

That being said you also stated that the entire S-CORE should use a single bazel version: https://github.com/eclipse-score/score/blob/0c6a3b62f582dd1b1e0bda912632391092e06601/docs/design_decisions/DR-001-infra-extension.md?plain=1#L165

To me the best solution would be to include exactly **THIS** bazel version in the devcontainer instead of using bazelisk.

Maybe we should discuss how much infrastructure (bazel mirror, bazel archive, devcontainer registry) we want to build, or if we want to create a design which needs less infrastructure. I lean towards a solution, which fulfills all requirements, but needs as less infrastructure in the background as possible.

docs/design_decisions/DR-001-infra-extension.md
lurtz: > otherwise changes to the container silently change builds

This can only happen if no fixed revisions are used. E.g. `ghcr.io/eclipse-score/devcontainer:latest` will silently change, but `ghcr.io/eclipse-score/devcontainer:v1.1.0` will always stay the same.

docs/design_decisions/DR-001-infra-extension.md
dcalavrezo-qorix: that's fair, but  at the moment all the repos have their own `.bazelversion` file, or?  example : https://github.com/eclipse-score/communication/blob/main/.bazelversion

Wouldn't that override whatever we set in the `devcontainer` ?

We'd have to set-up  an enforcing mechanism? or is there such already?

lurtz: IMHO there are multiple ways to achieve this inside the devcontainer.

- You install a specific bazel version in the devcontainer and set an environment variable, which overrides any `.bazelversion` file. However this happens silently
- We could also not ship bazelisk in the devcontainer, then the build will fail, if the `.bazelversion` does not match the version of the preinstalled bazel binary.

docs/design_decisions/DR-001-infra-extension.md
AlexanderLanin: Ignore caching for now ;-)

We have e.g. this for pinning tool versions: https://github.com/eclipse-score/devcontainer/blob/main/src/s-core-devcontainer/.devcontainer/s-core-local/versions.yaml

And such files for pinning python packages: https://github.com/eclipse-score/docs-as-code/blob/main/src/requirements.txt

FScholPer:
@dcalavrezo-qorix whats the state here?

FScholPer:
@dcalavrezo-qorix last call for action! I will close it next week if there is no feedback if we now can merge or not

MaximilianSoerenPollak:
> @dcalavrezo-qorix last call for action! I will close it next week if there is no feedback if we now can merge or not

Please do not close it yet. 
I have asked for feedback again in the Infra round and Dan is also preparing a PoC. 
I would propose to give it until next week, and if then no satisfactory conclusion or at least path forward is reached we take action?

dcalavrezo-qorix:
Let’s resume the discussion and look at a few remaining gaps around traceability and long-term reproducibility, if that is ok with you.

For ease of discussion I picked persistency as an example, but the same observations likely apply to most S-CORE repositories.

I would like to focus specifically on build reproducibility over a longer time horizon (10+ years).

1. MODULE.bazel.lock is not versioned

We currently do not generate and commit MODULE.bazel.lock.

Even if `bazel_dep` versions are fixed, dependency resolution can still drift due to:

- transitive selection changes
- registry metadata updates
- yanked releases


Without committing the lock file, we are not freezing the fully resolved module graph. From a reproducibility and auditability standpoint, this seems like a gap.

2. Use of overrides in MODULE.bazel

We still rely on overrides (for ex. git_override).
Even when pinned to a commit SHA, this assumes:

- long-term availability of the upstream repository
- no forced rewrite

For long-term reproducibility, overrides pointing to external GitHub repos may not be sufficient unless we define a mirroring or archival strategy.

3. External toolchain blobs should be mirrored

We fetch external toolchain artifacts (e.g. SDK archives, toolchain tarballs) from vendor URLs.

Even if we verify SHA256, long-term availability of those URLs is not guaranteed.
For true reproducibility, we should clarify whether:

all such artifacts must be mirrored into an S-CORE controlled artifact store, or we accept vendor URLs as a long-term dependency.

This should probably be made explicit in the design decision.

4. Devcontainer image tag is not immutable

The devcontainer currently references an image tag (e.g. v1.1.0).
Container tags can be re-pushed.

If we rely on devcontainer as part of the reproducible developer environment, we may want to:

pin by digest (@sha256:…), or

define a policy that tagged images are immutable and archived.

will add more stuff

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,244 @@
...
+https://www.apache.org/licenses/LICENSE-2.0
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+
+# DR-001-B-Infra-Extension: S-CORE Build Execution Contract
+
+
+* **Date:** 2026-01-20
```
MaximilianSoerenPollak, line 18: Should this not be changed to the latest edit?

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -132,11 +136,18 @@ This layer defines the **default user-space environment** in which builds are ex
...
-  - referenced by immutable digest
-  - archived for long-term reproducibility
+  - built from a **defined Ubuntu LTS baseline**
+  - compatible with common developer tooling (e.g. by following the [specification](https://containers.dev/))
+  - referenced by an **immutable image digest**
+  - archived for **long-term reproducibility**
+
+#### Baseline Preservation and Reproducibility
+- Once a devcontainer image is used in CI, its image digest becomes part of the build provenance
+- All such images must be archived and retrievable for **at least 10 years**
```
AlexanderLanin: Let's not define/mandate storage durations in public score. We cannot guarantee that. GitHub could go offline tomorrow. And amount of years probably varies for users.

maybe just add "designed to"?

All such images must be designed to be archived and retrievable for **at least 10 years**

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -159,7 +170,7 @@ This layer defines the **default user-space environment** in which builds are ex
 This layer defines **what Bazel controls and guarantees**.
 
 #### Bazel Versioning
-- Each repository must contain `.bazelversion`.
+- Each relevant repository must contain `.bazelversion`.
```
AlexanderLanin: not sure about that. certainly until we finally have that aligned.

generally speaking we dont write tool versions into each repo, and bazel is yet another tool.

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,244 @@
...
+
+
+# DR-001-B-Infra-Extension: S-CORE Build Execution Contract
+
+
+* **Date:** 2026-01-20
+
+```{dec_rec} S-CORE Build Execution Contract
+:id: dec_rec__infra__execution_contract
+:status: accepted
```
FScholPer, line 22: @dcalavrezo-qorix Can we just change to in progress and merge a first version?

dcalavrezo-qorix: I plan another round of reviews

Gottes Mühlen mahlen langsam

FScholPer:
I think DR would not be enough. We also need some kind of Stakeholder requirements -> tool_Req...

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,284 @@
...
+
+This program and the accompanying materials are made available under the
+terms of the Apache License Version 2.0 which is available at
+https://www.apache.org/licenses/LICENSE-2.0
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+
+# DR-001-B-Infra-Extension: S-CORE Build Execution Contract
```
a-zw, line 15: Why 001-B? Could just pick a higher number like 009.

antonkri (approved):

docs/design_decisions/DR-001-infra-extension.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Purpose
+
+This document defines the execution contract for S-CORE builds across developer machines
+and CI infrastructure.
+Its goal is to ensure **long-term reproducibility (≥10 years)**, **traceability** and
+**practical hermeticity**, despite changes in underlying infrastructure such as
+GitHub-hosted runners.
+
+It builds on [DR-001], which was concerned about the same topics, but was focused on tools only. This adds details where the original description was too fuzzy.
```
lurtz, line 38: was the intention to have link to DR-001 in the generated documentation? At the moment this is not the case

lurtz (approved):

MaximilianSoerenPollak (approved):

PR #2499, opened 2026-01-26
add design decision for break cyclic dependency between docs-as-code and process repository 

RolandJentschETAS:
Add design decision for break cyclic repository between docs-as-code and process repository

Rendered: https://eclipse-score.github.io/score/pr-2499/design_decisions/DR-007-infra.html

### DIFF
--- a/docs/design_decisions/DR-007-infra.rst
+++ b/docs/design_decisions/DR-007-infra.rst
@@ -41,6 +41,7 @@ This tight coupling makes maintenance and evolution of both repositories difficu
    left to right direction
    database process_description {
       artifact examples
+      artifact templates
    }
    database docs_as_code {
       artifact metamodel.yaml as yaml
@@ -48,6 +49,14 @@ This tight coupling makes maintenance and evolution of both repositories difficu
 
    process_description --> docs_as_code : defines metamodel
    docs_as_code --> process_description : checks metamodel
+
+The artifacts within those repos are:
+
+* "examples" are exemplary instances of the metamodel like :need:`feat__example_feature`.
+* "templates" (more precisely `folder templates <https://eclipse-score.github.io/process_description/main/folder_templates/index.html>`_)
+  are instances which can be copied when creating new modules
+  like :need:`doc__feature_name_architecture`.
+* "metamodel.yaml" is `this file <https://github.com/eclipse-score/docs-as-code/blob/v2.3.3/src/extensions/score_metamodel/metamodel.yaml>`__.
 
 This means to roll out a change to the process looks like this:
 
@@ -113,6 +122,7 @@ This eliminates the cyclic dependency by having a single source of truth for bot
    database "process_description\n+ docs_as_code" {
       artifact metamodel.yaml as yaml
       artifact examples
+      artifact templates
    }
 
 Effort 😡😡: Disruptive effort to merge repos.
@@ -128,7 +138,7 @@ Maintainability 💚: Good because everything is in one place.
 Option 2: Move meta model definition to process repository
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Define and maintain the `metamodel.yaml <https://github.com/eclipse-score/docs-as-code/blob/v2.3.3/src/extensions/score_metamodel/metamodel.yaml>`_
+Define and maintain the ``metamodel.yaml``
 solely in the process repository.
 The docs-as-code repository would then only provide the infrastructure for the meta model, not define or modify it.
 The process repository would be the authoritative source for the meta model.
@@ -141,6 +151,7 @@ Also tests (scripts) and examples would be maintained there.
    left to right direction
    database process_description {
       artifact examples
+      artifact templates
       artifact metamodel.yaml as yaml
    }
    database docs_as_code {
@@ -148,7 +159,7 @@ Also tests (scripts) and examples would be maintained there.
 
    process_description --> docs_as_code : defines metamodel
    yaml --> docs_as_code : as input
-   docs_as_code --> examples : checks metamodel
+   docs_as_code --> process_description : checks metamodel
 
 Implication:
 If the docs-as-code module would select the metamodel yaml version on its own,
@@ -169,25 +180,26 @@ Maintainability 💚: Good because of clear ownership.
 
 Clear Ownership 💚: Cleanly separated.
 
-Option 3: Move examples (Sphinx-Needs objects) to a third repository
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Option 3: Move examples to docs_as_code and templates to module_template
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Extract all example Sphinx-Needs objects (currently imported into docs-as-code) into ``module_template``.
-The process repository is independent and docs-as-code repositories would depend on this third repository for examples, breaking the cyclic dependency.
-But this adds complexity with an additional repository to maintain and still contains cyclic dependencies between all three repositories.
+Move all examples from ``process_description`` into ``docs_as_code`` repo
+and all templates from ``process_description`` into ``module_template`` repo.
+No relevant cyclic dependency remains.
 
 .. uml::
    :align: center
-   :caption: Move examples (Sphinx-Needs objects) to a third repository
+   :caption: Move examples to docs_as_code and templates to module_template
 
    left to right direction
    database process_description {
    }
    database docs_as_code {
       artifact metamodel.yaml as yaml
+      artifact examples
    }
    database module_template {
-      artifact examples
+      artifact templates
    }
 
    process_description --> docs_as_code : defines metamodel
@@ -222,10 +234,11 @@ This breaks the cycle by introducing a clear hierarchical dependency structure.
    database module_template {
       artifact metamodel.yaml as yaml
       artifact examples
+      artifact templates
    }
 
    process_description --> docs_as_code : defines metamodel
-   docs_as_code --> examples : checks metamodel
+   docs_as_code --> module_template : checks metamodel
    yaml --> docs_as_code : as input
 
 Effort 😡: High effort for the configurable ``metamodel.yaml``.
@@ -238,8 +251,9 @@ Maintainability 😡: More repos to maintain.
 
 Clear Ownership 💚: Clearly, the Process community takes authority over ``metamodel.yaml``.
 
-Option 5: Move examples to docs-as-code repository
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Option 5: Move examples and templates to docs-as-code repository
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 Move all example Sphinx-Needs objects from the ``process_description`` repository to the ``docs_as_code`` repository.
 The process repository would define requirements for the meta model, while docs-as-code would provide infrastructure, the meta model and host the examples that demonstrate the meta model.
 This breaks the cycle by removing the import dependency from docs-as-code back to the process repository.
@@ -253,6 +267,7 @@ This breaks the cycle by removing the import dependency from docs-as-code back t
    }
    database docs_as_code {
       artifact examples
+      artifact templates
       artifact metamodel.yaml as yaml
    }
 
@@ -266,7 +281,7 @@ UX 💚: Fine.
 
 Maintainability 💚: Good because of clear ownership.
 
-Clear Ownership 😡: Process community wants control of examples but they are in the ``docs_as_code`` repo.
+Clear Ownership 😡: Process community wants control of the templates but they are in the ``docs_as_code`` repo.
 
 Option 6: Change error handling from warnings as errors to warnings only
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,113 @@
...
+
+Currently, there are two repositories defining the docs-as-code principles with Sphinx and Sphinx-Needs:
+
+- The **process description repository** defines the requirements and the meta model for the Sphinx-Needs objects (requirements, architecture, processes, etc.). It includes also example Sphinx-Needs objects that illustrate the defined meta model.
+- The **docs-as-code repository** provides the base docs as code infrastructure and uses the process requirements to define the sphinx needs meta model and includes Sphinx-Needs objects from the process repository for testing.
+
+However, the process repository also uses the docs-as-code repository's as infrastructure for the rocess documentation and the meta model for Sphinx-Needs objects, creating a **cyclic dependency** between the two repositories. Any change in the process requirements (in the process repo) leads to a change in the docs-as-code meta model, but any change in the meta model can cause build errors in the docs-as-code repo, since it imports Sphinx-Needs objects from the process repo as examples. This tight coupling makes maintenance and evolution of both repositories difficult and error-prone.
```
a-zw:
```suggestion
However, the process repository also uses the docs-as-code repository's as infrastructure for the rocess documentation and the meta model for Sphinx-Needs objects, creating a **cyclic dependency** between the two repositories. Any change in the process requirements (in the process repo) leads to a change in the docs-as-code meta model, but any change in the docs-as-code meta model can cause build errors in the process_description repo and this happens during the docs-as-code build as Sphinx-Needs objects from the process repo are imported. This tight coupling makes maintenance and evolution of both repositories difficult and error-prone.
```

Also, I would suggest "line break after each sentence" because in a PR we can then comment on each sentence separately.

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,113 @@
...
+
+.. image:: img/DR-007-issue.drawio.svg
+   :alt: Cyclic dependency between process and docs-as-code repositories
+   :align: center
+
+Goals and Requirements
+^^^^^^^^^^^^^^^^^^^^^^
+
+- Avoid cyclic dependencies between repositories.
```
a-zw: This is a nice principle but not really a goal to pursue here.

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,113 @@
...
+- Enable independent evolution of process requirements and meta model definitions.
```
a-zw: This is already possible. Changing the process requirements does not implicitly change the metamodel implemented in docs-as-code.

Maybe you meant:

```suggestion
- Enable independent evolution of templates in process description and docs-as-code meta model definitions.
```

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,113 @@
...
+Combine the process description and docs-as-code repositories into a single repository. This eliminates the cyclic dependency by having a single source of truth for both the meta model and the Sphinx-Needs objects/examples.
+
+.. image:: img/DR-007-issue_option_1.drawio.svg
+   :alt: Merged repository eliminating cyclic dependency
+   :align: center
+
+
+Option 2: Move meta model definition to process repository
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Define and maintain the Sphinx-Needs meta model solely in the process repository. The docs-as-code repository would then only provide the infrastructure for the meta model, not define or modify it. The process repository would be the authoritative source for the meta model. Also tests and examples would be maintained there.
```
a-zw: By "Sphinx-Needs meta model" we mostly think of the [metamodel.yaml](https://github.com/eclipse-score/docs-as-code/blob/main/src/extensions/score_metamodel/metamodel.yaml) file. That would indeed be a nice abstraction but it is not the complete implementation. Some checks are implemented in Python. Example: https://github.com/eclipse-score/docs-as-code/blob/main/src/extensions/score_metamodel/checks/check_options.py#L150

So I assume this would mean to move the "score_metamodel" extension to process_description repository. That certainly would not be "low effort".

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,113 @@
...
+   :widths: 20, 10, 10, 10, 10
+
+   Effort, 😡, 💚, 😡, 😡
+   UX, 💚, 💚, 😐, 😐
+   Bureaucracy, 😐, 💚, 😡, 😡
+   Speed, 😡, 💚, 😡, 😡
+   Flexibility, 💚, 💚, 😐, 💚
+   Independence, 😐, 💚, 💚, 💚
+   Maintainability, 💚, 💚, 😐, 💚
+   Scalability, 😐, 😐, 😡, 💚
```
a-zw: I feel like option 1 and 2 are actually very similar, so I don't understand how this tables evaluates them differently with respect to Bureaucracy, Speed, and Independence.

docs/design_decisions/img/DR-007-issue_option_3.drawio.svg
a-zw, line 1: Is "examples" and "templates" synonymous? If yes, we should only use one the terms. If no, define them properly.

AlexanderLanin:
* As discussed please add option 5: move examples to docs-as-code. Maybe it doesn't make sense, maybe it does. It should be evaluated.
* Please add option 0 to the evaluation table: no change
* Please add explanations to each line in the evaluation table, explaining each of those keywords.

docs/design_decisions/img/DR-007-issue_option_3.drawio.svg
RolandJentschETAS: templates is a repository

docs/design_decisions/DR-007-infra.rst
RolandJentschETAS: Avoiding cyclic dependencies is the reason for the current design decision. It makes it currently complicated to update the meta model. So it is really the goal.

docs/design_decisions/DR-007-infra.rst
RolandJentschETAS: As described above, the docs as code repo imports the sphinx needs elements of the examples while defining the meta model for them. So any change of the meta model from docs as code breaks the build of docs-as-code repo. Therefore the meta model could not be changed. Vice versa the process repo examples can't be changed, as long as the meta model is not changed. Because of this cyclic dependency changing the meta model is complicated and only possible in a multi step appoach.

docs/design_decisions/DR-007-infra.rst
RolandJentschETAS: changed

docs/design_decisions/DR-007-infra.rst
RolandJentschETAS: I have added a description for the options as well as for the evaluation. And also rework the evaluation with my best guesses. Hope thats fine now.

docs/design_decisions/DR-007-infra.rst
AlexanderLanin: * 1 part of score_metamodel is generic - we keep that in docs-as-code
* 1 part of score_metamodel is score specific python code - that is stuck in docs-as-code, and we live with it. It's a small part.
* 1 part of score_metamodel goes to process (yml)
* 1 part of score_metamodel is quite a bunch of examples (rst files). Most of these should go to process. But I feel this is already details if & after we decide for option 2.

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Define and maintain the Sphinx-Needs meta model solely in the process repository. The docs-as-code repository would then only provide the infrastructure for the meta model, not define or modify it. The process repository would be the authoritative source for the meta model. Also tests and examples would be maintained there.
+
+.. image:: img/DR-007-issue_option_2.drawio.svg
+   :alt: Meta model defined in process repository
+   :align: center
+
+Option 3: Move examples (Sphinx-Needs objects) to a third repository
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Extract all example Sphinx-Needs objects (currently imported into docs-as-code) into a separate third repository. The process repository will become independent and docs-as-code repositories would depend on this third repository for examples, breaking the cyclic dependency. But this adds complexity with an additional repository to maintain and still contains cyclic dependencies between all three repositories.
```
AlexanderLanin: That's confusing. You are talking about examples in process?
Cyclic dependency is described in problem statement.
(You could also repeat it, but "currently imported into docs-as-code" is confusing as-is)

```suggestion
Extract all example Sphinx-Needs objects (currently in process_description) into a separate third repository. The process repository will become independent and docs-as-code repositories would depend on this third repository for examples, breaking the cyclic dependency. But this adds complexity with an additional repository to maintain and still contains cyclic dependencies between all three repositories.
```

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+
+.. image:: img/DR-007-issue_option_3.drawio.svg
+   :alt: Examples moved to a third repository
+   :align: center
+
+Option 4: Move meta model and examples into a separate repository
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Create or use a dedicated meta model repository that contains only the Sphinx-Needs meta model definitions and the examples. Both the process repository and docs-as-code repository would depend on this meta model repository (if necessary), making it the single source of truth. This breaks the cycle by introducing a clear hierarchical dependency structure.
```
AlexanderLanin: From an architecture point of view this is rather nice, so lets mention it.

```suggestion
Create or use a dedicated meta model repository that contains only the Sphinx-Needs meta model definitions and the examples. Both the process repository and docs-as-code repository would depend on this meta model repository (if necessary), making it the single source of truth. This breaks the cycle by introducing a clear hierarchical dependency structure.
Process repository becomes more docs-as-code independent. And docs-as-code becomes more score-independent.
```

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+- **Flexibility**: Ability to adapt and evolve the solution in respect to requirements of meta model and examples over time.
+- **Independence**: Degree of decoupling between repositories and autonomous evolution in respect to the meta model and examples.
+- **Maintainability**: Long-term ease of maintaining and updating the solution.
+- **Scalability**: Ability to handle growth in content, contributors, and complexity.
+
+.. csv-table::
+   :header: Criteria, Option 0, Option 1, Option 2, Option 3, Option 4, Option 5, Option 6
+   :widths: 20, 10, 10, 10, 10, 10, 10, 10
+
+   Effort,Speed     💚, 😡, 😐, 😐, 😡, 💚, 💚
```
AlexanderLanin: I think effort for 2 and 4 are rather identical: 😡
Effort for 3 and 5 should be identical? I'm guessing 😐? Repo setup is <1h.

```suggestion
   Effort     💚, 😐, 😡, 😐, 😡, 😐, 💚
```

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+   Speed,           💚, 😐, 😐, 😡, 😡, 💚, 💚
```
AlexanderLanin: I have no idea how to review Speed? What's the difference with effort? Overall rollout? But isn't that handled in effort already?

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+   UX,              😡, 💚, 💚, 😡, 💚, 💚, 😐
```
AlexanderLanin: UX includes keeping requirements, metamodel and examples in sync? Regardless of administrative overhead, this is complex across repositories.

```suggestion
   UX,              😡, 💚, 💚, 😡, 😐, 😐, 😐
```

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+   Bureaucracy,     😐, 💚, 😐, 😡, 😡, 😐, 😐
```
AlexanderLanin: Oh ok, we have Bureaucracy for what I just mentioned in UX. For me personally UX and Bureaucracy are one and the same?! I don't know how to consider UX without Bureaucracy. In order to work on the metamodel / examples, you need matching requirements.

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+   Flexibility,     😡, 💚, 💚, 😐, 💚, 💚, 😐
```
AlexanderLanin: Sorry, I dont get what this is. Is that like extendibility with future docs-as-code features?

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+   Independence,    😐, 😐, 😐, 💚, 💚, 💚, 😐
```
AlexanderLanin: independence of whom with whom? Isn't independence exactly what we want to avoid?

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,143 @@
...
+   Maintainability, 😡, 😐, 💚, 😐, 💚, 💚, 😡
+   Scalability,     😡, 😡, 😡, 😡, 💚, 😡, 😡
+
+**Rationale:**
+
+- **Option 0 (No change)**: No effort required, poor UX (frequent build failures), low bureaucracy (current process), immediate (no implementation needed), poor flexibility (locked into current structure), poor independence (cyclic dependency remains), poor maintainability (ongoing coordination burden), poor scalability (problem worsens as repos grow).
```
AlexanderLanin: Having smileys above and every smiley explained here doesn't work. If these explanations add value, they must be directly in the table. It's super hard to find the explanation of each smiley here.

a-zw: The approach I use is to have the explanations with the options. This keeps the table lean.

If an explanation about the criteria themselves (independent of the options) is needed, then this should be done in the "Context / Problem" section. Maybe there should even be a 1:1 relation between the goals and the criteria.

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,208 @@
...
+   :widths: 20, 15, 15, 15, 15, 15
+
+   Effort,          💚, 😡, 💚, 💚, 💚
+   Independence,    😡, 💚, 💚, 💚, 💚
+   UX,              😡, 💚, 💚, 💚, 💚
+   Maintainability, 😡, 💚, 💚, 💚, 😡
+
+Option 0 scores poorly with respect to independence, UX, and maintainability.
+Option 1 requires high effort.
+Option 6 compromises maintainability.
```
a-zw: Option 5 looks just a good as option 2 and there is not reason documented why option 2 is selected instead.

aschemmel-tech (commented):
Like discussed in the process community, I am also fine with selection of option 2.

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,208 @@
...
+The process repository would be the authoritative source for the meta model.
+Also tests and examples would be maintained there.
+
+.. image:: img/DR-007-issue_option_2.drawio.svg
+   :alt: Meta model defined in process repository
+   :align: center
+
+Effort 💚: Low effort.
+
+Independence 💚: Good because ``docs_as_code`` just consumes.
```
a-zw: Isn't independence 😡?

The thing is that docs_as_code (as of today) decides itself which version of the process_description it uses (certainly not HEAD or some branch). So, if metamodel.yaml is in process_description and one changes it there, the docs build would still use an older version. (Or does Bazel recognize the local repo is the module reference by an external module and uses that?)

This implies we would have to change `docs()` that one can configure which metamodel.yaml is used such that process_description can use its own. In other words, this option is not that trivial as it seems and maybe we even need to declare "effort" as 😡?

Corollary: Any 😡 here means that option 5 would be the winner.

docs/design_decisions/DR-007-infra.rst
RolandJentschETAS: Added...

a-zw: Resolved. With the latest changes, option 5 is the logical conclusion.

docs/design_decisions/DR-007-infra.rst
a-zw: Resolved.

docs/design_decisions/DR-007-infra.rst
```diff
@@ -197,19 +208,23 @@ Maintainability 😡: Poor because warnings won't be fixed as quickly as errors.
...
    UX,              😡, 💚, 💚, 💚, 💚
    Maintainability, 😡, 💚, 💚, 💚, 😡
 
 Option 0 scores poorly with respect to independence, UX, and maintainability.
 Option 1 requires high effort.
+Option 2 has minor downsides for effort and independence.
 Option 6 compromises maintainability.
 
-**Decision:** Option 2 or Option 5 both provide the same benefits but Option 5 requires less effort, so we pick it.
+**Decision:** Option 5 is obviously the best choice.
```
masc2023: See here, was also presented and discussed
https://github.com/orgs/eclipse-score/discussions/407#discussioncomment-15605348
Option 2 was preferred as it would give independence in the form, that process_description could be used also with other tools, or doc-as-code could even use other models, is that reflected here?

a-zw: Independence is actually slightly better with Option 5 hence the 💛 vs 💚 difference. Option 2 provides independence but requires process_description to specify the meta model "in full detail". For contrast, with Option 5 only vague textual requirements are needed.

Both, Option 2 and 5 would allow for alternative tools (like TLRC).

a-zw:
Now option 3 wins: Move examples to module_template repository

docs/design_decisions/DR-007-infra.rst
```diff
@@ -0,0 +1,311 @@
...
+Maintainability 💚: Good because of clear ownership.
+
+Clear Ownership 💚: Cleanly separated.
+
+Option 3: Move examples (Sphinx-Needs objects) to a third repository
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Extract all example Sphinx-Needs objects (currently imported into docs-as-code) into ``module_template``.
+The process repository is independent and docs-as-code repositories would depend on this third repository for examples, breaking the cyclic dependency.
+But this adds complexity with an additional repository to maintain and still contains cyclic dependencies between all three repositories.
```
aschemmel-tech: My impression from the discussion was not that this solution "still contains cyclic dependencies between all three repositories"

MaximilianSoerenPollak:
> Now option 3 wins: Move examples to module_template repository

Move the templates to module_template right? 
The examples are inside of Docs-As-Code, or am I looking at the wrong picture?

masc2023:
> > Now option 3 wins: Move examples to module_template repository
> 
> Move the templates to module_template right? The examples are inside of Docs-As-Code, or am I looking at the wrong picture?

Yes, that was discussed, examples in doc-as-code, templates for the modules into module_templates, may high-level templates stay in process_description, but they are very stable

docs/design_decisions/DR-007-infra.rst
a-zw: Good catch. That is outdated text which I forgot to update.

masc2023 (approved):

PR #2506, opened 2026-01-27
Update quality manager election

PandaeDo:


docs/platform_management_plan/role_assignment/platform_quality_manager.rst
```diff
@@ -64,12 +66,34 @@ S-CORE Commits:
...
+- committer in `process_description repository <https://github.com/eclipse-score/process_description>`_
+- committer in `score repository <https://github.com/eclipse-score/score>`_
+
 Election
 ========
 
 For the election further Committers with similar skills are asked for approval:
 
-- `<https://github.com/PandaeDo>`_
+- `<https://github.com/masc2023>`_
```
pahmann, line 98: Do we want to handle it like this? I would expect that Markus is not added, but we have a clearer statement that existing quality manager are eligible to and "For the election further Committers with similar skills are asked for approval" only lists further committers. WDYT?

masc2023: We did the same for electing Safety Manager, but it should be changes in that way, Election should be done by already elected Quality Managers plus additional please change your request

PR #2544, opened 2026-02-04
Symbol report verification report

pawelrutkaq:


docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,120 @@
...
+   * - 1
+     - False-positive: The function is reported as covered and increase coverage percentage, although it is not covered in reality (not included in `profraw` data)
+     - | The reported covarge number is reported higher than the real coverage.
+       | This may lead to false assumption that code is sufficiently covered by tests, although it is not.
+
+     - no
+     - no
+     - yes
+     - no
+     - high
```
masc2023, line 94: Without mitigation that seems not high confidence for me

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,120 @@
...
+   * - 3
+     - The function from source code is not included in the coverage report at all.
+     - | The reported coverage number is reported higher than the real coverage.
+       | This may lead to false assumption that code is sufficiently covered by tests, although it is not.
+     - no
+     - no
+     - yes
+     - no
+     - high
```
masc2023, line 142: You did not cover bugs introduced by instrumentation, standard mitigation here would required running the test twice, once with and once w/o instrumentation

masc2023 (changes requested):

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pawelrutkaq: There is general problem: This is tool verification report and code coverage is a process. So how shall i describe it if i am specifically interested whether post procesing tooling require qualification. (here symbol report and blanket)

masc2023: The code coverage think may part of another report or you must also explain it here as part of your use cases. And if the tools help to mitigate the rustc coverage issues, e.g. detect failure by instrumentation, than it should be added here

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+     - no
+     - yes
+     - no
+     - high
+   * - 3
+     - The function from source code is not included in the coverage report at all.
+     - | The reported coverage number is reported higher than the real coverage.
+       | This may lead to false assumption that code is sufficiently covered by tests, although it is not.
+     - no
+     - Function will not be present in the report, so it will be visible that it is not covered.
```
aschemmel-tech: Please describe better how this is visible, to be able to judge if it can be easily observed. E.g. if the coverage data is displayed in a view together with the source, I would belive that this is a high detection likelyhood. If it is only a seperate file logging all functions and their coverage I would see a low detection value appropriate.

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+     - Impact safety detection sufficient?
+     - Further additional safety measure required?
+     - Confidence (automatic calculation)
+   * - 1
+     - `profraw` does not contain data, but the tool reports coverage for the function.
+     - | The reported coverage number is reported higher than the real coverage.
+       | This may lead to false assumption that code is sufficiently covered by tests, although it is not.
+
+     - no
+     - Likelihood of this malfunction is low
```
aschemmel-tech: No reason given for this statement. Where is the measure?

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+     - Impact safety measures available?
+     - Impact safety detection sufficient?
+     - Further additional safety measure required?
+     - Confidence (automatic calculation)
+   * - 1
+     - `profraw` does not contain data, but the tool reports coverage for the function.
+     - | The reported coverage number is reported higher than the real coverage.
+       | This may lead to false assumption that code is sufficiently covered by tests, although it is not.
+
+     - no
```
aschemmel-tech: safety impact must be yes - because testing is not complete.

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+       | This may lead to false assumption that code is sufficiently covered by tests, although it is not.
+     - no
+     - Function will not be present in the report, so it will be visible that it is not covered.
+     - yes
+     - no
+     - high
+   * - 4
+     - Calculations are wrong, so the reported coverage number is not correct.
+     - | The reported coverage is different than the coverage in `profraw` files.
+     - no
```
aschemmel-tech: A "real" coverage below 100% but calculated/displayed as 100% would be a safety problem.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+     - Function will not be present in the report, so it will be visible that it is not covered.
+     - yes
+     - no
+     - high
+   * - 4
+     - Calculations are wrong, so the reported coverage number is not correct.
+     - | The reported coverage is different than the coverage in `profraw` files.
+     - no
+     - no
+     - yes
```
aschemmel-tech, line 140: What detection does this refer to?

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+     - no
+     - yes
+     - no
+     - high
+
+
+Result
+~~~~~~
+Symbol report and blanket do not require qualification for use in safety-related software development according to ISO 26262.
```
aschemmel-tech: as stated by masc2023 - without additional measures I cannot agree to this result

aschemmel-tech (changes requested):
see inline comments, proposal: take over the failure modes and mitigations from https://public-docs.ferrocene.dev/main/certification/core/safety-plan/tools.html#code-coverage

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+   :tags: tool_management
+
+Symbol Report and Blanket Verification Report
+=============================================
+
+Introduction
+------------
+Scope and purpose
+~~~~~~~~~~~~~~~~~
+Symbol report and blanket are tools for post-processing of code coverage data generated by Rust compiler.
```
pahmann: If they are tools I prefer to highlight their nature e.g. using capital letters or by putting them in quotation marks or set them in italic.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+Integrated in bazel.
+
+Environment
+~~~~~~~~~~~
+Requires Rust toolchain and Bazel build environment.
+
+Safety evaluation
+-----------------
+This section outlines the safety evaluation of symbol report and blanket for its use within the S-CORE project. This evaluation assumes that the Rust compiler is
+qualified and output of coverage data in `.profraw` format is correct. Due to that, we solely focus on post processing that is done by symbol report and blanket only.
```
pahmann, line 72: Will be use and measure the coverage based on the `.profraw` somewhere else? This can have impact on the tool classification, which is the reason I am asking.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+  :width: 100%
+  :align: center
+  :alt: Symbol report and blanket overview
+
+  Symbol report and blanket overview
+
+Available information
+~~~~~~~~~~~~~~~~~~~~~
+- Version: >= 1.90.0 [1]_
+- Official repository: https://github.com/ferrocene/ferrocene/tree/main/ferrocene/tools/blanket,  https://github.com/ferrocene/ferrocene/tree/main/ferrocene/tools/symbol-report
```
pahmann, line 51: Are these two tools to be used in parallel for all cases or are they intended to work also standalone. In this case two entries may be needed as tools and they cannot be treated as a single one.

docs/score_tools/symbol_report_blanket.rst
pahmann: Rounding can also become an issue.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+This section outlines the safety evaluation of symbol report and blanket for its use within the S-CORE project. This evaluation assumes that the Rust compiler is
+qualified and output of coverage data in `.profraw` format is correct. Due to that, we solely focus on post processing that is done by symbol report and blanket only.
+
+
+.. list-table:: Safety evaluation
+   :header-rows: 1
+   :widths: 1 2 8 2 6 4 2 2
+
+   * - Malfunction identification
+     - Use case description
```
pahmann, line 80: Your use cases are actually malfunctions and not use cases. See https://eclipse-score.github.io/score/pr-2544/score_tools/doc_as_code.html#doc_tool__doc_as_code as an example of the use case and the resulting malfunction.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,130 @@
...
+-----------------
+This section outlines the safety evaluation of symbol report and blanket for its use within the S-CORE project. This evaluation assumes that the Rust compiler is
+qualified and output of coverage data in `.profraw` format is correct. Due to that, we solely focus on post processing that is done by symbol report and blanket only.
+
+
+.. list-table:: Safety evaluation
+   :header-rows: 1
+   :widths: 1 2 8 2 6 4 2 2
+
+   * - Malfunction identification
```
pahmann, line 79: Can there be anything as a case that:
- The coverage is reported for another function.
- There is a version mismatch that the tool takes the data from a previous or other run.
- Could there be an issue for dynamic configuration like calling the unit test with configuration a but have configuration b mapped in code for the report?
 
I assume the use cases are limited which you want to do with the report, but the 4 malfunctions which you call use cases are incomplete.

pahmann (changes requested):

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pawelrutkaq: they work together.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pawelrutkaq: 1. added
2. The tools takes what is given by bazel as in picture. so its not tool reposiblity to handle that. But in any case of this, it would not casue reporting more coverage then it really is (realtivelly)
3. maybe, but that would make a report one big `scrambled eggs` ;)

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pawelrutkaq: I reworded but kep same layout. I see the layout of table is different acros tools so...

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pawelrutkaq: No, these tools are the only measurement for coverage of rust

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pawelrutkaq: Reworked table

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pawelrutkaq: Reworked table

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: Reworked table

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: Reworked table

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: Reworked table

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: Reworked table

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pawelrutkaq: Reworked table

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: Reworked table

pawelrutkaq:
@pahmann @aschemmel-tech @masc2023 tried to resolve comments please recheck

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
masc2023: Layout should be overall same, using https://eclipse-score.github.io/process_description//main/folder_templates/tools/tool_verification_report_template.html

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,154 @@
...
+     - Impact safety measures available?
+     - Impact safety detection sufficient?
+     - Further additional safety measure required?
+     - Confidence (automatic calculation)
+   * - 1
+     - False-positive: A function is reported as covered, although it is not covered
+     - Overreporting, could result in testing gap.
+     - yes
+     - | Likelihood of such an error low due to wide usage of the tool (many S-CORE modules and other projects like ferrocene)
+       | Additionally, every new tool release is tested by running tests in prepared integration testsuite to detect such errors. (PROPOSAL POINT)
```
PandaeDo: I would also add the expectation of the tester/user. So a failure can be detected.

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,154 @@
...
+     - yes
+     - no
+     - high
+   * - 2
+     - False-negative: A function is reported as not covered, although it is covered
+     - Underreporting, will not result in testing gap.
+     - yes
+     - Since we want to achieve 90%+ branch coverage this would stand out and be manually investigated.
+     - yes
```
PandaeDo:
```suggestion
     - no
```

When I understand it correctly this will not have a impact on safety.

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,154 @@
...
+     - no
+     - high
+   * - 3
+     - Overcounting: Total number of functions is too low
+     -  A function is not being considered, although it is part of the certified subset
+     - yes
+     - | `symbol-report` is developed to use exactly the same information as the compiler
+       | Additionally, every new tool release is tested by running tests in prepared integration testsuite to detect such errors. (PROPOSAL POINT)
```
PandaeDo: Also in this case the user expectation will lead in some cases to a failure detection.

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,154 @@
...
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+.. doc_tool:: symbol report and blanket
+   :id: doc_tool__symbol_report_blanket
+   :status: evaluated
+   :version: 1.90.0 (see [1])
+   :tcl: HIGH
+   :safety_affected: YES
+   :security_affected: NO
```
masc2023: Why NO, security requires also coverage

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,154 @@
...
+- Version: >= 1.90.0 [1]_
+- Official repository: https://github.com/ferrocene/ferrocene/tree/main/ferrocene/tools/blanket,  https://github.com/ferrocene/ferrocene/tree/main/ferrocene/tools/symbol-report
+- Additional information for usage in other safety projects: https://public-docs.ferrocene.dev/main/certification/core/safety-plan/tools.html#code-coverage
+
+
+Installation and integration
+----------------------------
+Installation
+~~~~~~~~~~~~
+| To add the Code coverage to your project or module follow guidelines in WIP
```
masc2023: Is there a link to WIP available, or nothing at the moment, then add a issue to follow up with that and link it here

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Integration
+~~~~~~~~~~~
+Integrated in bazel.
+
+Environment
+~~~~~~~~~~~
+Requires Rust toolchain and Bazel build environment.
+
+Safety evaluation
```
masc2023, line 69: Compare template report and other examples, add missing Security Evaluation, you can keep it blank for now

docs/score_tools/symbol_report_blanket.rst
masc2023: What means Proposal Point? Additional measure to reduce the risk? Then add issue, link it here, that we can follow up as it then supports the argumentation why the rating is high
If Proposal Point is additional measure, update the table with additional measure needed yes

docs/score_tools/symbol_report_blanket.rst
masc2023: are the 90% somewhere document, then link to it?

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
masc2023: ok, now you removed that part of the process, may then good idea to create another report for the missing part of the process then already

masc2023 (changes requested):

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: Means that this can be an mitigation if Safety collegues things its good to have/needed.

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: I dont know, i expect even 100%

docs/score_tools/symbol_report_blanket.rst
pahmann: If we have this as Proposal point this will be a measure on the low confidence. I would say, that we cannot simply say that the detection is sufficient just because we plan to write a test for this in the future.

docs/score_tools/symbol_report_blanket.rst
```diff
@@ -0,0 +1,154 @@
...
+     - Impact on safety?
+     - Impact safety measures available?
+     - Impact safety detection sufficient?
+     - Further additional safety measure required?
+     - Confidence (automatic calculation)
+   * - 1
+     - False-positive: A function is reported as covered, although it is not covered
+     - Overreporting, could result in testing gap.
+     - yes
+     - | Likelihood of such an error low due to wide usage of the tool (many S-CORE modules and other projects like ferrocene)
```
pahmann: The argument is very weak. It is a proven in use argumentation under the assumption that the use case is identical and the usage as well. Already in the case that there are tool configuration options this weakens your argument. Otherwise we could simply say. "Oh there are no issues in Linux and the likelihood is low, because it is used in so many devices".

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
pahmann: If there is no check by other tools and the tool is used to argue safety, plus manual checking is not feasible to guarantee operation is 100% the tool confidence level should decrease.

I would like to see more about the available testing and quality measures used to operate the tool to have improved argument why the likelihood of errors are reduced due to the process and quality rigor applied to the tool.

docs/score_tools/symbol_report_blanket.rst
pahmann: Here can be come linking: https://eclipse-score.github.io/score/main/platform_management_plan/software_verification.html#quality-criteria which requires update anyway as we have an audit finding on it still open.

pahmann (commented):
Compared to the first draft the new version improved. Still it feels not mature and more an early state of evaluation. I do not have the confidence that tool qualification is not needed based on the available information. 

Remark: Any information regarding the TCL from the source at Ferroscene? They have a bunch of public information on safety as they claim.

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: So we need to evaluate this report assuming future. This is required for feroceen on gives us estimates of price we need to put into this. So we shall agree that ie this point is neded and sufficient and we will request it from them.

PandaeDo (commented):
@alekseyborisyukvalidas Could you please support Pawel with the verification report?

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: I splited this as not enought and added additional measure column info.

docs/score_tools/symbol_report_blanket.rst
pawelrutkaq: done

docs/score_tools/symbol_report_blanket.rst
aschemmel-tech: line disappeared

docs/score_tools/symbol_report_blanket.rst
aschemmel-tech: line disappeared

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,170 @@
...
+     - yes
+     - Since we want to achieve 100% branch coverage (`check here <https://eclipse-score.github.io/score/main/platform_management_plan/software_verification.html#quality-criteria>`_) this would stand out and be manually investigated.
+     - no
+     - no
+     - high
+   * - 3
+     - Overcounting: Total number of functions is too low
+     -  A function is not being considered, although it is part of the certified subset
+     - yes
+     - `symbol-report` is developed to use exactly the same information as the compiler
```
masc2023, line 107: May you need to adapt here like in the cells above, **NO**

docs/score_tools/symbol_report_blanket.rst
aschemmel-tech: ok

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,170 @@
...
+     - no
+     - no
+     - yes
+     - no
+     - high
+   * - 5
+     - Line that can be executed not being reported as executable
+     - Underreporting, code that should be tested may not being tested
+     - yes
+     - `blanket` warns if a function has no executable line
```
masc2023, line 123: May you need to adapt here like in the cells above, **NO**, because you argue additional measure need

docs/score_tools/symbol_report_blanket.rst
aschemmel-tech: ok

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,170 @@
...
+     - yes
+     - **Yes**. Additionally, every new tool release is tested by running tests in prepared integration testsuite to detect such errors.
```
masc2023, line 125: may in addition you add a comment note, which explains in some sentences, what this integration testsuite does, because you mentioned in in all argumentations

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
aschemmel-tech: ok

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,170 @@
...
+     - False-negative: A function is reported as not covered, although it is covered
+     - Underreporting, will not result in testing gap.
+     - yes
+     - Since we want to achieve 100% branch coverage (`check here <https://eclipse-score.github.io/score/main/platform_management_plan/software_verification.html#quality-criteria>`_) this would stand out and be manually investigated.
+     - no
+     - no
+     - high
+   * - 3
+     - Overcounting: Total number of functions is too low
+     -  A function is not being considered, although it is part of the certified subset
```
aschemmel-tech, line 105: Is this a use case for S-CORE at all? What would we need the "Total number of functions" for? The relevant failure for S-CORE is described in Malfunction number 1 already, could be removed in my opinion.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,170 @@
...
+     - yes
+     - `symbol-report` is developed to use exactly the same information as the compiler
+     - yes
+     - **Yes**. Every new tool release is tested by running tests in prepared integration testsuite to detect such errors.
+     - high
+   * - 4
+     - Undercounting: Total number of functions is too high
```
aschemmel-tech, line 112: Is this a use case for S-CORE at all? What would we need the "Total number of functions" for? The relevant failure for S-CORE is described in Malfunction number 2 already, could be removed in my opinion.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
aschemmel-tech, line 107: `symbol-report` is mentioned as a mitigation measure which checks if the compiler has an error - unclear what compiler error this refers to.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
aschemmel-tech: Without this explanation it is not possible to judge if it is sufficient. It is also not clear who will develop and run this "integration testsuite". Ferrocene states for this Malfunction "(Future work) End-to-end test that ensures the correct lines are being reported as executable" which could be interpreted that they would implement/run such test.

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
```diff
@@ -0,0 +1,170 @@
...
+~~~~~~~~~~~
+Integrated in bazel.
+
+Environment
+~~~~~~~~~~~
+Requires Rust toolchain and Bazel build environment.
+
+Safety evaluation
+-----------------
+This section outlines the safety evaluation of `symbol report` and `blanket` for its use within the S-CORE project. This evaluation assumes that the Rust compiler is
```
aschemmel-tech, line 71: I see that this is intended to check if we need to have toolqualification of the tools `symbol report` and `blanket` developed by Ferrocene. But actually we would need an evaluation of the full toolchain including the -Cinstrument-coverage function and the llvm-profparser. The argumentation of Ferrocene that these are " widely used" is not enough to consider these "safe enough" - So maybe the "integration testsuite" should also cover these tools.

aschemmel-tech (commented):
see inline comments

docs/score_tools/tools_static_analysis_code_quality/symbol_report_blanket.rst
masc2023: So, finally, if we come not to an common agreement, we would need to have tool qualification for the complete chain, as we request it here according,
https://eclipse-score.github.io/process_description/main/process_areas/tool_management/guidance/tool_management_guideline.html#gd_guidl__tool_qualification
The tool qualification shall be based on the method validation of the software tool.

pawelrutkaq: ok, but @masc2023  @aschemmel-tech the compiler is qualified. So this is checked or ?

masc2023: > ok, but @masc2023 @aschemmel-tech the compiler is qualified. So this is checked or ?

The compiler is not part of this report, shall be an own report, result should be to be qualified. Qualified by the way does not mean in general error-free, only reducing the risk

aschemmel-tech: @pawelrutkaq : That a tool is qualified does not mean that every use case is qualified. Similar also for SW components which may provide ASIL and QM functions. For example I can not find the structural coverage use case here: https://public-docs.ferrocene.dev/main/qualification/plan/details.html#ferrocene-details

skade: > I see that this is intended to check if we need to have toolqualification of the tools `symbol report` and `blanket` developed by Ferrocene. But actually we would need an evaluation of the full toolchain including the -Cinstrument-coverage function and the llvm-profparser. The argumentation of Ferrocene that these are " widely used" is not enough to consider these "safe enough" - So maybe the "integration testsuite" should also cover these tools.

Just to be clear: our safety argument is not "widely used", but we prefer basing on pre-existing software that has a track record. Our argument is that blanket and symbol-report provide a combination that:

a) cross-checks each other
b) structurally removes overcounts
c) test failures introduced through instrumentation (or even worse, instrumentation hiding test failures) are ruled out by mandating the test suite to run with and without instrumentation and matching

and thus we land in TCL1. (This is not uncommon)

I do agree that there's more documentation and argumentation needed for usage in a safety context outside of ferrocene, however, we stand by the TCL 1 assessment.

I can see where usage of the tools _outside_ of ferrocene (in which the input and the output are very well known) may increase the assessment to TCL 2.

However, given that the current tools are at the state for the ferrocene 26.02 release (this week), this is not the documented use case right now. For that reason, my proposed mode of working would be:

* Refining the argumentation outside of Ferrocene
* Doing a TCL 2 qualification if one is deemed necessary

pawelrutkaq: @skade @aschemmel-tech maybe it would be easier to do breakout sesssions wityh safety expert from ferroceen to try to come up with way forward?

aschemmel-tech: Discussed this in the process community. @alekseyborisyukvalidas will reach out to ferroceene for a meeting on this.

pawelrutkaq: OK, let me know if You need help, I am in close contact with @skade so it may be easier for Ferroceen to setup this by someone who they already know (maybe @alekseyborisyukvalidas  has also contact so fine for me) ;)

PR #2551, opened 2026-02-05
Move infrastructure and integration topics to own folder

RolandJentschETAS:
Currently infrastructure and integration topics are in the feature folder. But that are technical topics, which does not come into the target. So that are no target features. They belongs to the platform as technical base, but the current mixture is confusing. Therefore it is moved to more fitting folder.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -60,9 +60,9 @@ Software artifacts
 
    .. grid-item-card::
 
-      Features
+      Features & Modules
       ^^^
-      :ref:`Features <features>` are the heart of the S-CORE software.
+      :ref:`Features <features>` and :ref:`Modules <modules>` are the heart of the S-CORE software.
 
    .. grid-item-card::
 


RolandJentschETAS:
> Are these information not anyhow obsolete and can be deleted?

No, this shall be moved in far future into another repository. But not deleted.

masc2023:
> > Are these information not anyhow obsolete and can be deleted?
> 
> No, this shall be moved in far future into another repository. But not deleted.

Needs discussion, the content is in my opinion outdated, and not relevant any more

anmittag:
Maybe we have to rethink the organization of the documentation as well as the repo structure (not all in score main repos)

a) For whom is the documentation: inside developer or outside user?
b) covering/following somehow the building blocks
c) focus on release - top down?
d) following the "v-model" ?

AlexanderLanin:
Let's remove this stuff. It's not worth the discussion.

@anmittag yes... to me it feels "naturally grown". And lets address that; in parallel. It should not block this PR.

masc2023 (approved):

AlexanderLanin:
@FScholPer please use squash when merging PRs with many useless commits by the same author (which is the normal case).

Otherwise git log becomes really hard to read: https://github.com/eclipse-score/score/commits/main/?since=2026-02-10&until=2026-02-10

PR #2560, opened 2026-02-07
DR-001-Arch: Consistent Stack vs. Reference Integration

qor-lb:
Formal decision record based on discussion during last architecture community workshop: https://github.com/orgs/eclipse-score/discussions/1922#discussioncomment-14871055

This intends to establish a discussion basis for further technical dr's such as:
- #2400
- #2346

> [!IMPORTANT]
> This PR will be kept open at least until 13th Feb before merging to ensure that necessary reviews can be conducted.

### DIFF
--- a/docs/design_decisions/DR-001-strat.md
+++ b/docs/design_decisions/DR-001-strat.md
@@ -24,9 +24,9 @@ SPDX-License-Identifier: Apache-2.0
 
 ---
 
-# Consistent Stack vs. Reference Integration
+## Consistent Stack vs. Reference Integration
 
-## Context
+### Context
 
 The project was initiated with ambiguous architectural assumptions. In some project material, S-CORE is described as a reference integration of independently evolving projects/modules. In parallel, discussions and expectations emerged that treat S-CORE as a coherent, continuously working stack.
 
@@ -41,7 +41,7 @@ Two categories of modules must be distinguished:
 1. Existing open-source projects/modules originating from other contexts and managed independently. These modules may be integrated into S-CORE if they are a good technical fit and align with S-CORE's functional goals. Their independent planning cadence and processes are explicitly acknowledged.
 2. Modules that do not yet exist and are created specifically to fulfill the purpose of S-CORE within the S-CORE GitHub organization. These modules are directly affected by this architectural decision.
 
-## Decision
+### Decision
 
 S-CORE is defined as a **continuously consistent stack**.
 
@@ -51,7 +51,7 @@ Modules created in the context of S-CORE exist to serve the stack. Their evoluti
 
 Existing independently managed open-source projects/modules may be integrated where appropriate. Their independence is accepted, but their integration into S-CORE is evaluated and justified exclusively in terms of stack-level goals.
 
-### No Constraint on Independent Module Delivery
+#### No Constraint on Independent Module Delivery
 
 This decision concerns architectural alignment and change justification. It does not impose constraints on modularity, packaging, or delivery models of the modules within the stack.
 
@@ -59,21 +59,36 @@ In particular, the consistent stack approach does not prevent modules or feature
 
 The decision only establishes that, when modules are integrated into S-CORE, their evolution and changes must be justified in terms of stack-level objectives.
 
-## Options Considered
+### Options Considered
 
-### Option 1: Reference Integration of Independent Modules
+The central question separating these options is not whether the S-CORE reference integration will be adopted directly and without adaptation in a downstream product (e.g. a series product) - it will probably not. The question is where integration effort is managed: driven by individual downstream needs at the module level, or consolidated at the stack level to maintain a continuously stable and transparent baseline.
+
+#### Option 1: Reference Integration of Independent Modules
 
 Modules evolve independently, with their own roadmaps and priorities. Integration occurs late and primarily before releases. Coordination effort is concentrated at integration time. Modules effectively behave as separate projects that are assembled into a reference configuration.
 
+In this model, a downstream project (e.g. a series project) can drive module evolution directly. Modules are updated or extended to meet a specific downstream need first; the impact on S-CORE integration is resolved subsequently. This reduces short-term friction for that one downstream context but increases the integration cost for S-CORE and for other downstream projects, which must absorb the resulting divergence without having been part of the original justification.
+
 This option favors module autonomy and minimizes continuous coordination. It shifts complexity to late integration phases and accepts reduced guarantees about the state of the full stack during development.
 
-### Option 2: Consistent Stack as Leading Use Case
+#### Option 2: Consistent Stack as Leading Use Case
 
 The stack is the primary architectural artifact. Modules contribute to and are validated against the behavior of the full stack. Continuous integration is used to ensure the stack remains functional throughout development.
 
-This option increases coordination and change management overhead during development. In return, it establishes a clear architectural contract. Changes are evaluated based on their impact on the stack, and the reference stack is expected to remain usable.
+In this model, changes must be justified in terms of stack-level objectives before integration. This may require additional upstreaming effort for any individual downstream project compared to Option 1. However, it establishes a stable and transparent reference integration that all downstream projects and end users can build upon, without needing to absorb divergences introduced for another downstream context.
 
-## Consequences
+The reference integration is not necessarily expected to be adopted without adaptation in downstream products. Its goal is to provide a continuously integrated, validated baseline - making downstream adaptation as straightforward as possible and providing a shared, openly visible justification platform for all downstream projects involved.
+
+This option also frames the reference integration as a continuous proof of integrability and quality, rather than a single downstream product delivery. In that framing, the reference integration is expected to:
+
+- demonstrate continuous integrability of all S-CORE modules across functional, performance, safety, and security perspectives
+- follow S-CORE process and quality expectations per module
+- define clear expectations for externally managed modules (for example, OpenSOVD) regarding required artifacts and interfaces
+- ensure gaps are addressed either by upstream artifacts or by explicit S-CORE maintenance responsibility
+
+This option increases coordination and change management overhead during development. In return, it establishes a clear architectural contract. Changes are evaluated based on their impact on the stack, and the reference stack is expected to remain usable throughout.
+
+### Consequences
 
 The consistent stack model constrains module-level autonomy. Not all module-local use cases are valid by default. Changes, especially breaking changes, must be justified in terms of stack-level requirements.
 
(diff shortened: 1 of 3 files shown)


masc2023:
Seems related to https://github.com/eclipse-score/score/pull/2346, Feature as stand-alone Delivery?

qor-lb:
> Seems related to #2346, Feature as stand-alone Delivery?

Yes agree, added it to the list in the description above.

PandaeDo (dismissed):
LGTM

opajonk (approved):
My understanding is the same, at least from the discussion in the last architecture on-site where I was also participating.

(Cross-linking another related DR: #2400)

FScholPer (approved):
What option is accepted or did I miss something?

docs/design_decisions/DR-001-strat.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Explicit Non-Goals
+
+This decision record does not:
+
+- Define concrete development processes or workflows.
+- Specify CI/CD implementations or quality gates.
+- Define release cadences or versioning strategies.
+- Provide a migration plan for existing modules.
+- Specify how architectural discussions are conducted or moderated.
```
aschemmel-tech, line 136: Please include something like: "This approach does not prevent the use-case of independent module (or feature) delivery. I.e. the modules/features are still independent enough to be able to be released as a SEooC.

a-zw (approved):

docs/design_decisions/DR-001-strat.md
qor-lb: @aschemmel-tech I added a paragraph to clarify that it is not the goal to constrain the modularity of the stack. Does this suffice or would you like to have the wording differently?

qor-lb:
> no accepted option defined

@FScholPer the decision is right after the context: https://github.com/eclipse-score/score/pull/2560/changes#diff-7605e72a4577a9aeeadf15aca38407709dc0eff885ee5684e8d46097ef932da8R44

docs/design_decisions/DR-001-strat.md
aschemmel-tech: Thank you for adding the paragraph, I think this suffices.

docs/design_decisions/DR-001-strat.md
```diff
@@ -0,0 +1,113 @@
...
+
+This decision is binding for the project. Any deviation requires a new, explicitly agreed Architecture Decision Record.
+
+### No Constraint on Independent Module Delivery
+
+This decision concerns architectural alignment and change justification. It does not impose constraints on modularity, packaging, or delivery models of the modules within the stack.
+
+In particular, the consistent stack approach does not prevent modules or features from being independently buildable or releasable, nor from being used outside the S-CORE stack.
+
+The decision only establishes that, when modules are integrated into S-CORE, their evolution and changes must be justified in terms of stack-level objectives.
```
a-zw, line 60: Imho the heart of the question is about responsibility for breaking changes and that goes beyond "changes must be justified in terms of stack-level objectives".

Let's assume module X makes a breaking change from version 42 to 43 and an S-CORE release is imminent. Now I see two options how integration can be handled and they somewhat correspond to the considered options 1 and 2:

Option 1: Module X does not integrate, thus we cannot release S-CORE yet. Integrators organize to get the version 43 improvements in.

Option 2: Module X does not integrate, thus we release S-CORE with old version 42. Module X if you want to be in the release, you need to work.

docs/design_decisions/DR-001-arch.md
```diff
@@ -0,0 +1,113 @@
...
+-->
+
+# DR-001-Arch: Consistent Stack vs. Reference Integration
+
+- **Date:** 2026-02-07
+
+```{dec_rec} Consistent Stack vs. Reference Integration
+:id: dec_rec__arch__consistent_stack_vs_reference
+:status: accepted
+:context: Architecture
```
a-zw:
```suggestion
:context: Strategy
```

This is not about architecture but about project strategy.

docs/design_decisions/DR-001-arch.md
```diff
@@ -0,0 +1,113 @@
...
+
+S-CORE is defined as a **continuously consistent stack**.
+
+The primary architectural goal of the project is the consistency and correctness of the full stack. All modules within S-CORE are expected to align with the objectives of the stack.
+
+Modules created in the context of S-CORE exist to serve the stack. Their evolution is guided by stack-level use cases, not by module-local objectives in isolation.
+
+Existing independently managed open-source projects/modules may be integrated where appropriate. Their independence is accepted, but their integration into S-CORE is evaluated and justified exclusively in terms of stack-level goals.
+
+This decision is binding for the project. Any deviation requires a new, explicitly agreed Architecture Decision Record.
```
a-zw:
```suggestion

```

This is the case for all decision records in general, so there is no need to mention this here.

qor-lb: is removed

docs/design_decisions/DR-001-arch.md
qor-lb: updated context

docs/design_decisions/DR-001-strat.md
qor-lb: @a-zw from my perspective both options would still be valid even under the current proposed decision. I have added a subsection to the _Impact on Governance and Planning_ section to try and clarify that. Let me know it this resolves your comment or if you feel differently.

qor-lb:
@aschemmel-tech @FScholPer are your concerns addressed? If yes, could you please approve so that we can move the PR forward?

docs/design_decisions/DR-001-strat.md
```diff
@@ -0,0 +1,123 @@
...
+
+This option favors module autonomy and minimizes continuous coordination. It shifts complexity to late integration phases and accepts reduced guarantees about the state of the full stack during development.
+
+### Option 2: Consistent Stack as Leading Use Case
+
+The stack is the primary architectural artifact. Modules contribute to and are validated against the behavior of the full stack. Continuous integration is used to ensure the stack remains functional throughout development.
+
+This option increases coordination and change management overhead during development. In return, it establishes a clear architectural contract. Changes are evaluated based on their impact on the stack, and the reference stack is expected to remain usable.
+
+## Consequences
```
FScholPer: Please add Option 3:    
- Goal for RefInt: 
       - To show that all S-CORE modules can be (continuously) integrated from a functional perspective, performance perspective and 
          dependability (safety and security)
      - Assumption: RefInt is not directly re-used in Series Products and is not a singular Safety Element out of Context
      - Each Module has to follow the S-CORE process 
      - Ensure Quality aspects (Quality Management Plan: Compiler errors, SCA, C0/C1, ....) in each of the modules and 
      - How to deal with stuff from outside? e.g. OpenSOVD 
         -  They need to provide the necessary artefacts as Docu, Safety Manual, etc.
         -  Or as work-around only: we have to have in S-CORE a maintainer to close the gaps
       

   - Assumption of Use of S-CORE RefInt
      - as much as possible re-use of the RefInt in Series Product without adaptations

docs/design_decisions/DR-001-strat.md
a-zw: Ok, I understand that this question is out of scope for this DR.

Now I wonder what is actually in scope? The central implication seems to be "Changes, especially breaking changes, must be justified in terms of stack-level requirements."

That feels like a weak wish. What happens if a module introduces a breaking change unjustified (intentionally or unintentionally)?

qor-lb: @a-zw the point of this DR is a step before enforcement: it's about establishing consensus what we want.

Before we can discuss how to handle unjustified breaking changes procedurally, or how to detect them with tooling, we need to agree on whether we want justification at all. The alternative is that S-CORE simply has to adapt during integration, and modules carry no obligation to the stack.

That's the question this DR intends answers: no, we don't just "deal with it", modules that are part of the stack are expected to justify breaking changes against stack-level requirements.

Everything you're asking about (what happens if they don't, how we enforce it, how automation can help) is exactly the set of follow-up discussions this DR is meant to enable. Those belong in subsequent DRs, because they only make sense once we've agreed on the principle. 

So I would push back slightly on "weak wish", it's an intentionally declarative statement. The teeth should come later, and that is by design.

FScholPer:
> > no accepted option defined
> 
> @FScholPer the decision is right after the context: https://github.com/eclipse-score/score/pull/2560/changes#diff-7605e72a4577a9aeeadf15aca38407709dc0eff885ee5684e8d46097ef932da8R44

can you just write the option which was Choosen or set the status to progress. Then we can just optimize the text in a next PR

qor-lb:
@FScholPer this is the decision (see [diff](https://github.com/eclipse-score/score/pull/2560/changes#diff-d55f8679f13e29a988135b2260574d7e3ba702c918eb57812a2bd96240aa5ab7R44) or [html](https://eclipse-score.github.io/score/pr-2560/design_decisions/DR-001-strat.html#decision)):

> S-CORE is defined as a **continuously consistent stack**.
> 
> The primary architectural goal of the project is the consistency and correctness of the full stack. All modules within S-CORE are expected to align with the objectives of the stack.
> 
> Modules created in the context of S-CORE exist to serve the stack. Their evolution is guided by stack-level use cases, not by module-local objectives in isolation.
> 
> Existing independently managed open-source projects/modules may be integrated where appropriate. Their independence is accepted, but their integration into S-CORE is evaluated and justified exclusively in terms of stack-level goals.

IMO merging this as "in progress" has the same effect as keeping this PR open. We should target to reach consensus soon so that we can baseline further DRs on this.

docs/design_decisions/DR-001-strat.md
qor-lb: @FScholPer after our discussion I think we are actually aligned. Your points aren't a separate option but fit naturally into Option 2, and I have updated the DR to make that clearer.

The thing that I think resolves the core of your comment: neither option assumes the reference integration is used 1:1 in a downstream product/series project. That's a shared assumption. The real question is just where you manage the integration effort - driven by individual downstream needs (Option 1), or at the stack level to keep a stable shared baseline (Option 2). I have added a framing paragraph to make this the explicit starting point for the options comparison.

I have also expanded the Option 2 description to explicitly cover the expectations you listed: continuous integrability across functional, performance, safety, and security perspectives; per-module adherence to S-CORE process and quality expectations; and clear interface expectations for externally managed modules like OpenSOVD, with gaps addressed either upstream or via explicit S-CORE maintenance responsibility.

One small thing on terminology: I replaced "series project/product" with "downstream project/product". Keeps the primary use case visible without implying this only matters for series production.

Does this land for you, or is there something you still feel is missing?

PR #2615, opened 2026-02-19
DR-008-infra first draft

a-zw:
Rendered: https://eclipse-score.github.io/score/pr-2615/design_decisions/DR-008-infra.html

### DIFF
--- a/docs/design_decisions/DR-008-infra.rst
+++ b/docs/design_decisions/DR-008-infra.rst
@@ -1,0 +1,85 @@
+..
+   Copyright (c) 2026 Contributors to the Eclipse Foundation
+
+   See the NOTICE file(s) distributed with this work for additional
+   information regarding copyright ownership.
+
+   This program and the accompanying materials are made available under the
+   terms of the Apache License Version 2.0 which is available at
+   https://www.apache.org/licenses/LICENSE-2.0
+
+   SPDX-License-Identifier: Apache-2.0
+
+DR-008-Infra: Docs Modularity
+================================================================
+
+.. dec_rec:: Modularize docs
+   :id: dec_rec__infra__docs_modularity
+   :status: proposed
+   :context: Infrastructure
+   :decision: tbd
+
+Context / Problem
+-----------------
+
+There is a proof-of-concept for an alternative documentation build concept introduced in
+`tooling PR 95 <https://github.com/eclipse-score/tooling/pull/95>`_.
+It improves the modularity of documentation but drops features of the existing documentation concept.
+
+Goals and Requirements
+^^^^^^^^^^^^^^^^^^^^^^
+
+- **Implementation Effort**: Don't spend much one-time effort to implement the change proposed here.
+- **Maintenance Effort**: Don't spend much ongoing effort due to increased complexity.
+- **Ease of use**: Developers don't like working on documentation, so we should make it easy on them.
+
+Options Considered
+------------------
+
+Option S: Single docs() only
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We keep the current docs-as-code concept and drop the PoC proposal.
+
+Implementation Effort 💚 No additional effort.
+
+Maintenance Effort 🤔 unclear in relation to PoC.
+
+Ease of use 💚 No change for developers.
+
+Option M: Many Bazel targets
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We drop the current approach and switch to the PoC approach.
+
+Implementation Effort 😡 Requires restructuring all documentation.
+
+Maintenance Effort 🤔 unclear
+
+Ease of use 😡 Requires developers to adapt to new workflow. No VSCode LSP integration.
+
+Option C: Combine the Best of Both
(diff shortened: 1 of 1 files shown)


MaximilianSoerenPollak:
Additional option here would be the current one. 

We keep the current Docs way and let the PoC exist in parallel but do not actively support it from docs-as-code/infra to not split our efforts. ?

docs/design_decisions/DR-008-infra.rst
```diff
@@ -0,0 +1,211 @@
...
+
+Context and Problem
+-------------------
+
+S-CORE aims to provide modules that can be smoothly integrated into safety-relevant
+ECU projects. To support this, S-CORE modules must provide their users with well-defined
+artifacts that clearly describe the content of a
+dependable element
+(`see process glossary <https://eclipse-score.github.io/process_description/main/glossary/index.html#terms>`_).
+The existing documentation build concept does not properly support this.
```
a-zw, line 30: @ramceb I'm not convinced that "The existing documentation build concept does not properly support this."

The current concept only has "a Bazel module" as reuseable element, so one per git repository. With option M, we can have any number of dependable elements. I don't see how that significantly improves the situation. Still "teams must manually structure documentation to match the S-CORE process".

docs/design_decisions/DR-008-infra.rst
```diff
@@ -0,0 +1,211 @@
...
+It improves the modularity of documentation and allows modules to be built
+hermetically by leveraging the features that Bazel provides as a build system.
+
+Goals
+-----
+
+- **Proper support of S-CORE adoption**: Enable teams to structure their documentation
+  according to the S-CORE process, with dedicated artifact types for requirements,
+  architectural design, safety analysis, and assumptions of use.
+- **Proper support of Bazel**: Ensure documentation builds adhere to Bazel's core
```
a-zw, line 43: @ramceb I'm not happy with how this goal is phrased. It sounds like "don't ask what Bazel can do for you, instead ask what you can do for Bazel."

The real goals are rather "build correctness" or "reproducability" or "build speed". The means to achieve them are hermeticity, caching, parallelism, and Bazel does a great job providing the mechanisms.

docs/design_decisions/DR-008-infra.rst
```diff
@@ -0,0 +1,211 @@
...
+  principles — hermeticity, reproducibility, and correct dependency tracking — enabling
+  action caching, remote caching, remote execution, and parallel builds.
```
a-zw, line 45: I'm not yet quite sure what this PR is about. If it is primarily about using `bazel build` (instead of `bazel run`), then we can have that much cheaper than option M. See the abandoned https://github.com/eclipse-score/docs-as-code/pull/286.

However, if it is about having concepts like "dependable element" within Bazel, then aspects like hermeticity are missing the point.

While both questions are closely related, is it really the case that deciding one implies the other? If not, then we should have to separate decision. If yes, then we should explicitly select in the Goals section which one is more important.

docs/design_decisions/DR-008-infra.rst
```diff
@@ -0,0 +1,211 @@
...
+   Rather than relying on a single shared ``conf.py``, the PR generates a tailored
+   Sphinx configuration for each module from a ``conf.template.py`` template using
+   ``ctx.actions.expand_template()``. External needs are loaded from the
+   build-action-produced JSON file, keeping each module's configuration self-contained
+   and reproducible.
+
+Assessment
+~~~~~~~~~~
+
+- **Proper support of S-CORE adoption** 💚 Purpose-built rules for each S-CORE
```
a-zw, line 164: How does the Module-API handle traceability to implementation and tests?

docs/design_decisions/DR-008-infra.rst
ramceb: Its correct that one git repository contains in general one bazel module. But one bazel module can contain several dependable_elements. So a bazel module corresponds to a "Delivery container" as described in [process description](https://github.com/eclipse-score/process_description/blob/6bd4f67a3854d0b34cf820a3eb10484ca7375ac2/process/general_concepts/score_building_blocks_concept.rst?plain=1#L52)

docs/design_decisions/DR-008-infra.rst
ramceb: I am fine with rewording this goal. What I want to say is that we should leverage the mechanisms of the build infrastructure to support adherence to the processes

docs/design_decisions/DR-008-infra.rst
ramceb: you are correct. We are addressing 2 points. Its just that the one (proper use of bazel) is a precondition for the other (supporting process adherence by means of bazel)

docs/design_decisions/DR-008-infra.rst
ramceb: The module_api concept foresees that each dependable element provides an API, where any user of the dependable element can obtain the information needed (e.g. traceability, assumptions_of_use, etc) to integrate the module in his system. The traceability information is calculated inside of the module as well as any documentation is generated. . 
e.g. if you have define a dependable_element "foo" for your module. 
Then you execute `bazel build/test //:foo` and the necessary artifacts e.g. coverage report, test report, etc will be provided by bazel either via test execution or documentation generation.

PR #2636, opened 2026-02-27
Add tool requirements list

PandaeDo:
Create of initial list tool requirements of process requirements that will not be solved by docs-as-code
- https://github.com/eclipse-score/process_description/issues/589#issuecomment-3965819361
- Compare to https://eclipse-score.github.io/docs-as-code/main/internals/requirements/process_overview.html

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -72,11 +72,6 @@ bazel_dep(name = "rules_java", version = "8.15.1")
 # Score custom modules loading
 #
 ###############################################################################
-bazel_dep(name = "score_tooling", version = "1.1.0")
+bazel_dep(name = "score_tooling", version = "1.1.2")
 bazel_dep(name = "score_docs_as_code", version = "3.0.1")
-bazel_dep(name = "score_process", version = "1.4.3")
-git_override(
-    module_name = "score_process",
-    commit = "55cfdccef3066ed0e8d712fdfdc7b314be6a88f6",
-    remote = "https://github.com/eclipse-score/process_description",
-)
+bazel_dep(name = "score_process", version = "1.4.4")


docs/score_tools/tools_requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+
+Tools Requirements
+==================
+List of all requirements for tools of S-CORE that are not defined or adressed to specific repositories like docs-as-code.
+
+GitHub
```
masc2023: Is this a header or just a first tool, for which requirements exists?

docs/score_tools/tools_requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+^^^^^^
+
+Satisfied
```
masc2023: IF the above is a list, with one tool only, may link from above to the list here, which process requirements the tools is fulffiling

docs/score_tools/tools_requirements/index.rst
```diff
@@ -0,0 +1,91 @@
...
+#########
+
+.. list-table:: GitHub
+    :header-rows: 1
+
+    * - ID
+      - Title
+      - Rationale
+
+    * - gd_req__doc_author
```
masc2023: should that not be need, so it can directly be linked to the process requirement?

docs/score_tools/tools_requirements/index.rst
PandaeDo: Changed description

docs/score_tools/tools_requirements/index.rst
PandaeDo: Changed description

docs/score_tools/tools_requirements/index.rst
PandaeDo: Add links

MODULE.bazel
```diff
@@ -73,5 +73,10 @@ bazel_dep(name = "rules_java", version = "8.15.1")
 #
 ###############################################################################
 bazel_dep(name = "score_tooling", version = "1.1.0")
-bazel_dep(name = "score_docs_as_code", version = "3.0.0")
+bazel_dep(name = "score_docs_as_code", version = "3.0.1")
 bazel_dep(name = "score_process", version = "1.4.3")
+git_override(
```
masc2023: would prefer we make a new release of process description, instead of using git_override

PandaeDo: I learned that we use override until we made a new release. But this is also possible

masc2023: Created:
DEBUG: registry_manager.main Modules to be updated: ['score_process']
 INFO: registry_manager.main ---------------------------------------------------
 ::notice::registry_manager.main Updating score_process from 1.4.3 to 1.4.4

masc2023: Available for you, https://github.com/eclipse-score/bazel_registry/pull/322/changes

MaximilianSoerenPollak: Yes we excplitily allowed using git_override now during development, but are (unsure if in 0.7 or 0.8 ref-int release) requiring all of them to be removed in the release that is used in ref-int as they aren't transient.

PandaeDo: Deleted the override

PR #2695, opened 2026-03-17
Update GCC to "no impact"

alekseyborisyukvalidas:
- MOM(n.1): https://github.com/orgs/eclipse-score/discussions/407#discussioncomment-16049119


aschemmel-tech (approved):
evaluation according to discussed use case

PR #2480, opened 2026-01-21
Comp Req inspection - Bitmanipulation

aschemmel-tech:
Resolves: #2479

PR #2696, opened 2026-03-18
lib/bitmanipulation: fix bounds safety requirement.

Rutik7:


docs/modules/baselibs/bitmanipulation/docs/requirements/index.rst
aschemmel-tech: As stated in the inspection comments (see https://github.com/eclipse-score/score/pull/2480), this aou may be obsolete.

Rutik7: Done

PR #2702, opened 2026-03-20
Draft QNX restriction documentation

aschemmel-tech:


ANegm-ETAS:
This is a helpful addition but it raises a question:

The process describes linking Assumptions of Use (AoUs) to architecture elements to show compliance. This works well for positive requirements, but a significant number of restrictions in the QNX safety manual are negative—they prohibit the use of certain functions.

How, then, should a feature team 'link' the absence of a forbidden function to an architecture element to demonstrate compliance?

aschemmel-tech:
> This is a helpful addition but it raises a question:
> 
> The process describes linking Assumptions of Use (AoUs) to architecture elements to show compliance. This works well for positive requirements, but a significant number of restrictions in the QNX safety manual are negative—they prohibit the use of certain functions.
> 
> How, then, should a feature team 'link' the absence of a forbidden function to an architecture element to demonstrate compliance?

Let's for example have an AoU like "the component shall not use exceptions". By linking this to the component you confirm that you did some check to prove this, for example run a test/chek-script or do a manual code analysis.

ANegm-ETAS:
> Let's for example have an AoU like "the component shall not use exceptions". By linking this to the component you confirm that you did some check to prove this, for example run a test/chek-script or do a manual code analysis.

so in this example, Are we expecting all architecture elements in all components to link to this AOU and create a test case for them as well ?

aschemmel-tech:
> > Let's for example have an AoU like "the component shall not use exceptions". By linking this to the component you confirm that you did some check to prove this, for example run a test/chek-script or do a manual code analysis.
> 
> so in this example, Are we expecting all architecture elements in all components to link to this AOU and create a test case for them as well ?

To be precise: all architecture static views of all the components. But there is an alternative to this: we can also link to an existing requirement or we cover the AoU by e,g, a coding guideline entry (without linking, but by adding a comment in the description). The most efficient way of course would be to write a static code analysis check to cover this for all components at once.

ANegm-ETAS:
> > so in this example, Are we expecting all architecture elements in all components to link to this AOU and create a test case for them as well ?
> 
> To be precise: all architecture static views of all the components. But there is an alternative to this: we can also link to an existing requirement or we cover the AoU by e,g, a coding guideline entry (without linking, but by adding a comment in the description). The most efficient way of course would be to write a static code analysis check to cover this for all components at once.

we actually looked into creating a static analysis check for QNX restrictions and although it's technically possible according to legal we can only do so without allowing any direct inference of the restrictions (no one should be able to use the tool to reverse engineer the qnx safety manual) and Only violations may be reported , so no explanation of the rule violation either .

we could cover the AoUs by a coding guideline but wouldn't the coding guideline here consist of the restriction list similar to the AoUs themselves . Would linking the AoUs to the Implementation checklist be acceptable coverage

aschemmel-tech:
Added a way to verify QNX AoU for non-functional, non-testable restrictions (which is the standard way in S-CORE to do requirements verification by analysis).

docs/modules/os/docs/requirements/aou_req.rst
```diff
@@ -12,8 +12,8 @@
    # SPDX-License-Identifier: Apache-2.0
    # *******************************************************************************
 
-Assumptions of Use
-##################
+Generic Assumptions of Use
```
masc2023, line 15: Why not "Operating System Assumptions of Use", compare QNX title

docs/modules/os/docs/requirements/aou_req_qnx.rst
```diff
@@ -0,0 +1,60 @@
...
+(static or dynamic views) the AoUs which their module's components cover.
+This can be used by the system integrator as a starting point to show compliance to the QNX AoU.
+In the below list all AoU/Restrictions are set to "invalid" if those are not applicable to the S-CORE
+SW platform, including an argumentation why.
+
+The verification for such an AoU is done in the same way as for other requirements, i.e. if these are
+non-functional and "non-testable" then a "requirement analysis" is done as described in
+:need:`doc__component_name_req_inspection` checklist item REQ_08_02. To document this a PR is created
+to modify the component architecture "fulfils" attribute, add the respective requirements to the
+requirements inspection checklist's scope and document findings like for other inspections.
```
ANegm-ETAS, line 40: I have a few questions on this paragraph:

1-  so do we imply that for functional requirements we expect to introduce tests in all components.

2- For non-functional requirements do we need to update architecture “fulfils” attribute and in addition to that we then would perform a Requirements Inspection on the AOUs ?
Its not very clear to me where the requirement inspection comes in, did you mean implementation checklist

aschemmel-tech: 1: yes, this is the expecation that every functional requirement/AoU should be able to be tested, we may need to keep these test specs (but not the results) private as you rightly pointed out
2: Yes, the fulfils together with the verification evidence in the requirements inspection document show the requirements analysis - see :need:`doc__component_name_req_inspection` checklist item REQ_08_02

aschemmel-tech: My proposal in call with @ANegm-ETAS  to improve this by adding a hint to the QNX restrictions to the SW development plan (https://eclipse-score.github.io/score/main/platform_management_plan/software_development.html section "coding guidelines") I have to withdraw again. I do not think we should show a selection of a certain OS at this prominent point. Would need a project leads decision - could be part of a future improvement.

PR #2747, opened 2026-03-31
Decision record - reference_integration scope

PiotrKorkus:
# Decision record for reference_integration scope

## Review checklist
- [ ] Persistency @vinodreddy-g 
- [ ] Logging @pawelrutkaq @rmaddikery @arsibo 
- [x] Lifecycle @pawelrutkaq @NicolasFussberger 
- [ ] Kyron @pawelrutkaq 
- [ ] Orchestration @vinodreddy-g 
- [x] Baselibs @4og 
- [x] Baselibs_rust @pawelrutkaq @4og 
- [ ] Communication @LittleHuba 
- [ ] FEO @johannes-esr 
- [ ] Infrastructure @AlexanderLanin 
- [ ] Technical Leads @antonkri @FScholPer @qor-lb 

Every CFT shall provide their status - OK or NOK

### DIFF
--- a/docs/design_decisions/DR-008-int.rst
+++ b/docs/design_decisions/DR-008-int.rst
@@ -1,0 +1,110 @@
+..
+   Copyright (c) 2026 Contributors to the Eclipse Foundation
+
+   See the NOTICE file(s) distributed with this work for additional
+   information regarding copyright ownership.
+
+   This program and the accompanying materials are made available under the
+   terms of the Apache License Version 2.0 which is available at
+   https://www.apache.org/licenses/LICENSE-2.0
+
+   SPDX-License-Identifier: Apache-2.0
+
+DR-008-Int: S-CORE integration scope in reference_integration repository
+========================================================================
+
+- **Date:** 2026-03-31
+
+.. dec_rec:: Integration scope in reference_integration repository
+   :id: dec_rec__int__scope_reference_integration
+   :status: accepted
+   :context: Integration
+   :decision: Option 2
+
+Context / Problem
+-----------------
+
+Currently there is no clear definition of scope and responsibilities of the ``reference_integration`` repository.
+The process of integration of S-CORE modules is time consuming and leaves open space for misalignment
+between the process and the feature teams.
+
+Goals and Requirements
+^^^^^^^^^^^^^^^^^^^^^^
+
+- **Effort**: The integration process should be as efficient as possible with clear feedback loops.
+- **Independence**: Modules should be independent with their implementation scope.
+- **Integration**: Modules should have clear common requirements for S-CORE release integration.
+- **Clear Ownership**: Each module should have a clear vision of responsibilities for integrating into reference_integration.
+- **Maintainability**: Keep long-term maintenance effort low.
+
+Non-Goals
+~~~~~~~~~
+
+- Controlling every release of each Module - Modules can release independently and some releases can be not integrated into S-CORE.
+
+Options Considered
+------------------
+
+Option 1: Execute only Feature Integration Tests in reference_integration repository
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Reference_integration repository is a place where only Feature Integration Tests are executed.
+Every Module is responsible for executing static tests, UTs, CITs in their respective repositories,
+used configurations, compiling flags and workflows should be aligned with S-CORE guidelines.
+
+Reference_integration will provide a means to conduct continuous integration, based on only Feature Integration Tests,
+of latest Modules states to ensure early feedback and breaking changes notifications.
+
+For S-CORE releases, Modules need to deliver artifacts from local tests as their release assets to be used for certification and documentation.
+
+Pros:
+* Modules have full control over their own scope
+* Quick integration jobs in reference_integration repository as only Feature Integration tests are executed
+
+Cons:
+* No control over configuration, compiling flags, variants
+* Documentation and artifacts are distributed across different repositories
+* Full-stack S-CORE release might resolve to different version of dependencies that were tested in Module repositories
+which makes the release not compliant with S-CORE release requirements.
+
+Option 2: Re-execute all quality checks in the reference_integration repository
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Reference_integration repository is a place where all kinds of tests - static, UTs, CITs, FITs are re-executed
+for full verification of the S-CORE full-stack. Every Module needs to ensure their tests are executable in reference_integration
+repository. For the verification process, common configuration, compiling flags are used across all Modules to ensure the compliance with current S-CORE guidelines.
+
+Reference_integration will provide a means to conduct continuous integration of latest Modules states to ensure early feedback and breaking changes notifications.
+
+All necessary artifacts and single documentation build are generated in reference_integration repository to be used for certification and documentation
+of S-CORE releases.
+
+Pros:
+* Single point of truth for configurations, compiling flags, documentation and certification artifacts
+* Better control over resolved dependencies
+* Better traceability of the tests and documentation for full-stack S-CORE releases
+
+Cons:
+* Additional effort for the Module teams to maintain their tests and dependencies to be fully executable in reference_integration repository
+* Longer integration jobs in reference_integration repository as all tests are executed
+
+Evaluation
+----------
+
+Option 1 makes S-CORE provide a set of SEooCs without being a full-stack project that has been selected
+in :need:`dec_rec__strat__consistent_stack_vs_reference`.
+Modules are responsible for their own compliance with the S-CORE release requirements, which makes it difficult
+to validate the final S-CORE release for the process compliance.
+The biggest risk is that Module A can deliver a release and artifacts with dependency on Module B in version x.y.z,
+but in reference_integration thus overall S-CORE release, Module B will be resolved in version x.y.z+1,
+which makes Module A artifacts not compliant.
+
+Option 2 truly makes S-CORE a full-stack project and provides a single point of truth for the
+configurations, compiling flags, variants, documentation and certification artifacts.
+However, it requires additional effort for the module teams to maintain their tests and dependencies to be fully
+executable in the reference_integration repository with both public APIs and tests.
+It allows generating a single documentation build with full traceability across different repositories which will stay persistent
+and will not require external linking.
+
+**Decision:** Option 2 has been most favored by the community.
+We accept additional effort for the integration process in exchange for a better traceability.
--- a/docs/design_decisions/DR-008-int.rst
+++ b/docs/design_decisions/DR-008-int.rst
@@ -19,7 +19,7 @@ DR-008-Int: S-CORE integration scope in reference_integration repository
    :id: dec_rec__int__scope_reference_integration
    :status: accepted
    :context: Integration
-   :decision: Option 2
+   :decision: Option 3
 
 Context / Problem
 -----------------
@@ -58,14 +58,15 @@ of latest Modules states to ensure early feedback and breaking changes notificat
 For S-CORE releases, Modules need to deliver artifacts from local tests as their release assets to be used for certification and documentation.
 
 Pros:
+
 * Modules have full control over their own scope
 * Quick integration jobs in reference_integration repository as only Feature Integration tests are executed
 
 Cons:
+
 * No control over configuration, compiling flags, variants
 * Documentation and artifacts are distributed across different repositories
-* Full-stack S-CORE release might resolve to different version of dependencies that were tested in Module repositories
-which makes the release not compliant with S-CORE release requirements.
+* Full-stack S-CORE release might resolve to different version of dependencies that were tested in Module repositories which makes the release not compliant with S-CORE release requirements.
 
 Option 2: Re-execute all quality checks in the reference_integration repository
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -80,13 +81,90 @@ All necessary artifacts and single documentation build are generated in referenc
 of S-CORE releases.
 
 Pros:
+
 * Single point of truth for configurations, compiling flags, documentation and certification artifacts
 * Better control over resolved dependencies
 * Better traceability of the tests and documentation for full-stack S-CORE releases
 
 Cons:
+
 * Additional effort for the Module teams to maintain their tests and dependencies to be fully executable in reference_integration repository
 * Longer integration jobs in reference_integration repository as all tests are executed
+
+Option 3: Front-load quality checks in modules with lightweight reference_integration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Modules follow the S-CORE process and run quality tooling in their own context against their own resolved
+dependency set. Across modules, teams agree on *what* quality tooling and toolchains to use, but each
+module runs them independently. Most importantly, teams agree on dependency versions for the upcoming
+S-CORE release: if a module is releasing a new version into the next S-CORE release, every module that
+depends on it must migrate to that version before the release. Modules minimize non-dev, non-S-CORE
+dependencies in their MODULE.bazel files.
+
+Module Owners deliver as part of their release artifacts in the common form allowing aggregation and
+full documentation build in reference_integration. Bazel dependency analysis will allow verification
+that collected artifacts are valid for S-CORE release - resolved dependencies for the Module (in scope of their repository)
+must match the dependencies for S-CORE release and resolved in reference_integration.
+
+Continous integration of latest Modules based on hashes from their main branches will not allow verification
+of the artifacts and full documentation build. That will remain exclusive for S-CORE releases.
+
+Pros:
+
+* Verifying quality where it originates is more reliable — modules have full knowledge of their domain, their specific quality requirements, and the exact dependency context they were tested against
+* Technically simpler reference_integration with a well-defined, minimal scope
+* Avoids re-executing all checks in a different dependency context, which can mask or introduce issues
+* Provides early feedback through regular release candidate publishing
+
+Cons:
+
+* Requires tighter communication and coordination across module teams
+* Agreeing on and enforcing a shared dependency manifest adds process overhead
+* Backward compatibility guarantees need to be actively maintained across all modules
+
+Comparison
+----------
+
+.. list-table:: Reference_integration options comparison
+   :header-rows: 1
+   :widths: 30 20 20 30
+
+   * - Criterion
+     - Option 1
+     - Option 2
+     - Option 3
+   * - Quality checks location
+     - Module repositories only
+     - reference_integration (re-executed)
+     - Module repositories (front-loaded)
+   * - Feature integration tests
+     - reference_integration
+     - reference_integration
+     - reference_integration
+   * - Dependency mismatch risk
+     - High
+     - Low
+     - Low (via version agreements)
+   * - Integration job duration
+     - Short
+     - Long
+     - Short
+   * - Artifact & documentation consolidation
+     - Distributed across repositories
+     - Single point of truth
+     - Consolidated at release time
+   * - Module team maintenance effort
+     - Low
+     - High
+     - Medium
+   * - Cross-module coordination overhead
+     - Low
+     - Medium (triage of CI failures)
+     - Medium (shared dependency manifest)
+   * - Traceability for S-CORE release
+     - Weak
+     - Strong
+     - Strong (via release artifacts)
 
 Evaluation
 ----------
@@ -106,5 +184,21 @@ executable in the reference_integration repository with both public APIs and tes
 It allows generating a single documentation build with full traceability across different repositories which will stay persistent
 and will not require external linking.
 
-**Decision:** Option 2 has been most favored by the community.
-We accept additional effort for the integration process in exchange for a better traceability.
+Option 3 front-loads quality verification into the modules themselves and keeps reference_integration
+lightweight. It addresses the dependency mismatch risk of Option 1 through coordinated version
+agreements across modules rather than by re-executing checks centrally. However, it relies on
+disciplined inter-module coordination and backward compatibility guarantees, which adds process
+overhead and requires a shared dependency manifest to remain manageable.
+
+.. _option3_infographic:
+
+.. figure:: _assets/DR-008-int-option3_infographic.svg
+   :align: center
+   :width: 75%
+
+   Option 3 visual breakdown
+
+**Decision:** Option 3 got positive feedback and was selected by the community.
+We accept a trade-off of full documentation beeing available only for S-CORE releases,
+while having a more efficient and reliable integration process without extra efforts to maintain
+internal test targets in reference_integration repository.
--- a/docs/design_decisions/DR-008-int.rst
+++ b/docs/design_decisions/DR-008-int.rst
@@ -13,13 +13,13 @@
 DR-008-Int: S-CORE integration scope in reference_integration repository
 ========================================================================
 
-- **Date:** 2026-03-31
+- **Date:** 2026-06-11
 
 .. dec_rec:: Integration scope in reference_integration repository
    :id: dec_rec__int__scope_reference_integration
    :status: accepted
    :context: Integration
-   :decision: To be decided in voting
+   :decision: Option 4
 
 Context / Problem
 -----------------
@@ -126,7 +126,7 @@ Module-side activities required for an S-CORE release integration:
 * `Module Release Notes <https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workproducts.html#wp__module_sw_release_note>`_
   produced from the S-CORE module release-note template and verified per the
   `module release-note workflow <https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workflow.html#wf__rel_mod_rel_note>`_.
-   :widths: 30 20 20 30 24
+
 reference_integration activities required for an S-CORE release:
 
 * Feature Integration Tests (FITs).
@@ -139,6 +139,14 @@ reference_integration activities required for an S-CORE release:
   `Platform Handbook <https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workflow.html#wf__rel_platform_handbook>`_,
   and the
   `Platform Verification Report <https://eclipse-score.github.io/process_description/main/process_areas/verification/verification_workproducts.html#wp__verification_platform_ver_report>`_.
+
+.. _option3_infographic:
+
+.. figure:: _assets/DR-008-int-option3_infographic.svg
+  :align: center
+  :width: 75%
+
+  Option 3 visual breakdown
 
 Pros:
 
@@ -283,14 +291,6 @@ it improves feedback on dependency and compatibility issues by testing against a
 full-stack baseline rather than relying only on version agreements. The remaining cost is additional
 pipeline orchestration and a need to define how documentation and release evidence are consolidated
 when Module validation runs with temporary dependency overrides.
-
-.. _option3_infographic:
-
-.. figure:: _assets/DR-008-int-option3_infographic.svg
-  :align: center
-  :width: 75%
-
-  Option 3 visual breakdown
 
 **Decision:** Option 4 got positive feedback and was selected by the community.
 We accept the additional orchestration needed to validate Modules against the dependency set
--- a/docs/design_decisions/DR-008-int.rst
+++ b/docs/design_decisions/DR-008-int.rst
@@ -19,7 +19,7 @@ DR-008-Int: S-CORE integration scope in reference_integration repository
    :id: dec_rec__int__scope_reference_integration
    :status: accepted
    :context: Integration
-   :decision: Option 3
+   :decision: To be decided in voting
 
 Context / Problem
 -----------------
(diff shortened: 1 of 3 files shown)


pawelrutkaq (changes requested):
Blocker as we need feedback gathering from all (Each CFT shall give OK/NOK)

antonkri (dismissed):

AlexanderLanin (dismissed):

evinoth1206 (approved):

vinodreddy-g (dismissed):

NicolasFussberger (dismissed):

qor-lb (dismissed):

anmittag:
Assuming Agreeing is voting for Option 2. ETAS position is clear Option 2.

anmittag (dismissed):

AlexanderLanin:
> Assuming Agreeing is voting for Option 2. ETAS position is clear Option 2.

Correct. The document states:

>**Decision:** Option 2 has been most favored by the community.
We accept additional effort for the integration process in exchange for a better traceability.

---

Although I'm personally not sure this is extra effort, or better traceability, I favor option 2 for all the other reasons listed in the document :-)

4og (changes requested):
Unfortunately, I cannot agree with Option 2 as in my opinion not all viable alternatives have been considered. I will propose one below as Option 3. In general I agree with the strive towards a "full-stack" solution with common tooling and toolchain for all S-CORE modules (this already makes Option 1 non-viable as it dictates full independence). I understand that the main reasoning for Option 2 is to mitigate the risk that Minimal Version Selection (MVS) algorithm in Bazel brings when all modules are put together, and to have better controllability of the quality tooling used in the project. However, Option 2 has major flaws that made me think about alternatives that are good enough to mitigate these risks.

### Flaws of Option 2

**Flaw 1: Non-dev dependencies for quality tooling**

Running quality tooling in the reference integration requires making quality tooling dependencies in modules non-dev-dependencies. This has been discussed in several calls already. @LittleHuba was mainly raising this issue. (For specifics on version selecting algorithm, see https://bazel.build/external/module#version-selection, there is also a link to Go's reasoning why MVS works this way).

Maybe non-dev dependencies are problematic because they increase the dependency footprint, which is an anti-pattern in software engineering when the goal is actually to reduce the number of dependencies. In production integration environments there are much more dependencies, not only coming from S-CORE modules, and the same MVS algorithm will be kicking in as it has to satisfy dependency chains of these now non-dev quality tools (this includes all transitive dependencies along the chain!) alongside everything else in those environments. This is not a theoretical issue. We are currently trying to incorporate S-CORE modules into our system and see how MVS is changing dependencies of non-S-CORE components and breaking the build.

I understand we can "hack" it and still make all quality dependencies dev-dependencies, and maintain patches in the reference integration to un-dev them. But I don't feel it's a good way forward as we're doing the reference integration not only for the sake of having it, but also to serve as a blueprint for actual automotive integration projects.

**Flaw 2: Not all quality checks can be re-run generically**

Re-running quality checks in the reference integration on the basis that Bazel chooses a different version set for dependencies may work for some [work products](https://eclipse-score.github.io/process_description/main/workproducts/index.html), but not all. For example, some things may need to be done manually, such as verifying for AoU violations. And most importantly, a centralized quality pipeline will not cover module-specific checks that don't belong in a generalized setup. This is just one example, but let's say a security analysis may advise to run a fuzzer on a specific component. A module may have similar domain-specific quality requirements that cannot be centralized.

**Flaw 3: Unclear feedback loop**

What is the feedback loop when the reference integration discovers a failure because of the version selection?
I suppose the affected modules will need to fix the issue and release an update. Over time, this means everyone converges to use the same versions of dependencies, which makes the quality checks in the reference integration redundant as they end up repeating exactly what was already verified in the modules themselves with no added value. And here we come to Option 3.

### Proposal: Option 3

The main idea is to do front-loading in modules themselves via organizational and technical means:

- Modules follow the S-CORE process.
- We agree across modules on *what* quality tooling and toolchains to use, but each module runs them in its own context with its own resolved dependency set (a module can still use something extra for module-specific use cases).
- Most importantly, we agree on dependency versions for the upcoming S-CORE release. Example: if `score_logging` is releasing v1.5 to the next S-CORE release, every other module that depends on logging has to migrate to v1.5 (see also backward compatibility note under General Guideline below).
- Modules minimize the number of non-dev and non-S-CORE dependencies in their MODULE.bazel files.

**What checks remain in the reference integration:**

- Consolidate safety artifacts from all module releases.
- Verify that the module safety package adheres to the S-CORE process (we can discuss if it can be done somewhere else, e.g. the Bazel registry).
- Run feature integration tests.
- Run consistency checks to discover mismatches in the dependency graph. The `bazel mod` command can help here.
  - If a mismatch is detected: affected modules will need to release an update.
  - Or depending on what kind of mismatch it is: provide safety argumentation why it's not a problem.

**To avoid big-bangs shortly before a release and discover problems earlier:**

- Module maintainers regularly publish release candidates of their modules to the reference integration, either in form of releasing to the registry or as git hashes. I believe this can be well automated, though the exact mechanism would need to be worked out.
- Immediate feedback to module owners in case of version mismatches, failed integration tests, or incomplete safety packages.

**On the coordination cost:** Option 3 requires tighter communication across modules. In practice this could be supported by a shared dependency manifest (a file listing agreed-upon dependency versions for the upcoming release, covering not only direct but also transitive dependencies) and a release planning process where module teams align on timelines and versions. I don't think this cost is much higher than with Option 2, because Option 2 also requires coordination when reference integration failures need to be triaged and fixed by the responsible module teams.

**General Guideline**: Strive for backward compatibility guarantees in all modules and clear deprecation policies. This is particularly important for Option 3 to work, because it reduces friction when modules update to agreed-upon dependency versions. It's also important in general as Bazel's version selection doesn't care about SemVer. It uses it (a relaxed version of it to be precise) for comparison only to know which version is newer or older during the resolution process. Bazel may very well choose for example major version 4.0 for all even if some module depends on 1.2. I know it can be selectively worked around with `multiple_version_override`, but I'm not sure we want to go this way.

I think with Option 3 we can avoid all three flaws mentioned above and make the integration technically simpler. Verifying quality where it originates, in the modules, is more reliable than repeating it in the reference integration. It's because modules have full knowledge of their own domain, their specific quality requirements, and the exact dependency context they were tested against.

I'm also sure we can brainstorm more options (e.g. like bumping dependencies to a consistent state and running checks in the scope of the module and not in the scope of the reference integration).

At this stage of the project I think it is more important to gather learnings about using Bazel with bzlmod in integration projects (not only in S-CORE itself, but also in series projects) rather than investing into a complicated integration setup as described in Option 2, which may turn non-viable. Depending on the learnings we can steer into one or another direction.

pawelrutkaq:
So since I am only here a "move topic forward guy," I recirculate this again to TL: @anmittag @thilo-schmitt @FScholPer  @qor-lb, please pick up in TL meeting how to move forward - for now, this is a **blocker** for **0.7** ;)

From tech opinion: This (Option 3) is viable, but part of the agreement would be module owners agree to run specific CI workflows (actions, whatever), i.e., on release, so we can gather **consistent** artefacts (to be defined by ref_int needs). This could not be anymore each one doing something a little bit different (ofc - running more, own checks always is okay). 
Note: Using git hashes in ref int during CI(main) (milestone phase 0 - 6 weeks of dev) will not be able to provide a quality artefacts view of modules.

AlexanderLanin:
Thanks @PiotrKorkus. I've just removed all agreements. @everyone please review the new proposal to go for option 3.

docs/design_decisions/_assets/DR-008-int-option3_infographic.svg
FScholPer, line 1: What I not understand is how to resolve compilation issues. Is that addressed with consistency checks or what if module A is compiling well in its own repo but not in reference because differences in flags, used compilers etc. Also how will be safety compliance addressed as this is not only artifacts but also checks like unit test coverage, static / dynamic code analysis. Manual Checklist fulfilment. Impact Analysis and tracing differences. How to handle if module a brings in tool x but this is not qualified etc.

AlexanderLanin:
Personally I firmly believe that Option 2 is the most efficient solution *for a project*, based on both personal and company-wide experience with integration across multiple projects.

Having said that, it may not be the best fit for a distributed open source project. I need to dwell on that a little more.

Regarding Option 3 — maybe I misunderstand some points?

* Cross-module changes require local changes in every module. Upgrading googletest or introducing a new compiler flag can only be evaluated by performing the change in each module individually. This shifts integration effort from automation to coordination and slows down experimentation significantly.
* Hardware- or environment-specific unit/component test execution will likely require per-module solutions, which makes portability harder.
* Immediate feedback is reduced. For example, a change in baselibs-main can break a unit test in communication-main, but no-one will notice until baselibs creates a release and communication-main explicitly upgrades. This delays feedback and allows incompatibilities to accumulate silently. In practice, this resembles a form of “big bang integration”, only partially mitigated by feature integration tests. It creates a risk of release-oriented “catch-up integration” rather than continuous full-stack validation.
* Drift in development dependencies (e.g. test frameworks, tooling, generators) is not addressed.

docs/design_decisions/_assets/DR-008-int-option3_infographic.svg
pawelrutkaq: **artifacts** here means - unit test results, unit test coverage, static analysis reports, whatever other reports

- Module owners are reposible to make it work in reference_integration - that's what will be part of the next DR
- We will check bazel graph for ensuring consistency
- If the module brings tool X that is not qualified, it's not a problem. This can only be in form on "on top" of qualified checks - so all will provide the same, uniform artefacts
- All the safety configurations like flags, cpp polices comes from bazel mod so it will be ensured by point 2. and if the module changes something, he will anyway need to provide AoU globally why we need that.

PiotrKorkus:
> Personally I firmly believe that Option 2 is the most efficient solution _for a project_, based on both personal and company-wide experience with integration across multiple projects.
> 
> Having said that, it may not be the best fit for a distributed open source project. I need to dwell on that a little more.
> 
> Regarding Option 3 — maybe I misunderstand some points?
> 
> * Cross-module changes require local changes in every module. Upgrading googletest or introducing a new compiler flag can only be evaluated by performing the change in each module individually. This shifts integration effort from automation to coordination and slows down experimentation significantly.

Any fixes would in either case had to be done in Module repository. With following process and defining such upgrades up front of release it should be maintainable. 

> * Hardware- or environment-specific unit/component test execution will likely require per-module solutions, which makes portability harder.

But in the end we are interested only in complete, valid artifacts from tests. This with combination of running examples  and passed FITs in ref_int is enough for the S-CORE release. 

> * Immediate feedback is reduced. For example, a change in baselibs-main can break a unit test in communication-main, but no-one will notice until baselibs creates a release and communication-main explicitly upgrades. This delays feedback and allows incompatibilities to accumulate silently. In practice, this resembles a form of “big bang integration”, only partially mitigated by feature integration tests. It creates a risk of release-oriented “catch-up integration” rather than continuous full-stack validation.

Agree, this is kind of feedback will be delayed. In practice, I can see lots of modules depending on each other are using git_overrides so maybe we need to introduce tools for continuous integration for the Modules' dependencies? This way we have only 1-2 deps to upgrade if there is new version and check. That also simplifies looking for a root cause of failure, as we know bazel's error messages are often not straight forward.

> * Drift in development dependencies (e.g. test frameworks, tooling, generators) is not addressed.

Dependency resolve check in reference_integration will check also them. Version mismatch will reject integration of the Module.

docs/design_decisions/_assets/DR-008-int-option3_infographic.svg
FScholPer: Can you please also add ther requirement to check against static codeanalysis findings and vulnerabilities: https://github.com/orgs/eclipse-score/security/overview. If they are not be fixed they can' be integrated into the release. Everything here: https://github.com/orgs/eclipse-score/security/overview

FScholPer (changes requested):
as commented vulnerability scans and static findings should be fixed.

docs/design_decisions/DR-008-int.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Cons:
+
+* Additional effort for the Module teams to maintain their tests and dependencies to be fully executable in reference_integration repository
+* Longer integration jobs in reference_integration repository as all tests are executed
+
+Option 3: Front-load quality checks in modules with lightweight reference_integration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Modules follow the S-CORE process and run quality tooling in their own context against their own resolved
```
masc2023, line 97: There are not only tooling to run, also some manual quality checks, e.g. https://eclipse-score.github.io/process_description/main/process_areas/quality_management/quality_workflow.html#wf__exe_featprocess_conformance_checks, https://eclipse-score.github.io/process_description/main/process_areas/quality_management/quality_workflow.html#wf__exe_wp_review, possible automatic measurable outcome to check, https://eclipse-score.github.io/process_description/main/process_areas/quality_management/quality_workproducts.html#wp__qms_report.

Further Release Notes must be available and verified, possible measurable output, are the Module Release Notes as described here, https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workproducts.html#wp__module_sw_release_note, based on https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workflow.html#wf__rel_mod_rel_note, and template here: https://eclipse-score.github.io/process_description/main/folder_templates/modules/module_name/docs/release/release_note.html#doc__module_name_release_note

docs/design_decisions/DR-008-int.rst
```diff
@@ -0,0 +1,204 @@
...
+dependency set. Across modules, teams agree on *what* quality tooling and toolchains to use, but each
+module runs them independently. Most importantly, teams agree on dependency versions for the upcoming
```
masc2023, line 99: Tooling to use must be documented at one place here: https://eclipse-score.github.io/score/main/score_tools/score_tools_evaluation_list.html, These shall contain all tools with the required version. Tool evaluation or classification information is also there and the required qualification report, if version is changed, must be updated

docs/design_decisions/DR-008-int.rst
```diff
@@ -0,0 +1,204 @@
...
+S-CORE release: if a module is releasing a new version into the next S-CORE release, every module that
+depends on it must migrate to that version before the release. Modules minimize non-dev, non-S-CORE
+dependencies in their MODULE.bazel files.
```
masc2023, line 102: How to automate that?

docs/design_decisions/DR-008-int.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Module Owners deliver as part of their release artifacts in the common form allowing aggregation and
+full documentation build in reference_integration. Bazel dependency analysis will allow verification
+that collected artifacts are valid for S-CORE release - resolved dependencies for the Module (in scope of their repository)
+must match the dependencies for S-CORE release and resolved in reference_integration.
+
```
masc2023, line 108: Further checkpoint as @FScholPer , all vulnerabilities in the module folders must be solved

docs/design_decisions/_assets/DR-008-int-option3_infographic.svg
masc2023: As we have to generate also Platform Release Notes, it may an advantage to generate them automatically based on the existing Module Release Notes, if possible. Nevertheless on platform level we have to todo also some final quality check and verification of the Release Notes as described here:
https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workflow.html#wf__rel_platform_rel_note, https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workflow.html#wf__vy_ap_pltrelease and do not forget to update/generate the handbook, https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workflow.html#wf__rel_platform_handbook, quality check https://eclipse-score.github.io/process_description/main/process_areas/release_management/release_workflow.html#wf__rel_platform_handbook, with existing work product, https://eclipse-score.github.io/process_description/main/process_areas/verification/verification_workproducts.html#wp__verification_platform_ver_report

masc2023 (changes requested):

antonkri:
Hey @4og ,

thanks for you proposal. I would have some comments or questions to the Option 2 flaws, that you describe:

**Flaw 1: Non-dev dependencies for quality tooling**
Why do we need to make quality tooling dependencies in S-Core modules non-dev dependencies? These dependencies can be/stay dev dependencies. Imagine the situation, that you integrate S-Core modules into a commercial distribution at some company. There you will want to use quality tooling to ensure, that S-Core modules, that you integrate, fulfill your quality standards. Therefore your will add dependencies to bazel modules, that provide these tools to you and will let those tools run on modules that you integrate from S-Core. Why can not we do the same in the reference integration? In the reference integration we will add dependency to the S-Core quality tooling no matter what the module is using inside it's own repo and will you this tooling to ensure, that the S-Core module, that we integrate, fulfills S-Core process quality requirements. From my opinion, that also what the commercial distribution should do at the end. They should use all the quality tooling that S-Core provides to them and may be add some commercial tools on top. So it should be totally fine to use the quality tooling of it's choice by every component in its own repo as dev-dependency and then to use S-Core quality tooling in the reference integration to ensure compatibility with S-Core dev process. Ideally, the tools that are used by the component in it's own repository and in the reference integration are the same.

**Flaw 2: Not all quality checks can be re-run generically**
Yes, agree, but we need to have a common ground for all modules. A minimal set of checks, e.g. unit-tests, code analysis and so on, that every module fulfills and we can prove it in an automated way. That there will be always some exceptions, that is clear. 

**Flaw 3: Unclear feedback loop**
If with the time all modules use the same tooling and we have a clear understanding of the dependencies we all use, then the integration is just "piece of cake" and all the checks are running smoothly and without problems. I need to add here just a small remark, that in my 10 years as integrator at BMW such situation as this has never happened and integration is always the biggest problem.

At the end "it is not only about bazel" and how it resolves the dependencies. Bazel is just a tool. What me and I think we all should achieve, is an automated workflow, that ensures, that all modules in s-core have a common ground, minimal set of tools, that **automatically** checks, that S-Core modules fulfill the S-Core process. I agree, that you can also achieve this without running all the checks in the reference_integration. But I also do not see currently a clear plan how to do this with Option 3. It just gives some general guidance and some ideas. In my opinion, just relying on the software modules, that they do the right thing and doing the manual reviews from time to time is not an option. May be you could elaborate on the option 3 a little bit more and explain, how it should ensure in the automated way with clear Yes/No gates, that S-Core modules are compliant to S-Core process/tools. I would really appreciate this. Till then I will stay with Option 2 in my mind.

antonkri (changes requested):
see my comments

antonkri:
Hey @4og,

- I've played around on the weekend and see your point with reusing of dev dependencies in the root module.
- I, to be honest, than I do not know, how the integration of the s-core modules into the distributions will look like in the future, as it least the commercial distributions will need to rerun all the checks for the modules internally once again.
- May be (as we've discussed on Friday), let us not decide for the strategy right now and move from the release to the release and gain the knowledge. Let us leave the release v0.7 as it is and for release v0.8 let us understand, what exactly we want to achieve in terms of following the S-Core process and then decide which Option would fit the most for the v0.8 release use-case.

antonkri:
Hey @4og ,

I would like to add some more thoughts on the topic with *dev_dependencies*.  Yes, if you declare some dependencies to the tooling as *dev_dependencies*, then you are not allowed to use bazel targets from the module, that use these tools, in the root module. But, I think, that is not exactly what we want.

I think every module should provide "integration interfaces". That should be the filegroups (or smth. comparable), that can be used during the integration to run tools on them, e.g. a filegroup containing all rst files relevant for documentation and so on (I guess [rules_score](https://github.com/eclipse-score/tooling/blob/c3b0602173e2d375a47b3faf03e8888125f192ba/bazel/rules/rules_score/docs/index.rst) go in this direction). Then we exactly know, which rst files or units-tests are "released" by the module and we can then run the S-Core tooling in the reference_integration repository to ensure, that the module is compliant to the S-Core process. The same filegroups should be used by the modules internally to run the checks using "their tooling", that should be ideally superset of the S-Core tooling.

docs/design_decisions/DR-008-int.rst
anmittag: your comment is fine from process description point of view, for the decision record I guess we now have to come to a point if we continue with option 1, 2 or 3 and then in detailing of the chosen option these findings have to be closed.

anmittag:
**ETAS vote for Option 2:**

Some arguments, mainly drawbacks for Option 3:

dependency handling:  this would require modules to release exactly in the dependency order. If there is another release of e.g. baselibs due to security/bugs/any reasons, then everyone else also needs to release again. This effort is not considered/underestimated. Any changes in lower build order, we have to trigger the builds by following build orders. 
 
quality artifacts: are generated from component pipelines, but the "binary" from that pipeline is not used in release.  But for releases, "binaries" etc. are created from reference int. For safety we have to guarantee we create exact "binaries" from reference int. Argument could be hermetic bazel build -> but there should be some asil qualified tool to validate this, else we have a problem in re-using the quality artifacts
 
difficult to bring in breaking changes: as we have to follow build orders and tests are not executed at a single place. Major co-ordination effort.
 
functionalities: are to be added in a backwards compatible way - but without being able to trigger the builds / tests of other repos that uses the exact code, it is not really possible to say for sure whether the changes done are really backwards compatible.

@antonkri , @FScholPer, @qor-lb , @johannes-esr

docs/design_decisions/DR-008-int.rst
```diff
@@ -0,0 +1,235 @@
...
+     - Medium
+   * - Cross-module coordination overhead
+     - Low
+     - Medium (triage of CI failures)
+     - Medium (shared dependency manifest)
+   * - Traceability for S-CORE release
+     - Weak
+     - Strong
+     - Strong (via release artifacts)
+
```
RolandJentschETAS, line 256: There is impact of safety and security missing.

docs/design_decisions/DR-008-int.rst
```diff
@@ -0,0 +1,235 @@
...
+However, it requires additional effort for the module teams to maintain their tests and dependencies to be fully
+executable in the reference_integration repository with both public APIs and tests.
+It allows generating a single documentation build with full traceability across different repositories which will stay persistent
+and will not require external linking.
+
+Option 3 front-loads quality verification into the modules themselves and keeps reference_integration
+lightweight. It addresses the dependency mismatch risk of Option 1 through coordinated version
+agreements across modules rather than by re-executing checks centrally. However, it relies on
+disciplined inter-module coordination and backward compatibility guarantees, which adds process
+overhead and requires a shared dependency manifest to remain manageable.
```
RolandJentschETAS, line 279: There is a big point missing, that is not only about the module dependencies, but all infrastructure or compiler or linker options. Otherwise the test results inside the module may not more valid, as long as the modules does not deliver as library.

RolandJentschETAS (changes requested):

4og:
As announced in https://github.com/orgs/eclipse-score/discussions/1595#discussioncomment-17059805, we're proposing **Option 4** as a hybrid of Option 2 and Option 3 that tries to tackle their respective drawbacks.

Option 4 is built on the following principles:

- The reference integration is a **blueprint** for S-CORE users (e.g. distributors) integrating S-CORE modules into a safety-certified product.
- It must give confidence that S-CORE modules work together and are safety-certifiable, with a short feedback cycle when compatibility breaks.
- Users may use only a subset of S-CORE modules.
- Users will likely use their own build toolchain, and tests will need to run on hardware.

## How it works

See graphical overview of the workflow here: https://github.com/orgs/eclipse-score/discussions/1595#discussioncomment-17059805 (kudos to @OliverHeilwagen for the great diagram!)

The reference integration maintains a **stable known good**, which is a pinned set of module versions that are known to work together.

When a trigger occurs, such as a new module release candidate\*, a new third-party dependency version, a new dev tool version, or a new toolchain update, a **new candidate** is created by updating one or more versions in the known good set.

\* What exactly a release candidate means needs to be defined. It could be a new version in the Bazel registry or a git hash from the module repository. Ideally, to keep feedback cycles short, module version bumps in the reference integration should happen frequently.

The reference integration pipeline then executes in two stages:

**Stage 1: Integration-scoped checks (in `./reference_integration`):**
1. Build the reference integration for all supported platforms.
2. Run platform integration tests.
3. Run feature integration tests.
4. Retrieve relevant resolved versions including transitive dependencies (e.g. via `bazel mod graph`).
5. Trigger module-scoped builds/tests with the resolved versions (Stage 2 below).

**Stage 2: Module-scoped checks (in each `./module_n`):**

For each module, the pipeline ensures that the module is built and tested against the **dependency versions resolved by the reference integration**. The exact mechanism (e.g. patching `MODULE.bazel` with override directives, injecting a lock file, or other means) is an implementation detail to be worked out, but the goal is the same: the module must use the same dependency set as the full stack. These adjustments are temporary; they exist only for the pipeline run and do not need to be merged into the module repository.

The pipeline then runs the module's component and unit tests in the module's own repository context. Since the reference integration is a blueprint, users would follow a similar workflow but with their own toolchain and on their own hardware.

This step can additionally be **delegated to an automatically created PR** in the module repository, which triggers the module's full CI and makes failures easier to diagnose. I propose we implement this delegation, because reproducing issues locally is harder when the module is checked out only as part of the reference integration workflow.

**Reporting:**

Quality reports are consolidated from both reference integration results and module-scoped results.

### Module repository structure

In contrast to Option 2, each module keeps its `MODULE.bazel` with only **absolute minimum dependencies**. Quality tooling dependencies (e.g. ITF, googletest) remain `dev_dependency = true`.

Each module also provides a "sub-module" that carries minimum promoted dev tooling dependencies and enables **black-box testing** of the module. This is similar in spirit to [Bazel Central Registry test modules](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#test-module). Example in baselibs: https://github.com/eclipse-score/baselibs/tree/main/examples/integration

Whether it makes sense to run these sub-module tests (or "acceptance tests", the naming is open) in the reference integration is up for discussion.

## Quality checks and where they are executed

### In `score/reference_integration` with S-CORE toolchain (Stage 1 above):

* Reference integration build for all supported platforms
* Platform integration tests
* Feature integration tests

### In `score/reference_integration` with a module checked out as Bazel root module, using reference integration toolchain and injected dependencies (Stage 2 above):

* Module component and unit tests

These test executions can additionally be delegated to an automatically created PR in the module repository. This way the module's full CI runs and failures are easier to diagnose.

### In module repositories:

* Build for all supported platforms using the S-CORE toolchain
* Component and unit tests
* Static analysis
* Memory sanitizers
* Code coverage
* Acceptance tests (black-box tests that verify the module is usable as a dependency)
* Manual inspections
* Safety analysis
* Any other module-specific checks (e.g. fuzzing, benchmarks, builds with other toolchains)

## Open questions (up for discussion)

* It is unclear how to consolidate combined documentation in the reference integration. Module documentation would need to be pulled from module releases, but the released documentation may not match the patched dependency versions used during the integration pipeline run.

antonkri:
> As announced in https://github.com/orgs/eclipse-score/discussions/1595#discussioncomment-17059805, we're proposing **Option 4** as a hybrid of Option 2 and Option 3 that tries to tackle their respective drawbacks.
> 
> Option 4 is built on the following principles:
> 
> * The reference integration is a **blueprint** for S-CORE users (e.g. distributors) integrating S-CORE modules into a safety-certified product.
> * It must give confidence that S-CORE modules work together and are safety-certifiable, with a short feedback cycle when compatibility breaks.
> * Users may use only a subset of S-CORE modules.
> * Users will likely use their own build toolchain, and tests will need to run on hardware.
> 
> ## How it works
> See graphical overview of the workflow here: https://github.com/orgs/eclipse-score/discussions/1595#discussioncomment-17059805 (kudos to @OliverHeilwagen for the great diagram!)
> 
> The reference integration maintains a **stable known good**, which is a pinned set of module versions that are known to work together.
> 
> When a trigger occurs, such as a new module release candidate*, a new third-party dependency version, a new dev tool version, or a new toolchain update, a **new candidate** is created by updating one or more versions in the known good set.
> 
> * What exactly a release candidate means needs to be defined. It could be a new version in the Bazel registry or a git hash from the module repository. Ideally, to keep feedback cycles short, module version bumps in the reference integration should happen frequently.
> 
> The reference integration pipeline then executes in two stages:
> 
> **Stage 1: Integration-scoped checks (in `./reference_integration`):**
> 
> 1. Build the reference integration for all supported platforms.
> 2. Run platform integration tests.
> 3. Run feature integration tests.
> 4. Retrieve relevant resolved versions including transitive dependencies (e.g. via `bazel mod graph`).
> 5. Trigger module-scoped builds/tests with the resolved versions (Stage 2 below).
> 
> **Stage 2: Module-scoped checks (in each `./module_n`):**
> 
> For each module, the pipeline ensures that the module is built and tested against the **dependency versions resolved by the reference integration**. The exact mechanism (e.g. patching `MODULE.bazel` with override directives, injecting a lock file, or other means) is an implementation detail to be worked out, but the goal is the same: the module must use the same dependency set as the full stack. These adjustments are temporary; they exist only for the pipeline run and do not need to be merged into the module repository.
> 
> The pipeline then runs the module's component and unit tests in the module's own repository context. Since the reference integration is a blueprint, users would follow a similar workflow but with their own toolchain and on their own hardware.
> 
> This step can additionally be **delegated to an automatically created PR** in the module repository, which triggers the module's full CI and makes failures easier to diagnose. I propose we implement this delegation, because reproducing issues locally is harder when the module is checked out only as part of the reference integration workflow.
> 
> **Reporting:**
> 
> Quality reports are consolidated from both reference integration results and module-scoped results.
> 
> ### Module repository structure
> In contrast to Option 2, each module keeps its `MODULE.bazel` with only **absolute minimum dependencies**. Quality tooling dependencies (e.g. ITF, googletest) remain `dev_dependency = true`.
> 
> Each module also provides a "sub-module" that carries minimum promoted dev tooling dependencies and enables **black-box testing** of the module. This is similar in spirit to [Bazel Central Registry test modules](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#test-module). Example in baselibs: https://github.com/eclipse-score/baselibs/tree/main/examples/integration
> 
> Whether it makes sense to run these sub-module tests (or "acceptance tests", the naming is open) in the reference integration is up for discussion.
> 
> ## Quality checks and where they are executed
> ### In `score/reference_integration` with S-CORE toolchain (Stage 1 above):
> * Reference integration build for all supported platforms
> * Platform integration tests
> * Feature integration tests
> 
> ### In `score/reference_integration` with a module checked out as Bazel root module, using reference integration toolchain and injected dependencies (Stage 2 above):
> * Module component and unit tests
> 
> These test executions can additionally be delegated to an automatically created PR in the module repository. This way the module's full CI runs and failures are easier to diagnose.
> 
> ### In module repositories:
> * Build for all supported platforms using the S-CORE toolchain
> * Component and unit tests
> * Static analysis
> * Memory sanitizers
> * Code coverage
> * Acceptance tests (black-box tests that verify the module is usable as a dependency)
> * Manual inspections
> * Safety analysis
> * Any other module-specific checks (e.g. fuzzing, benchmarks, builds with other toolchains)
> 
> ## Open questions (up for discussion)
> * It is unclear how to consolidate combined documentation in the reference integration. Module documentation would need to be pulled from module releases, but the released documentation may not match the patched dependency versions used during the integration pipeline run.

I like the Option 4 very much, I think this is a way to go. Only two remarks: 
- I would run the sanitizers with feature and platform integration tests. As in complex test scenarios you see more as with simple unit and component tests.
- coverage should be also measured to ensure, that values are comparable (should not be the same) to the ones measured in the module, see: https://github.com/orgs/eclipse-score/discussions/407#discussioncomment-16973453. I think it is needed, in case we mess up with compiler flags and code pathes are totally different

AlexanderLanin:
@4og @OliverHeilwagen the text and image state that the module is patched on the fly pre-release and that state is not preserved and does not flow back to the module. But with that in mind we cannot do module releases, as they have different dependencies during internal analysis?! Or we don't reuse the artifacts generated during module release? Or does this imply we abandon module releases? Unfortunately the "Proposed version rollout mechanism" is a little fuzzy at that point.

OliverHeilwagen:
> d image state that the module is patched on the fly pre-release and that state is not preserved and does not flow back to the module. But with that in mind we cannot do module releases, as they have different dependencies during internal analysis?! Or we don't reuse the artifacts generated during module release? Or does this imply we abandon module releases? Unfortunately the "Proposed version rollout mechanism" is a little fuzzy at tha

@AlexanderLanin 
Module releases shall not be abandoned.
I doubt reuse of patched MODULE.bazel files for module releases.

Here are the rationales:
- External projects may pick only 1 module (+ it's dependencies) from S-CORE. Modules must therefore be releasable on their own.
- The MODULE.bazel file may contain:
   - patch with git_override(s) for release candidate(s) (probing - not module release relevant)
   - patch with git_override(s) for release versions (we should avoid git_override in module releases, refer to last point)
   - patch for dedicated version(s) of transient dependency/ies (version resolution in broader context - not module release relevant)
   - no patch (already aligned version - rollout of a successful tested release candidate e.g. via dependabot)

The need for .bazelrc patching can hopefully be avoided in S-CORE.

4og:
@antonkri,
Running the sanitizer does indeed make sense, but I'm not sure about the coverage. Would it be sufficient to ensure that all requirements are covered by tests and that these tests are fully executed?

antonkri:
@4og 
- I think, that measurement of the coverage was requested by the process community (check the link I've put in the previous comment).
- I think requirements coverage is not sufficient. I think, the argumentation is in general the same, as if it would be why you need unit-test coverage at all if you have 100% requirements coverage. In distribution/reference integration you have different compiler settings, toolchain and so on and you can not be sure, that code is executed the same way and therefore you need to ensure, that your unit-test coverage is still good. The SEooC argument doesn't work here, as you will need to produce SEooC for every toolchain you component could be used in.

OliverHeilwagen:
@antonkri
The goal in S-CORE should be that there are no differences w.r.t. toolchain and compiler options between the modules & reference integration. We should have a common "toolchain section" in MODULE.bazel + .bazelrc.

All modules should depend on the same toolchains (score_toolchains_rust, score_bazel_cpp_toolchains) and corresponding .bazelrc settings. Usage of --copt and --cxxopt in .bazelrc should be prohibited and only allowed on bazel targets if necessary.

The module coverage result (with patched MODULE.bazel) shall be a valid result for the reference integration context.
Open points from my side are:
- toolchain feature switches (e.g. --features=-strict_warnings)
- sdp version

Hopefully we can avoid the need for patching the module .bazelrc file in S-CORE.

In projects that depend on S-CORE modules, it is expected that patching of module .bazelrc for custom toolchains will be necessary. This could also be the reason why such a patching mechanism is showcased in S-CORE.

PiotrKorkus:
DR was updated with Option 4. Please update your review feedbacks.
Approve = agree to continue with Option 4.

Deadline for the voting is this Thursday TLC meeting ( 14:00 CET)

anmittag (approved):

anmittag:
Option 4 is the consolidated one now :-) finally

4og (approved):

pawelrutkaq (approved):

NicolasFussberger (approved):

antonkri (approved):

FScholPer (approved):
Okay for me but we should clear define in the requirements what does that mean / how that should be implemented

RolandJentschETAS (approved):

docs/design_decisions/DR-008-int.rst
masc2023: Documented here
https://github.com/eclipse-score/score/issues/2977

masc2023 (approved):
For selection of the option only, fine, further details need to be added, linked here https://github.com/eclipse-score/score/issues/2977

PR #2762, opened 2026-04-08
Add version to Docs-as-code

MaximilianSoerenPollak:
Docs-as-code has to have a version for consumer tests to work.  Each 'bazel_dep' should **always** have a version.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -73,7 +73,7 @@ bazel_dep(name = "rules_java", version = "8.15.1")
 #
 ###############################################################################
 bazel_dep(name = "score_tooling", version = "1.1.2")
-bazel_dep(name = "score_docs_as_code")
+bazel_dep(name = "score_docs_as_code", version = "3.0.1")
 git_override(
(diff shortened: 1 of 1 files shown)


AlexanderLanin:
how did that work without a version??

MaximilianSoerenPollak:
> how did that work without a version??

Git override makes it possible I think.

FScholPer (approved):

AlexanderLanin:
oh, so now you specifiy a version and 1 line later you override it. That's also a little strange.

MaximilianSoerenPollak:
> oh, so now you specifiy a version and 1 line later you override it. That's also a little strange.

I guess that indeed is also true.

AlexanderLanin:
The advantage if it's not specified: the user (i.e., one level up) doesn't get a random error due to an incorrect docs-as-code version before the override, but instead gets the specific message “specify a version, for crying out loud.”

MaximilianSoerenPollak:
Will adapt the consumer tests instead and leave this as is to not introduce further issues.

PR #2767, opened 2026-04-09
Move Dockerfile one level up

pahmann:
I am actually not sure if it was by intention to have the .devcontainer with the Dockerfile under .devcontainer so a nested folder structure.

I got an error message in vscode while rebuilding, saw the move to the lower level folder, reverted it, and came back to a working container.

Just close the PR, in case I am doing something wrong here.


AlexanderLanin (approved):
Looks like the devcontainer is not used / needed in this repo, otherwise someone would have noticed. Fix looks correct.

pahmann:
> Looks like the devcontainer is not used / needed in this repo, otherwise someone would have noticed. Fix looks correct.

😅 I use it.

Thanks for merging.

PR #2814, opened 2026-04-17
docs: c++ compiler is required to work with bazels pip integration

AlexanderLanin:
As discussed in #infrastructure a C++ compiler must be installed on the host system, otherwise compile_pip_requirements is not working.

### DIFF
--- a/docs/contribute/development/development_environment.rst
+++ b/docs/contribute/development/development_environment.rst
@@ -61,8 +61,9 @@ General Setup
 * Install the required dependencies for your operating system. This includes:
 
   * Bazelisk
-  * Python 3.x
-  * Graphviz/Dot (for generating diagrams)
+  * Python 3.x (due to bugs in Bazel, which requires Python to be present on the system)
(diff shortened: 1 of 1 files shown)


FScholPer (approved):

docs/contribute/development/development_environment.rst
```diff
@@ -61,8 +61,9 @@ General Setup
 * Install the required dependencies for your operating system. This includes:
 
   * Bazelisk
-  * Python 3.x
-  * Graphviz/Dot (for generating diagrams)
+  * Python 3.x (due to bugs in Bazel, which requires Python to be present on the system)
+  * Graphviz/Dot (for generating diagrams, as it cannot be provided via Bazel)
+  * C++ compiler like gcc (due to bugs in Bazel's Python rules, which require a C++ compiler to be present on the system)
```
limdor, line 66: For things like this it would be great if links to the bugs are attached, in this way we can follow up on them and potentially revisit the decision once the bugs are resolved.

PR #2820, opened 2026-04-20
Baselibs component DFA

aschemmel-tech:
Refers: eclipse-score/baselibs#239


4og (changes requested):
After syncing with @aschemmel-tech we concluded that DFA should be done per library, so statements like below are misleading.

> As the component's archtitecture does not have a decomposition into further components,
there are no failures additional to the ones analyzed on feature level.

aschemmel-tech:
As the respective work products were moved to baselibs repository, this PR closed and cloned into https://github.com/eclipse-score/baselibs/pull/268

PR #2825, opened 2026-04-22
update folder structure to DR001 (independent module repos)

RolandJentschETAS:
update folder structure to DR001 (independent module repos).

See https://eclipse-score.github.io/score/main/design_decisions/DR-001-proc.html

### DIFF
--- a/docs/contribute/general/folder.rst
+++ b/docs/contribute/general/folder.rst
@@ -124,7 +124,7 @@ See also `module template repository <https://github.com/eclipse-score/module_te
                                              Folder containing all artifacts corresponding to one component.
          src/                          -> Source files of the component consisting of
                                              Include files
-                                             Source Files
+                                             Source Files [:need:`wp__sw_implementation`]
                                              Unit tests [:need:`wp__verification_sw_unit_test`]
                                              Test doubles
             <lower_level_comp>/        -> lower level component following <component_name> folder structure
@@ -135,7 +135,7 @@ See also `module template repository <https://github.com/eclipse-score/module_te
             architecture/              -> Component architecture (only if lower level components exist)
                                              [:need:`wp__component_arch`].
             detailed_design/           -> Detailed Design [:need:`wp__sw_implementation`] and
-                                          code inspection [:need:`wp__sw_implementation_inspection`],
+                                          Detail design inspection [:need:`wp__sw_implementation_inspection`],
             requirements/              -> Component requirements [:need:`wp__requirements_comp`],
                                              requirements inspection [:need:`wp__requirements_inspect`]
             safety_analysis/           -> Safety analysis on component level (only if component architecture exists)
--- a/docs/contribute/general/folder.rst
+++ b/docs/contribute/general/folder.rst
@@ -60,8 +60,8 @@ The following shows the folder structure of the platform repository (ordered alp
         requirements/                  -> Requirements on platform level
             stakeholder/               -> Stakeholder requirements [:need:`wp__requirements_stkh`]
             tool/                      -> Tool requirements (not covered in tool repositories)
-            platform_assumptions/      -> Assumptions of use on platform level [:need:`wp__requirements_sw_platform_aou`]
                                              [:need:`wp__requirements_proc_tool`]
+            platform_assumptions/      -> Assumptions of use on platform level [:need:`wp__requirements_sw_platform_aou`]
         safety/                        -> Safety documentation on platform level (SEooC):
                                              [:need:`wp__platform_dfa`], [:need:`wp__platform_safety_manual`],
                                              [:need:`wp__platform_safety_package`], [:need:`wp__fdr_reports`],
--- a/docs/contribute/general/folder.rst
+++ b/docs/contribute/general/folder.rst
@@ -91,8 +91,9 @@ See also `module template repository <https://github.com/eclipse-score/module_te
 .. parsed-literal::
 
    <module_name>/                      -> Folder containing all artifacts corresponding to one module.
+                                             As folder optional if the repository only contains a single module.
       docs/                            -> Documentation of the module
-        features/                      -> All features of the platform
+        features/                      -> All features of the module.
             <feature_name>/            -> Features including sub-folders and feature/component (change) request
                                              [:need:`wp__feat_request`], [:need:`wp__cmpt_request`]
                architecture/           -> Feature architecture [:need:`wp__feature_arch`], [:need:`wp__sw_arch_verification`]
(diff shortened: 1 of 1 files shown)


docs/contribute/general/folder.rst
```diff
@@ -27,15 +27,8 @@ The following shows the folder structure of the platform repository (ordered alp
     docs/                              -> Global documentation of the platform
         contribute/                    -> Description on how to contribute
         features/                      -> All features of the platform
-            <feature_name>/            -> Features including sub-folders and feature (change) request [:need:`wp__feat_request`]
-               architecture/           -> Feature architecture [:need:`wp__feature_arch`]
```
aschemmel-tech, line 31: I think we should also have a folder to put in the "Logical Interface" needs. So keep "architecture" or rename to "interface"

docs/contribute/general/folder.rst
```diff
@@ -27,15 +27,8 @@ The following shows the folder structure of the platform repository (ordered alp
     docs/                              -> Global documentation of the platform
         contribute/                    -> Description on how to contribute
         features/                      -> All features of the platform
```
aschemmel-tech: align with current use and add "design_decisions" here

docs/contribute/general/folder.rst
```diff
@@ -27,15 +27,8 @@ The following shows the folder structure of the platform repository (ordered alp
...
+            <feature_name>/            -> Features including feature id, logical interfaces and feature (change) request [:need:`wp__feat_request`] and belonging sub-folders
                requirements/           -> Feature requirements [:need:`wp__requirements_feat`]
-               safety_analysis/        -> Safety analysis on feature level [:need:`wp__feature_fmea`], [:need:`wp__feature_dfa`]
-               safety_planning/        -> Feature specific safety workproducts planning
-               security_analysis/      -> Security analysis on feature level [:need:`wp__feature_security_analysis`]
-               verification/           -> Feature verification, consisting of Architecture review [:need:`wp__sw_arch_verification`],
-                  integration_tests/   -> Feature Integration tests [:need:`wp__verification_feat_int_test`]
-
         glossary/                      -> Glossary of abbreviations used in the platform context
         introduction/                  -> Background, scope and high-level architecture of the platform features
```
aschemmel-tech: align with current use and add "handbook" here

docs/contribute/general/folder.rst
aschemmel-tech: should remove "modules" here now

aschemmel-tech, line 65: add wp__requirements_sw_platform_aou here?

docs/contribute/general/folder.rst
```diff
@@ -69,47 +62,53 @@ Module Folder Structure
...
+                                          code inspection [:need:`wp__sw_implementation_inspection`],
             requirements/              -> Component requirements [:need:`wp__requirements_comp`]
             safety_analysis/           -> Safety analysis on component level (only if component architecture exists)
                                              [:need:`wp__sw_component_fmea`], [:need:`wp__sw_component_dfa`]
             security_analysis/         -> Security analysis on component level (only if component architecture exists)
-                                             [:need:`wp__sw_component_security_analysis`]
-            verification/              -> Component verification, consisting of Architecture review [:need:`wp__sw_arch_verification`],
-                                          code inspection [:need:`wp__sw_implementation_inspection`],
-               integration_tests/      -> Component integration tests [:need:`wp__verification_comp_int_test`]
-            <lower_level_comp>/        -> Lower level component following <component_name> folder structure
```
aschemmel-tech, line 115: need to keep also lower_level_comp docs folders

docs/contribute/general/folder.rst
```diff
@@ -69,47 +62,53 @@ Module Folder Structure
...
-            details/                   -> Private implementation details of the component following <component_name> folder structure
-
+            tests/                     -> Component-level tests (e.g., unit tests) [wp__verification_sw_unit_test]
+         tests/                        -> Module-level tests (e.g., feature integration tests, system tests) [wp__verification_comp_int_test]
+                                       -> Feature Integration tests [:need:`wp__verification_feat_int_test`]
          docs/                         -> Documentation of the component
             architecture/              -> Component architecture (only if lower level components exist)
                                              [:need:`wp__component_arch`].
             detailed_design/           -> Detailed Design [:need:`wp__sw_implementation`]
+                                          code inspection [:need:`wp__sw_implementation_inspection`],
```
aschemmel-tech: detailed design & code inspection

aschemmel-tech (changes requested):
some findings inline

PandaeDo (changes requested):
I would appriciate if we can complete the definition to all of the wp's that are defined in the processes. This would avoid that they where stored somewhere. And we also can check if somebody is doing so. Currently we have 74 wp's and 31 are not explicitly named.

RolandJentschETAS:
> I would appriciate if we can complete the definition to all of the wp's that are defined in the processes. This would avoid that they where stored somewhere. And we also can check if somebody is doing so. Currently we have 74 wp's and 31 are not explicitly named.

Good idea.

aschemmel-tech:
> I would appriciate if we can complete the definition to all of the wp's that are defined in the processes. This would avoid that they where stored somewhere. And we also can check if somebody is doing so. Currently we have 74 wp's and 31 are not explicitly named.

Did not quite get the point. Can you give an example of a wp which is not "named"? Do you mean every wp should be mentioned in this folder structure? Then I agree.

PandaeDo:
> Did not quite get the point. Can you give an example of a wp which is not "named"? Do you mean every wp should be mentioned in this folder structure? Then I agree.

Yes. All wp should be mentioned in this folder structure.

docs/contribute/general/folder.rst
RolandJentschETAS: Added

docs/contribute/general/folder.rst
RolandJentschETAS: changed

docs/contribute/general/folder.rst
RolandJentschETAS: I have added an "and". Is this enough ?

docs/contribute/general/folder.rst
RolandJentschETAS: Added again with some comments

docs/contribute/general/folder.rst
RolandJentschETAS: Not sure. This is the entry point for the jump page to the modules in the platform. Does we move it to the reference integration ?

RolandJentschETAS: done

docs/contribute/general/folder.rst
RolandJentschETAS: No. This repeats in the src folder of the lower level component itself, because there is it resembled.

RolandJentschETAS:
> > Did not quite get the point. Can you give an example of a wp which is not "named"? Do you mean every wp should be mentioned in this folder structure? Then I agree.
> 
> Yes. All wp should be mentioned in this folder structure.

Have now updated the list with all of the work packages.

docs/contribute/general/folder.rst
aschemmel-tech: Ah, I see. So content is only a list with link to the "official" module repositories documentation. OK

docs/contribute/general/folder.rst
aschemmel-tech: ok, I see that might then also be the reason for renaming src to score.

docs/contribute/general/folder.rst
```diff
@@ -69,47 +86,59 @@ Module Folder Structure
...
-                                             module securty package [:need:`wp__module_security_package`],
+                                             module security package [:need:`wp__module_security_package`],
                                              formal documents reviews [:need:`wp__fdr_reports_security`],
                                              module SW bill of material [:need:`wp__sw_module_sbom`]
          verification_report/          -> Module verification report
-                                             components verifications [:need:`wp__verification_module_ver_report`],
-                                             safety analysis formal reviews [:need:`wp__fdr_reports`]
-
+                                             module verifications [:need:`wp__verification_module_ver_report`],
+   score/                              -> Folder containing all source code artifacts corresponding to the module.
```
aschemmel-tech: It contains more than source code artefacts. Better call "Folder containing all artifacts corresponding to all components of the module."

docs/contribute/general/folder.rst
aschemmel-tech: Ah, sorry misunderstanding: what I meant is: "detailed design inspection & code inspection" which are both in wp__sw_implementation_inspection

aschemmel-tech (changes requested):
re-review: some comments resolved, one new found

docs/contribute/general/folder.rst
RolandJentschETAS: Done.

docs/contribute/general/folder.rst
RolandJentschETAS: That means wp__sw_implementation is more in src folder (line 125) ?

aschemmel-tech: wp_sw_implementation is code and detailed design - because originally we expected that the detailed design would be part of the source code files (e.g. doxygen comments). But now it is defined as rst for the detailed design (manually filled) - I do not know if it stays like this.

aschemmel-tech: and wp__sw_implementation_inspection is one template/document for detailed design and code - we should not split this up.

RolandJentschETAS: Changed

docs/contribute/general/folder.rst
```diff
@@ -69,47 +86,59 @@ Module Folder Structure
...
+                                             module security package [:need:`wp__module_security_package`],
                                              formal documents reviews [:need:`wp__fdr_reports_security`],
                                              module SW bill of material [:need:`wp__sw_module_sbom`]
          verification_report/          -> Module verification report
-                                             components verifications [:need:`wp__verification_module_ver_report`],
-                                             safety analysis formal reviews [:need:`wp__fdr_reports`]
-
+                                             module verifications [:need:`wp__verification_module_ver_report`],
+   score/                              -> Folder containing all artifacts corresponding to all components of the module.
       <component_name>/                -> Components of the module.
```
PandaeDo: I would find it a litte bit easier to use the same structure. We have docs/feature/<feature_name>/ and <module_name>/docs/

docs/contribute/general/folder.rst
```diff
@@ -69,47 +86,59 @@ Module Folder Structure
...
                                              Folder containing all artifacts corresponding to one component.
-
          src/                          -> Source files of the component consisting of
```
PandaeDo: Might switch this with docs to keep alphabetial order

docs/contribute/general/folder.rst
```diff
@@ -69,47 +86,59 @@ Module Folder Structure
...
                                              Include files
-                                             Source Files
+                                             Source Files [:need:`wp__sw_implementation`]
                                              Unit tests [:need:`wp__verification_sw_unit_test`]
                                              Test doubles
             <lower_level_comp>/        -> lower level component following <component_name> folder structure
```
PandaeDo: Might switch this with details to keep alphabetial order

PandaeDo (changes requested):
Some small additional findings and why you don't add all of the workproducts to the folder structure? I think the definition could really be helpful for the moment until we are not able to tool based solution that you can store it whereever (the computer) want's it.

docs/contribute/general/folder.rst
RolandJentschETAS: Done.

docs/contribute/general/folder.rst
RolandJentschETAS: Done

docs/contribute/general/folder.rst
RolandJentschETAS: I do not understand ...  You mean because "docs" is on module level first ? Or you mean the opposite of platform structure and module structure ?

<module_name> is only optional if (by whatever reason) you have two modules in a repo. Never seen this use case, but I led it open. Normally <module_name> is empty and therefore it starts from root. So in the platform it starts with /docs and in the module as well with /docs. On both places then we go on with /docs/features for the feature documentation.

PandaeDo (approved):
LGTM, me findings are solved/discussed.

aschemmel-tech (approved):
comments resolved

PR #2841, opened 2026-04-27
Bitmanipulation comp FMEA

aschemmel-tech:
Refers: eclipse-score/baselibs#239


aschemmel-tech:
need to think why https://eclipse-score.github.io/score/main/modules/baselibs/bitmanipulation/docs/requirements/index.html#aou_req__bitmanipulation__type_constraints is not covered

aschemmel-tech:
moved to https://github.com/eclipse-score/baselibs/pull/262

PR #2899, opened 2026-05-15
add design decision static versus dynamic libraries

RolandJentschETAS:


docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+| Single application on ECU | Static — no sharing benefit, simpler safety case |
+| 2–5 applications sharing the module | Dynamic may be beneficial — evaluate RAM/flash savings vs. ABI management cost |
+| 5–20+ applications sharing the module | Dynamic strongly recommended — significant memory/flash/update benefits |
+| Hard real-time single-purpose controller | Static — maximum determinism, no dynamic linking overhead |
+| High-performance compute with OTA requirements | Dynamic — update granularity and memory sharing outweigh the overhead |
+
+S-CORE provides both variants for opted-in modules. The integrator's deployment configuration (which variant to link) is outside S-CORE's scope but should be documented in S-CORE's integration guide with guidance for typical scenarios.
+
+### Special Case: Infrastructure Libraries (baselibs)
+
```
pawelrutkaq, line 222: what about com ? almost each app will use com and com looks huge in binary size ;)

docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+| **Breakpoints** | Straightforward — all code addresses fixed at link time | Works, but lazy-bound PLT entries may require `set breakpoint pending on` or break on the resolved target |
+| **Source-level stepping** | Seamless if debug info (`-g`) is included | Seamless if debug info is available per `.so` — identical experience once symbols are loaded |
+| **Core dump analysis** | Self-contained — the single binary plus the core file is sufficient to reconstruct the full state | Requires the **exact matching `.so` versions** alongside the core file. If `.so` files have been updated since the crash, stack traces may be incorrect |
+| **Stripped production binaries** | External debug info (`.debug` file) must match the deployed binary | External debug info must match **per `.so`** — requires tracking debug symbols for each library version independently |
+| **Hot code reload during development** | Not possible — must relink and restart the entire binary | Possible in principle (`dlclose` / `dlopen`) — useful for rapid iteration on a single module without full restart |
```
rmaddikery: not sure if this is even practical.hot code reload is not safe for c++ shared libs with virtual interfaces. dlclose unpas code pages, any liv objects with vtable pointers into that .so becomes use-after-free - hence subsequent virtual calls jump to unmapped memory. Either remove this OR add a hint/warning that dlclose() must never be used when objects from the .so are still alive.

docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+| Breaking API change | Consumer recompiles — detected at build time | Consumer may crash at runtime if `.so` is updated without recompile |
+| ABI versioning | Not applicable | Soname versioning required (e.g., `libscore_comm.so.1` → `.so.2`) |
+| Backward compatibility | Not required | Must be maintained within a major version |
+| Support lifetime | Per release | Per ABI version — may span multiple releases |
+
+**Key questions for S-CORE:**
+1. **How long is an ABI version supported?** If S-CORE guarantees ABI stability for 2 years, consumers can update middleware `.so` files without rebuilding. If ABI stability is not guaranteed, the benefit of independent updateability is reduced.
+2. **How are breaking changes communicated?** Soname bumps signal ABI breaks. Consumers linking against `libscore_comm.so.1` will not accidentally load `libscore_comm.so.2` — the dynamic linker rejects the mismatch.
+3. **Tooling:** ABI compliance checking tools (`abidiff`, `abi-compliance-checker`) can be integrated into CI to detect unintentional ABI breaks before release. Bazel can be configured to run these checks as part of the build.
```
rmaddikery: abidiff alone is not enough. For modules where C++ types cross the .so boundary (which is the common  ofcourse unless a pure C ABI is enforced), toolchain identity is the atmost importance.

docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+
+- The default case (static) remains simple for safety argumentation, build system, debugging, and multi-repo integration.
+- Modules that benefit from dynamic delivery can opt in without forcing complexity on the rest of the platform.
+- OEMs gain targeted update granularity for the most critical middleware components while maintaining deterministic behavior for the rest.
+- Cross-language module delivery is cleanly supported where needed.
+- LGPL compliance is addressed structurally rather than through workarounds.
+
+### Negative / Costs
+
+- Opted-in modules carry additional overhead: dual build targets, ABI management, Soname versioning, ABI compliance CI checks, and deployment manifest generation.
```
rmaddikery: should also include a toolchain identity CI gate alongside abidiff

docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+| Initialization order | Determined by link order — well-defined within one binary | Determined by dependency graph + load order — may vary across runs or platform versions |
+| Circular dependencies | Detected at link time (error) | May cause undefined initialization order or `dlopen` failures |
+| Global state in constructors | Executes once in binary startup — predictable | May execute at different times relative to other libraries |
+
+**Risks:**
+- If S-CORE module A's constructor depends on module B being initialized first, the dynamic linker must respect this dependency order. ELF `DT_NEEDED` entries establish this, but complex graphs may lead to surprising behavior.
+- Constructor-heavy designs are fragile with dynamic linking. Best practice: minimize constructor logic, use explicit initialization functions called by the application.
+
+**Recommendation for S-CORE:**
+- Avoid reliance on `__attribute__((constructor))` for critical initialization.
```
rmaddikery: Additionally .so constructors should not call into the host module. Must not modify global process state, not spawn threads.

docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+
+The linking model affects what constitutes a "tested configuration" and who is responsible for testing which combinations:
+
+| Aspect | Static | Dynamic |
+|--------|:------:|:-------:|
+| **Unit/module test artifact** | Static library tested in isolation (test binary links statically) | Same — module tests are independent of delivery form |
+| **Integration test artifact** | The final linked binary = test subject | The `.so` + a test harness binary = test subject |
+| **Deployment configuration tested** | Binary is self-contained — no deployment variability | Specific combination of `.so` versions must be tested together |
+| **Combinatorial explosion** | Each consumer binary is one test configuration | Each combination of consumer × `.so` versions is a configuration |
+| **Who tests?** | Consumer is responsible for their binary | Middleware provider tests the `.so` in reference configurations; consumer tests their specific integration |
```
rmaddikery: ensure ABI stability guarantee and CI tooling

docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+
+### Thread Safety and Lazy Binding
+
+The default ELF dynamic linker behavior uses **lazy binding**: symbols are resolved on first call rather than at load time. This interacts with multi-threading:
+
+| Scenario | Behavior |
+|----------|----------|
+| Single-threaded first call | Lazy resolution triggers linker code → safe |
+| Multi-threaded first call to same symbol | Multiple threads may trigger resolution simultaneously → implementation-dependent |
+| `LD_BIND_NOW` / `-Wl,-z,now` (eager binding) | All symbols resolved at load time before any application thread runs → fully deterministic |
```
rmaddikery: Should incldue RTLD_NOW | RTLD_LOCAL with dlopen
https://pubs.opengroup.org/onlinepubs/009695399/functions/dlopen.html

docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+| Core dump with `.so` build-IDs | Yes (well-supported by GDB, coredumpctl) | Supported (QNX crash dump system) |
+
+#### Linux-Specific Considerations
+
+- **Mature ecosystem:** Dynamic linking is the dominant model on Linux. The toolchain (GCC, LLVM, ld, ldd, ldconfig) is highly optimized for shared libraries. Package managers, distribution models, and most middleware on Linux assume dynamic linking.
+- **musl libc and static linking:** For fully static binaries on Linux, `musl` (as alternative to glibc) provides clean static linking without hidden dynamic dependencies. Bazel supports musl-based static toolchains. This can simplify deployment on embedded Linux targets.
+- **glibc and `dlopen` within static binaries:** glibc uses `dlopen` internally (e.g., for NSS, iconv). Fully static binaries with glibc may behave unexpectedly. This is a known limitation — not a concern for QNX.
+- **ASLR:** Both PIE (static) and PIC (shared) binaries support full ASLR on Linux. No security difference between the models regarding address randomization.
+
+#### QNX-Specific Considerations
```
rmaddikery: Needs to be extended with security specific reqs for hardening

docs/design_decisions/DR-002-arch.md
```diff
@@ -0,0 +1,847 @@
...
+|-----------|:--------------------:|:--------------------:|
+| Static linking | Fully supported | Fully supported |
+| Shared libraries (`.so`) | Fully supported (ELF, `ld-linux.so`) | Fully supported (ELF, `ldqnx.so`) |
+| `dlopen` / `dlclose` (runtime loading) | Fully supported | Supported |
+| Page sharing of `.text` across processes | Yes (kernel maps shared `.so` pages) | Yes (procnto shares mapped pages) |
+| Position-Independent Code (PIC) required for `.so` | Yes | Yes |
+| `LD_PRELOAD` / library interposition | Yes — potential security concern | Limited — QNX restricts in secure configurations |
+| Full RELRO hardening | Supported (`-Wl,-z,relro,-z,now`) | Supported |
+| Symbol versioning (ELF version scripts) | Fully supported | Supported (limited ecosystem tooling) |
+| `DT_RUNPATH` / `DT_RPATH` | Fully supported | Supported |
```
rmaddikery: Note that RUNPATH can be overridden by LD_LIBRARY_PATH. Ensure RPATH is used

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: COM is like the other modules and is a very good candidate for a dynamic library. So there is no need for a special chapter. nevertheless baselibs is a special one, because of hie multiple APIs and template/header delivery.

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: But that's for dynamic loading at runtime. Not sure if this appropriate for any safety certified libraries/applications. Might be, that for QM it is fine.

RolandJentschETAS: Added below

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: Added

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: Added

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: Added ...

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: Added below as link/compiler options

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: Added

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: Added

docs/design_decisions/DR-002-arch.md
RolandJentschETAS: Added

opajonk (commented):
I read the documents and I must say: great work, a very detailed assessment of static vs. dynamic libraries.

I would have had one point in favor of static linking, purely in case of processes which execute safety workloads - but I think this is not strictly necessary, the relevant arguments are already present.

As an example, in EB corbos Linux for Safety Applications, static linking for so-called "high-integrity" applications (i.e. applications which carry safety workloads) is at least as we speak an assumption of use; in addition to all the rationales already given here, it is a surprisingly complex issue to even determine if an application was loaded **definitely correctly** when a dynamic linker is involved. Think now of partially updating shared objects for such an application - it is getting really hard to argue that is a good idea. So I definitely support the decision of "static library" (which from my PoV is a **must** for safety), plus shared library opt-in (which is totally fine for QM applications).

RolandJentschETAS:
> I read the documents and I must say: great work, a very detailed assessment of static vs. dynamic libraries.
> 
> I would have had one point in favor of static linking, purely in case of processes which execute safety workloads - but I think this is not strictly necessary, the relevant arguments are already present.
> 
> As an example, in EB corbos Linux for Safety Applications, static linking for so-called "high-integrity" applications (i.e. applications which carry safety workloads) is at least as we speak an assumption of use; in addition to all the rationales already given here, it is a surprisingly complex issue to even determine if an application was loaded **definitely correctly** when a dynamic linker is involved. Think now of partially updating shared objects for such an application - it is getting really hard to argue that is a good idea. So I definitely support the decision of "static library" (which from my PoV is a **must** for safety), plus shared library opt-in (which is totally fine for QM applications).

For me dynamic loading is not contradicting with safety. It is depending on the ASIL Level and safety measures which are involved. For example, if the symbol loading occurs at startup time, then a supervision of the startup is enough to ensure ABI compatibility. (See the chapter for that:  https://eclipse-score.github.io/score/pr-2899/design_decisions/DR-002-arch/considerations.html#thread-safety-and-lazy-binding) A drift of the functionality could be ruled out by a if the libraries are qualified and tested in the full system or if the safety and ffi analysis with according measures have analysed and considered that. There are multiple projects which have an ASIL B rating and dynamic libraries in the field.

qor-lb (approved):

PR #2933, opened 2026-05-27
fix: update safety attribute

a-zw:
feat_req__baselibs__core_utilities must be ASIL-B because feat_arc_sta__baselibs__static_view_arch and
feat_arc_dyn__baselibs__dynamic_view_arch fulfill it and these are ASIL-B.

Same with feat_req__persistency__variant_management and feat_arc_sta__persistency__static.

### DIFF
--- a/docs/features/baselibs/docs/requirements/index.rst
+++ b/docs/features/baselibs/docs/requirements/index.rst
@@ -26,7 +26,7 @@ Requirements
    :id: feat_req__baselibs__core_utilities
    :reqtype: Functional
    :security: NO
-   :safety: QM
+   :safety: ASIL_B
    :satisfies: stkh_req__functional_req__base_libraries
    :status: valid
 
--- a/docs/features/persistency/requirements/index.rst
+++ b/docs/features/persistency/requirements/index.rst
@@ -49,7 +49,7 @@ Requirements
    :id: feat_req__persistency__variant_management
    :reqtype: Non-Functional
    :security: NO
-   :safety: QM
+   :safety: ASIL_B
    :satisfies: stkh_req__overall_goals__variant_management
    :status: valid
 


4og (approved):

docs/features/baselibs/docs/requirements/index.rst
```diff
@@ -26,7 +26,7 @@ Requirements
    :id: feat_req__baselibs__core_utilities
    :reqtype: Functional
    :security: NO
-   :safety: QM
+   :safety: ASIL_B
```
aschemmel-tech, line 29: Requirements inherit the ASIL level from their parents and not from the ASIL level of their feature architecture, so the argumentation for changing this to "ASIL B" is not correct. Feature architecture can show components of different ASIL level (if the components provide FFI) and still be classified the higest level of these. But as the Utils Library component requirements are also ASIL B, this change is ok.

docs/features/persistency/requirements/index.rst
```diff
@@ -49,7 +49,7 @@ Requirements
    :id: feat_req__persistency__variant_management
    :reqtype: Non-Functional
    :security: NO
-   :safety: QM
+   :safety: ASIL_B
```
aschemmel-tech, line 52: Requirements inherit the ASIL level from their parents and not from the ASIL level of their feature architecture, so the argumentation for changing this to "ASIL B" is not correct. Feature architecture can show components of different ASIL level (if the components provide FFI) and still be classified the higest level of these. Unclear here what is correct ASIL level, because feat_req__persistency__variant_management is currently not fulfilled by a persistency component.

aschemmel-tech (changes requested):

docs/features/persistency/requirements/index.rst
PandaeDo, line 52:
```suggestion
   :safety: QM
```
Agree on Alexanders argumentation. We might consider the check when it's not covering this variant.

PandaeDo (changes requested):

PR #2968, opened 2026-06-11
increase docs_as_code to version 4.3

RolandJentschETAS:
This pull request updates a dependency version in the `MODULE.bazel` file. The only change is upgrading the `score_docs_as_code` dependency from version 4.2.0 to 4.3.0.


AlexanderLanin:
overlaps #2969 🤣

PR #2983, opened 2026-06-15
Add satisfied_by requirement links

aschemmel-tech:
Refers: eclipse-score/process_description#418


aschemmel-tech:
this is the next step after https://github.com/eclipse-score/score/pull/2978

PR #2986, opened 2026-06-15
chore(docs/cicd): make docs workflow secure

AlexanderLanin:
Use new secure docs workflow split, where anything running bazel is treated as untrusted PR context. And that artefact is later picked up by a secure workflow running from main branch (therefore it will not work in this PR, similar to pull_request_target).

As a little byproduct, this makes score repo ready for merge queues! One of the first merge queue capable repos. As everything PR-related is now handled by on-pr workflow!


lurtz (commented):
in [inc_someip_gateway](https://github.com/eclipse-score/inc_someip_gateway/blob/5e336fabbf45f060e5bc7618b220fd7406812ff8/.github/workflows/ci.yml#L42-L67) I used a meta job, which depends on all the required ones and checks their status. Then you only need to configure that one as required in Github / Otterdog. More on that can be found at the [pull request](https://github.com/eclipse-score/inc_someip_gateway/pull/54).

So far the approach in inc_someip_gateway worked nicely. I never had to touch the .eclipsefdn repo again.

IMHO we should consolidate the patterns being used. IIRC devcontainer, communication and baselibs are using merge queues as well.

anmittag:
@AlexanderLanin any update on the PR?

PR #2997, opened 2026-06-18
Feature Request for DDS Gateway

Abhishek2581:
DDS Gateway: Recreated PR following accidental merge/closure of the previous PR. No functional changes beyond the requested modifications.


LittleHuba (changes requested):
Blocking this PR formally until agreement is achieved in Communication FT.

docs/features/communication/dds_gateway/index.rst
LittleHuba, line 1: I would like to place the same requests as we agreed on with the SOME/IP gateway.

1. How a translation from intra-ecu to inter-ecu communication is integrated in the system is very much up to the integrator.
2. The DDS gateway implementation that S-CORE provides is only a showcase on how this problem can be solved. Users must be able to implement/integrate a different DDS translation concept (e.g. one without explicit gateway process).
3. The DDS logic must be a binding beneath mw::com (e.g. the gateway internally uses mw::com to interact with the DDS stack) This is the foundation of supporting different integration concepts.

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,178 @@
...
+
+  - Runtime type definition via configuration or via dynamic library
+  - No dependency on DDS IDL generation
+  - Enables data translation and consistent serialization across middleware boundaries
+  - Supports DDS standard encodings (e.g., XCDR1 and XCDR2) for interoperability
+
+- End-to-End (E2E) protection:
+
+  - Centralized handling of Counter, CRC, and DataID
+  - Validation and protection configurable per route
```
LittleHuba: E2E protection handling might be impossible to achieve in a centralized handling. In some of our supported operating systems the network stack is not qualified.
The SOME/IP gateway works around this with a split of the "gateway process" into a QM and an ASIL-B part.

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,178 @@
...
+The DDS Gateway provides:
+
+- Bridging between ``mw::com (LoLa)`` and DDS via the gateway:
+   - ``mw::com (LoLa)`` → DDS GW → ``mw::com (LoLa)`` (inter-ECU communication via DDS)
+   - ``mw::com (LoLa)`` → DDS GW → DDS applications
+   - DDS applications → DDS GW → ``mw::com (LoLa)``
+
+- Configurable routing:
+
+  - Mapping between ``mw::com`` events(TBD for fields and methods) and DDS topics
```
LittleHuba: I'd be interested in more detail on this mapping.
What are the supported mapping cases? What are the corner cases where support is not provided?
It is hard to judge feasibility without further details.

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,178 @@
...
+  - Support for DDS domain-based routing
```
LittleHuba: Same here, please add more details / link to references.
I'm not a DDS expert so I need some more background to judge.

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,178 @@
...
+
+- Dynamic Type handling:
+
+  - Runtime type definition via configuration or via dynamic library
```
LittleHuba: This sounds like we have a huge overlap with the SOME/IP efforts. The format might differ, but the general task is the same. I'd like to see an investigation on how much we can reuse from the upcoming SOME/IP infrastructure.

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,178 @@
...
+  - Supports DDS standard encodings (e.g., XCDR1 and XCDR2) for interoperability
+
+- End-to-End (E2E) protection:
+
+  - Centralized handling of Counter, CRC, and DataID
+  - Validation and protection configurable per route
+
+- DDS stack abstraction:
+
+  - Pluggable DDS implementations via defined interfaces
```
LittleHuba: Ideally this stack abstraction is mw::com (as highlighted in a previous comment)

docs/features/communication/dds_gateway/index.rst
JochenSatETAS: Maybe one of the first decisions is whether we want to provide the service-grouping also to DDS and which kind of service discovery we want to do in this case. Or more specific do we want to communicate with AUTOSAR-stacks over DDS?

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,178 @@
...
+
+- Configurable routing:
+
+  - Mapping between ``mw::com`` events(TBD for fields and methods) and DDS topics
+  - Support for DDS domain-based routing
+
+- Dynamic Type handling:
+
+  - Runtime type definition via configuration or via dynamic library
+  - No dependency on DDS IDL generation
```
JochenSatETAS: Typically a DDS-stack provides a type-safe API (for C++). This API is generated. I'm not aware that DDS-stacks maintain a stable API between the generated and non-generated parts. Maybe I missed something here, but we need to discuss this.

docs/features/communication/dds_gateway/index.rst
JochenSatETAS: I'm not aware that a DDS-library with a split which enables the E2E-protection running in one process and the network handling running in another process exists. AFAIK CycloneDDS does not have that split.

But I agree that we want something like that in order to enable a usage of DDS in ASIL-scenarios.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: Agreed.

We will align the DDS solution with the architectural concept used for the SOME/IP gateway.

The DDS Gateway described in this FR will be positioned as the **reference implementation** demonstrating one possible translation concept between intra-ECU and inter-ECU communication. Integrators are not required to use this explicit gateway process and may implement alternative translation concepts, including solutions without a dedicated gateway process.

To support this flexibility, the DDS functionality will be provided as a **DDS binding beneath `mw::com`**. The reference DDS Gateway will internally use this binding rather than interacting directly with the DDS stack. This establishes a common integration layer that enables different deployment and translation concepts while keeping the gateway as the reference implementation.

The FR will be updated to clarify this architecture.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: > I'm not aware that a DDS-library with a split which enables the E2E-protection running in one process and the network handling running in another process exists. AFAIK CycloneDDS does not have that split.
> 
> But I agree that we want something like that in order to enable a usage of DDS in ASIL-scenarios.

Good point.

The intention is not to require the DDS stack itself to implement or be aware of E2E protection.

Our current direction is that the DDS stack remains responsible only for transporting the payload, while the Gateway (or another ASIL component built on top of the DDS binding) performs the E2E protection and validation. Since the gateway uses runtime type information, it can insert and validate the E2E information within the payload before handing it to the DDS stack and after receiving it.

We will update the FR to clarify that E2E handling is performed by the translation layer rather than implying it is part of the DDS stack or a centralized network component.

Abhishek2581: > I'm not aware that a DDS-library with a split which enables the E2E-protection running in one process and the network handling running in another process exists. AFAIK CycloneDDS does not have that split.
> 
> But I agree that we want something like that in order to enable a usage of DDS in ASIL-scenarios.

Agreed.

The intention is not to rely on a DDS implementation providing a QM/ASIL split internally. Instead, the DDS stack is considered the transport layer, while E2E handling is performed by the translation layer built on top of the DDS binding.

This keeps the E2E logic independent of the underlying DDS implementation and avoids requiring a specific DDS stack architecture.

We'll clarify this in the FR so it does not imply that such a split already exists within current DDS implementations.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: Good point.

The initial scope is intentionally focused on configuration-driven event communication.

The supported mapping cases are:

- One mw::com event is mapped to one DDS Topic.
- DDS Topic data is mapped back to the corresponding mw::com event.
- The DDS binding creates the corresponding DDS Topic, DataReader and DataWriter from deployment configuration.
- Runtime type information is referenced from the deployment configuration through an external JSON type description.
- DDS-specific attributes such as QoS policies and DDS Domain IDs are provided through DDS deployment configuration.
- Standard DDS endpoint discovery is used to determine availability of configured communication endpoints and map this to the mw::com availability model.
- Communication with native DDS applications is supported when the DDS topic, type, QoS and domain configuration match.

The corner cases / limitations for the initial scope are:

- mw::com fields are not supported initially.
- mw::com methods are not supported initially.
- Automatic mapping without deployment configuration is not supported.
- DDS deployment extensions for QoS, Domain IDs and type-description references still need detailed definition.
- DDS Domain ID allocation strategy still needs detailed definition. For example, different DDS domains may be used to separate QM and ASIL communication routes.

I'll update the FR to clarify the supported event-to-topic mapping, the use of DDS discovery, and the current limitations/open architectural points.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: Good point. I've updated the FR to provide more details on the mapping and added a short architectural overview of DDS concepts (Topics, DataReaders/DataWriters, and Domains) so that the proposed integration can be understood without requiring detailed DDS knowledge.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: We'll discuss this with the SOME/IP team to identify common infrastructure and reusable components. The goal is to align the DDS and SOME/IP approaches where possible while keeping middleware-specific functionality isolated behind their respective bindings.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: Agreed. The architecture has been updated accordingly. The primary abstraction is now the DDS communication binding beneath `mw::com`, while the DDS Gateway is positioned as a reference implementation built on top of this binding. This enables alternative integration concepts to reuse the same `mw::com` binding without requiring an explicit gateway process.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: Good point. The intention is not to depend on vendor-specific DynamicData APIs provided by individual DDS implementations.

The DDS binding owns the runtime type handling as well as runtime serialization/deserialization based on the configured runtime type definitions. The underlying DDS stack is accessed through an abstract interface responsible for DDS communication, allowing different DDS implementations to provide their own adapter without exposing vendor-specific runtime type APIs to the binding.

I've updated the FR to reflect this architecture by moving runtime type handling and serialization/deserialization into the DDS binding and defining the DDS stack interaction through abstract interfaces.

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,273 @@
...
+- Configuration-driven mapping between ``mw::com`` and DDS:
+
+  - Each ``mw::com`` event is mapped to one DDS Topic
+  - DDS Topic data is mapped back to the corresponding ``mw::com`` event
+  - DDS deployment configuration extends the ``mw::com`` deployment
+    configuration with DDS-specific attributes such as runtime type references,
+    QoS policies and DDS Domain IDs
+  - Standard DDS endpoint discovery is used to determine the availability of
+    configured DDS endpoints and map this to the ``mw::com`` availability model
+
```
JochenSatETAS: How is the ServiceDiscovery done on the service-level?

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,273 @@
...
+It provides:
+
+- Bridging between ``mw::com (LoLa)`` and DDS:
+   - ``mw::com (LoLa)`` → DDS GW → ``mw::com (LoLa)``
+   - ``mw::com (LoLa)`` → DDS GW → DDS applications
+   - DDS applications → DDS GW → ``mw::com (LoLa)``
+
+- End-to-End (E2E) protection:
+  - The translation layer performs generation and validation of Counter, CRC, and DataID
+  - Independent of the underlying DDS stack implementation
```
JochenSatETAS: Should we detail more the intention about E2E/Code-Generation/Overlay-Protocol? For me it looks like here are some implicit assumptions made, but I'd like them to be explicit so I can check for consistency.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: DDS discovery operates at the Topic and endpoint level, whereas `mw::com`
availability is exposed at the service level.

The gateway configuration defines the mapping between DDS Topics and the
events of a corresponding `mw::com` service instance.

The DDS Communication Component monitors the discovery state of the configured
DDS Topics and derives service availability from the required Topics of the
configured service instance.

Once the required DDS Topics become available, the Gateway Component offers
the corresponding local `mw::com` service. Loss of the required DDS Topics
results in withdrawal of the local service availability.

To make this explicit, I added a service availability propagation section and
a high-level diagram to the Feature Request.

docs/features/communication/dds_gateway/index.rst
Abhishek2581: Agreed. The previous text left some assumptions implicit.

The intended model is that E2E processing is owned by the Gateway Component
and remains independent of the selected DDS stack implementation.

Runtime type information associated with the configured communication route is
used for serialization, deserialization, and E2E generation/validation.

The DDS Communication Component is responsible for DDS communication and
transports serialized payloads, while interpretation and validation of E2E
information remain within the Gateway Component.

I updated the Feature Request to make these responsibilities explicit and
added a high-level data-flow diagram showing the interaction between runtime
type handling, E2E processing, serialization, and DDS communication.

docs/features/communication/dds_gateway/index.rst
LittleHuba: Let me clarify this further to not restrict this in the wrong way:
There are two basic approaches:
1. Applications know in their deployment that they talk to the DDS daemon (e.g. binding approach)
2. Applications are completely unaware that they talk to the DDS daemon (e.g. gateway approach)

The gateway approach bases on the binding approach and adds an additional automated translation layer that fully hides any DDS specific information.
There are use cases out for both approaches. Since gatewaying bases on bindings, we must get the binding approach sound.

For the translation logic, we can likely only provide an example, since this is normally very vendor specific.
But showing this example is valuable (hence, why we do this for SOME/IP).

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,512 @@
...
+               |                                     ^
+               | mw::com API                         | mw::com API
+               v                                     |
+       +---------------------+               +---------------------+
+       |   DDS Binding       |               |   DDS Binding       |
+       +----------+----------+               +----------+----------+                       
+                  |                                     ^
+                  | DDS                                 | DDS
+                  v                                     |
+         ===================   DDS NETWORK   ===================
```
LittleHuba: Clarification: Will DDS work fully without any daemon that acts as intermediary per host for any DDS traffic?
If no, please add the daemon here and consider that you have essentially two cases:
1. Daemon interacts with multiple applications (binding usecase)
2. Daemon interacts with one single application that translates and forwards via LoLa to actual user applications (gateway)

docs/features/communication/dds_gateway/index.rst
```diff
@@ -0,0 +1,512 @@
...
+
+   Availability loss propagates in the reverse direction:
+
+   DDS endpoint lost
+        -> DDS Communication Component withdraws route
+        -> Gateway Component StopOfferService()
+        -> local service unavailable
+- End-to-End (E2E) protection:
+
+  - E2E protection is implemented by the gateway translation layer and is
```
LittleHuba: This might only work partially, since something like alive counters (as they exist in SOME/IP E2E) will be application specific. Also consider that there might be applications that still consume payloads even when E2E verification failed.

docs/features/communication/dds_gateway/index.rst
JochenSatETAS: Thanks for the answer, I will continue this discussion in the newly added section.

docs/features/communication/dds_gateway/use_cases/index.rst
```diff
@@ -0,0 +1,298 @@
...
+---------------------------
+
+An ``mw::com`` publisher or subscriber communicates with a compatible native
+DDS participant.
+
+Both sides use compatible DDS Domain, Topic, type, and QoS information.
+
+Endpoint-based availability may be used when the native DDS participant does
+not provide explicit service-instance state.
+
```
JochenSatETAS, line 43: How does this work with the serialization and E2E-protection above the DDS-interface?

docs/features/communication/dds_gateway/architecture/index.rst
```diff
@@ -0,0 +1,589 @@
...
+   |  +--------------------+---------------------+  |
+   |                       |                        |
+   |                   mw::com API                  |
+   |                       |                        |
+   |  +--------------------v---------------------+  |
+   |  | DDS Binding                              |  |
+   |  |                                          |  |
+   |  | - mw::com service adaptation             |  |
+   |  | - Runtime type handling                  |  |
+   |  | - Serialization / deserialization        |  |
```
JochenSatETAS: Does this serialization follow the payload-format specification of DDS?

docs/features/communication/dds_gateway/architecture/index.rst
```diff
@@ -0,0 +1,589 @@
...
+                           | LoLa IPC
+                           v
+   +------------------------------------------------+
+   | DDS Gateway Process                            |
+   |                                                |
+   | - Generic Proxy / Generic Skeleton             |
+   | - LoLa service interaction                     |
+   | - Route translation                            |
+   | - Runtime type handling                        |
+   | - Serialization / deserialization              |
```
JochenSatETAS: Same here, does this do the serialization according to the DDS-payload-format? Then parts of the DDS-stack are implemented here? A complete DDS-stack will have the serialization included.

docs/features/communication/dds_gateway/architecture/index.rst
Abhishek2581: Yes DDS GW shall follow DDS CDR formating

docs/features/communication/dds_gateway/use_cases/index.rst
Abhishek2581: For data exchange, the DDS Gateway shall serialize the application data using the DDS CDR encoding configured by the integrator for the corresponding communication route. The integrator shall configure the DDS Gateway to use the same DDS CDR encoding (XCDR1 or XCDR2) and the same application-data definition as the interoperating native DDS application.

When E2E protection is enabled, the E2E fields form part of the serialized application payload and are encoded together with the application data. Therefore, a native DDS application shall implement the same E2E profile to interoperate with an E2E-enabled communication route.

If a native DDS application does not support the E2E result interface defined by this feature request, end-to-end E2E protection cannot be provided. In such cases, E2E protection shall be disabled for the corresponding communication route through deployment configuration.

docs/features/communication/dds_gateway/architecture/index.rst
Abhishek2581: Yes. The DDS Gateway serializes the application data into the configured DDS payload representation before passing it to the DDS Communication Daemon. The daemon forwards the same serialized payload to the DDS stack using its CDR interface. The DDS stack transmits this payload over the network without modifying the application payload representation; it only performs the DDS middleware functions such as RTPS encapsulation, QoS, reliability, fragmentation, and transport.

On the receiving side, the DDS stack delivers the same serialized payload to the DDS Communication Daemon, which forwards it unchanged to the DDS Gateway. The DDS Gateway then deserializes the payload into the corresponding mw::com representation.

anmittag:
@Abhishek2581 any update on the PR?

anmittag:
@qor-lb , @arsibo please review and close the PR

qor-lb:
@anmittag this FEP is still in the phase of finding a shepherd - only afterwards we can proceed with the FCP.

@FScholPer you were thinking about taking over the shepherd role. Would you still be open to shepherd this FEP?

FScholPer:
@qor-lb I will do the shepherd

PR #3055, opened 2026-07-01
DR-009-ACL-Concept: Generic Format for Access Control List for Crypto

christianbe9:
SOME/IP and IPC (and preparation for further use cases)

Introducing a common format for an Access Control List, such that it is
only required to use one common parser, one common format for reviews
and also one common "ACL Manager" aka editor for ACLs and also compiler
for the ACL into a binary format.


docs/design_decisions/DR-009-arch-ACL-Concept.md
masc2023, line 1: According to existing DR, this should be DR-003-Arch, compare folder

christianbe9: fine for me - then we would also need to rename DR-008-int, as we currently have 2 different "counting" shemes:

for *int* we just in general incremented+1
for arch and infra we incremented only "arch internally" +1 and "infra internally" +1

masc2023: > fine for me - then we would also need to rename DR-008-int, as we currently have 2 different "counting" shemes:
> 
> for _int_ we just in general incremented+1 for arch and infra we incremented only "arch internally" +1 and "infra internally" +1

DR-xyz-<infra|arch|proc>, each should be in row, int may wrong, so
we have DR-001-arch, DR-002-arch -> DR-003-arch is with you

christianbe9: I rebased everything so it also matches the branch name: https://github.com/eclipse-score/score/pull/3058#issuecomment-4858253330

i just dont understand based on what the reviewers are chosen, as there are only 2 now.

masc2023: > I rebased everything so it also matches the branch name: [#3058 (comment)](https://github.com/eclipse-score/score/pull/3058#issuecomment-4858253330)
> 
> i just dont understand based on what the reviewers are chosen, as there are only 2 now.

The system detects, if Architecture DR -> Architecture people has to review, 
compare https://github.com/eclipse-score/score/blob/main/.github/CODEOWNERS
Row 31:
/docs/design_decisions/*arch*         @qor-lb @arsibo

By the Way, your PR is closed, you need to reopen it

christianbe9: Understood, thanks a lot!
So that every matches (commit names, branch name, filename) I updated everything, thats why this review got closed and was automatically generated here: https://github.com/eclipse-score/score/pull/3058#issuecomment-4858253330

And I think due to the intial wrong naming by me, the reviewer list now also automatically changed, what I understood now why :)

PR #3058, opened 2026-07-01
DR-003-ARCH-ACL-Concept: Generic Format for Access Control List

christianbe9:
Introducing a common format for an Access Control List (Crypto, SOME/IP,
IPC and further potential usecases), such that it is
only required to use one common parser, one common format for reviews
and also one common "ACL Manager" aka editor for ACLs and also compiler
for the ACL into a binary format.


docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
...
+   Context
+   -------
+   Access Control is required on multiple components; i.e.:
+   
+   - IPC (LoLa)
+   Who is allowed to publish which services (who can create/modify a file in the according directory)
+   For Service Providers: Once the IPC Channel/SharedMemory is created, whom to grant which rights to what area (data/control)
+   This permissions will be based on uid.
+   
+   - SOME/IPC
```
GuruprasadBhatL:
```suggestion
   - SOME/IP
```
?

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
...
+   - Crypto
+   Needs to know who (which application - i.e. which uid (every application shall run as own user))
+   -- created/owns a key
+   -- is allowed to write/change a key
+   -- is allowed to use a key for any operations (e.g. sign/verify/encrypt/decrypt/keyderive/etc.)
+   -- is allowed to read/extract a key (e.g. pub keys)
+
+   Consequences
+   ------------
+   To prevent each component from implementing compeletley own mechanism for persmission enforcement; the decision schould be an alignment on a common ACL format, so that all components can use:
```
GuruprasadBhatL:
```suggestion
   To prevent each component from implementing compeletley own mechanism for persmission enforcement; the decision should be an alignment on a common ACL format, so that all components can use:
```

GuruprasadBhatL:
A few generic questions : 

1. Is JSON as a config format centrally approved in eclipse/score ?
2. Is the proposal is to have ACL configuration for all features (Crypto/LoLa/SomeIP) in one file ? 
3. How is relationship established between ACL config and config of individual features like Crypto/LoLa.

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
...
+		  "version": 5,
+		  "tlsenable": 1,
+		  "ipsecenable": 1,
+		  "macsecenable":1,			  
+		  "specifics": {},
+		  "allow": 
+		  {
+			"uids": 
+			{
+			  "101":{"rights": ["write"],"name": "oem.app.1","version": [],"criticality": "ASIL-B","specifics": {}},
```
GuruprasadBhatL: Would like to understand significance of ``criticality`` in the current context.

How does it influence ACL configuration.

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
...
+	  ]
+	},
+	"rights": {
+	  "type": "array",
+	  "items": {
+		"type": "string",
+		"enum": [
+		  "read",
+		  "write",
+		  "monitor"
```
GuruprasadBhatL: what is monitor right ?

christianbe9:
> A few generic questions :
> 
> 1. Is JSON as a config format centrally approved in eclipse/score ?
I would say yes, as all other components (e.g. lifecycle) are also using json as config format
> 2. Is the proposal is to have ACL configuration for all features (Crypto/LoLa/SomeIP) in one file ?
No, this can be, but must not. I just put everything into one file as the idea is to have everything in the same format and have one parser capable of parsing all the properties. So you can have 3 configs (1 IPC, 1 SOME/IP, 1 Crypto) or 2 configs (1 IPC & SOME/IP, 1 crypto) or 1 config containing all thats up to the integrator. 
> 3. How is relationship established between ACL config and config of individual features like Crypto/LoLa.
ACL config is input to Crypto lib, LoLa and someipd; i.e. all 3 include the parser lib and use the acl, either from the same path or from 3 different paths, so the path where this ACL is located is configured per component.
Based on the ACL the enforcement is then done in each component

docs/design_decisions/DR-003-arch-ACL-Concept.md
christianbe9: Idea behind monitor is (I'm not sure how valid) is only related to IPC, for SOME/IP and crypto its same as read, for IPC the idea is to have only read permission to data and control plane, so that a montoring application cannot interfere with the normal behaviour, however that means you could have partial reads from control plane, therefore control plane might be monitored but not so sure how good of an idea it is. Concrete example was for a Intrusion detection, health monitoring, etc.; e.g. you could check if somebody is blocking in the control pane for too long

docs/design_decisions/DR-003-arch-ACL-Concept.md
christianbe9: As we do an ACL my question is: should we include criticality in the config as well to do a cross check/ensure such that there is no mixture "by accident" of criticality due to wrong configuration or something; i.e. can a ASIL application and QM application both interact via one IPC which should not be the case afaik. So this is not a security thing, question is if we also should use this for safety purpose or not.

ChansAlive:
> > A few generic questions :
> > 
> > 1. Is JSON as a config format centrally approved in eclipse/score ?
> >    I would say yes, as all other components (e.g. lifecycle) are also using json as config format
> > 2. Is the proposal is to have ACL configuration for all features (Crypto/LoLa/SomeIP) in one file ?
> >    No, this can be, but must not. I just put everything into one file as the idea is to have everything in the same format and have one parser capable of parsing all the properties. So you can have 3 configs (1 IPC, 1 SOME/IP, 1 Crypto) or 2 configs (1 IPC & SOME/IP, 1 crypto) or 1 config containing all thats up to the integrator.
> > 3. How is relationship established between ACL config and config of individual features like Crypto/LoLa.
> >    ACL config is input to Crypto lib, LoLa and someipd; i.e. all 3 include the parser lib and use the acl, either from the same path or from 3 different paths, so the path where this ACL is located is configured per component.
> >    Based on the ACL the enforcement is then done in each component

Hi,

`ACL config is input to Crypto lib`.  Considering we have a library daemon architecture for crypto stack, is this still valid? Would this be crypto daemon or still crypto lib?

christianbe9:
> > > A few generic questions :
> > > 
> > > 1. Is JSON as a config format centrally approved in eclipse/score ?
> > >    I would say yes, as all other components (e.g. lifecycle) are also using json as config format
> > > 2. Is the proposal is to have ACL configuration for all features (Crypto/LoLa/SomeIP) in one file ?
> > >    No, this can be, but must not. I just put everything into one file as the idea is to have everything in the same format and have one parser capable of parsing all the properties. So you can have 3 configs (1 IPC, 1 SOME/IP, 1 Crypto) or 2 configs (1 IPC & SOME/IP, 1 crypto) or 1 config containing all thats up to the integrator.
> > > 3. How is relationship established between ACL config and config of individual features like Crypto/LoLa.
> > >    ACL config is input to Crypto lib, LoLa and someipd; i.e. all 3 include the parser lib and use the acl, either from the same path or from 3 different paths, so the path where this ACL is located is configured per component.
> > >    Based on the ACL the enforcement is then done in each component
> 
> Hi,
> 
> `ACL config is input to Crypto lib`. Considering we have a library daemon architecture for crypto stack, is this still valid? Would this be crypto daemon or still crypto lib?

Yes, that was bad wording by me; so ACL would be input to Crypto Daemon (in a setup where you have a mw:crypto for the application which handles the IPC to the Deaemon using LoLa  and a Daemon which actually handles the jobs). The crypto lib itself will be part of the application, so it will not include the ACL check (as application would check itself then).

GuruprasadBhatL:
**Original Question** : How is relationship established between ACL config and config of individual features like Crypto/LoLa.

**Response** : 
ACL config is input to Crypto lib, LoLa and someipd; i.e. all 3 include the parser lib and use the acl, either from the same path or from 3 different paths, so the path where this ACL is located is configured per component.
Based on the ACL the enforcement is then done in each component

**Follow up question** : 

How is ACL format enforced in config of SomeIP/Crypto ? is is there a way for SOMEIP/Crypto config to inherit (enforce interface schema ) from ACL JSON schema ? 

if this is clarified, I have no further questions.

christianbe9:
> **Original Question** : How is relationship established between ACL config and config of individual features like Crypto/LoLa.
> 
> **Response** : ACL config is input to Crypto lib, LoLa and someipd; i.e. all 3 include the parser lib and use the acl, either from the same path or from 3 different paths, so the path where this ACL is located is configured per component. Based on the ACL the enforcement is then done in each component
> 
> **Follow up question** :
> 
> How is ACL format enforced in config of SomeIP/Crypto ? is is there a way for SOMEIP/Crypto config to inherit (enforce interface schema ) from ACL JSON schema ?
> 
> if this is clarified, I have no further questions.

From my side fine, makes sense, we can define it such that all ACLs must inerhit this ACL JSON schema included here. 

And then if some component needs something in addition we update the main ACL JSON schema such that it always remains central.

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
...
+{
+  "policyVersion": 1,
+  "policyId": "example-acl-policy-001",
+  "defaultEffect": "deny",
+  "generatedBy": "install-update-manager",
+  "generatedAt": "2026-06-17T14:28:00Z",
+
+  "services": 
+  {
+	"111":
```
FScholPer: What is this?

docs/design_decisions/DR-003-arch-ACL-Concept.md
FScholPer: I would move that to another schema file which is module specific

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
...
+   - updates due to parser vulnerabilities can be centrallized
+   
+   The base proposal is as follows:
+   - a json file as configuration
+   - for runtime efficiency: compilation of the json file to a flatbuffer/binary file
+   
+   Proposed json format & properties (example): 
+   
+  
+```json
```
FScholPer: Can we move the example to the bottom and first start with the schema

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
...
+   - Crypto
+   Needs to know who (which application - i.e. which uid (every application shall run as own user))
+   -- created/owns a key
+   -- is allowed to write/change a key
+   -- is allowed to use a key for any operations (e.g. sign/verify/encrypt/decrypt/keyderive/etc.)
+   -- is allowed to read/extract a key (e.g. pub keys)
+
+   Consequences
+   ------------
+   To prevent each component from implementing compeletley own mechanism for persmission enforcement; the decision should be an alignment on a common ACL format, so that all components can use:
```
FScholPer: typo permission

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
+# DR-009-ACL-Concept: Common ACL format for all components
```
FScholPer: I think we also need stakeholder requirements later on

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,553 @@
...
+	  "uniqueItems": true
+	},
+	"specifics": {
+	  "description": "OEM-specific or project-specific extension object. Unknown properties are allowed only here.",
+	  "type": "object",
+	  "additionalProperties": {
+		"$ref": "#/$defs/specificValue"
+	  }
+	},
+	"specificValue": {
```
FScholPer, line 178: I think it would make sense to define specific schema and handler of the extension

anmittag:
@arsibo please review

docs/design_decisions/DR-003-arch-ACL-Concept.md
christianbe9: We can define it in a seperate schema for sure, but I would like to keep it as generic as possible, as more abstractions  often are confusing; is there anything else except read and write which is related to communication? 
read -> subscribe/1:1/consume/request aka receive response
write -> publish/1:1/produce/send response
monitor -> observe, without ability to lock while reading

Then we could define this in a seperate scheme with those 3 base rights, and projects could, if they want refine it.

anmittag:
@christianbe9 any update on the PR?

christianbe9:
Hi, yes we had a breakout session yesterday with crypto specifics, so there will be new updates coming today; after this we will also start the discussion within the architecture group, i.e. com (lola/someip). When this is finished as well the PR is ready to go.

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,774 @@
...
+				"memoryLimiting": {"slots":4},
+				"jobLimiting": {"parallel":1}	
+			}
+		}
+	},
+	"keyspaces":
+	{
+		"0":
+		{
+			"keyspaceName": "tls_auth",
```
ChansAlive, line 731: What is a key space? Why do we need it?

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,774 @@
...
+			"keys":
```
ChansAlive, line 732: This must be keyslot right?

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,774 @@
...
+	},
+	"operationalRights": 
+	{
+		"uid=1": 
+		{
+			"Software":
+			{
+				"nonKeyspaceDependent": ["hash","keygen","keyimport"],
+				"memoryLimiting": {"bytes":8178},
+				"jobLimiting": {"parallel":3}
```
ChansAlive, line 704: How is this linked to the operations? This could mean every operation can be instantiated 3 times in parallel right? (Or in total 3?)

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,774 @@
...
+        "keyderive",
+        "wrap",
+        "unwrap",
+        "mac",
+        "random",
+        "keygen",
+        "keyimport",
+        "keyexport",
+        "keydelete"
+      ]
```
ChansAlive, line 464: Is this extensible considering we might introduce more and more operations and context in crypto?

ChansAlive (commented):
Initial review

anmittag:
@christianbe9 please update your PR according to the discussions. set it to draft if not finalized

docs/design_decisions/DR-003-arch-ACL-Concept.md
```diff
@@ -0,0 +1,885 @@
...
+		"modify",
+		"confirmTrusted"
+	  ],
+	  "properties": {
+		"modify": {  
+		  "$ref": "#/$defs/subjectList"
+		},
+		"confirmTrusted": {
+			"description": "If put to True, then even though an application A might have rights to modify/update a certificate in a certificate slot, the owner of the truststore must accept it, otherwhise it will be marked invalid for the truststore. If put to False, if application A has rights to modify/uipdate a certificate, then its by default trusted by the other parties as well.",
+			"type": "boolean"
```
ChansAlive, line 495: I would say this shall be the property of the membership for a `certslot` in a `truststore`. 

ACL shall define who is able to modify the `truststore`. Whether a certificate is conditionally accepted or not shall be the property of the `truststore`, which can then be influenced by the user having the modify rights.

PS: This is planned to be a config of the crypto configuration for the trust store members rather than in ACL.

PR #3094, opened 2026-07-16
chore(ci): run more checks on-pr

AlexanderLanin:
Instead of running a few checks directly, let's run the [*full blown mega clever central on-pr workflow*](https://github.com/eclipse-score/cicd-workflows/blob/on-pr/.github/workflows/on-pr.md)


.github/workflows/on-pr.yml
a-zw, line 17:
```
  pull_request_target:
```

Otherwise it does not trigger for pull requests from forks (like etas-contrib), does it?

AlexanderLanin: @a-zw pull_request_target is only required if you want the pull request to have access to secrets. Thats quite difficult to get right. And in this case we dont need it, we only need content: read.

See checks, it did trigger:
image

MaximilianSoerenPollak (approved):
For me this seems to be a good idea, switch everything to the right workflow with the right permissions.

PR #3108, opened 2026-07-21
remove v0.5 beta in feature description

anmittag:
Cleaning-Up of the Feature Requirements:

Removed V0.5 Beta in the Headlines of Base Libs, Communication, FEO, Orchestration and Persistency

### DIFF
--- a/docs/features/baselibs/index.rst
+++ b/docs/features/baselibs/index.rst
@@ -14,8 +14,8 @@
 
 .. _baselibs_feature:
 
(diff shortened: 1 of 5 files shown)


FScholPer:
The idea was to mark the components which are part of the last release. But that was a manual try. Probably better to do that in an automatic fashion. One Solution could also be to add some kind of grouping of features which are "valid_from" of that release

FScholPer (approved):

PR #3119, opened 2026-07-24
IntermediateSafetyGoalASILB

anmittag:
Remove Requirement Note from "ASIL - B with 0.5"


FScholPer (changes requested):
I would suggest to align on a range of qm to asil-d but this is to the module to decide. The goal of having the complete stack in ASIL-D is just not doable in open source

FScholPer (approved):

PR #3137, opened 2026-07-29
change code coverage

RolandJentschETAS:
This pull request updates the documentation for code coverage analysis in C++ projects, specifically clarifying the recommended approach for coverage measurement and its justification for safety certification. The main changes focus on switching from GCC/gcov-based coverage to LLVM/clang-based coverage on the host, and providing a detailed rationale for this methodology in the context of safety standards.

**Code Coverage Documentation Updates:**

- Updated the coverage section to recommend using LLVM/clang-based source coverage on the Linux host instead of GCC/gcov or QNX/qcc-based solutions, and clarified that coverage is not collected on the QNX target due to lack of support.
- Added a new "Argumentation" subsection explaining why host-based coverage with target test execution is sufficient for ISO 26262-6 ASIL_B, including a two-step verification process (quantitative on host, qualitative on target), and the benefits of avoiding instrumentation on the target.
- Updated the UML diagram and tool references to replace `gcov + gcovr` with `llvm-cov + llvm-profdata`, and removed QNX as a coverage collection target.


docs/contribute/development/cpp/code_analysis.rst
```diff
@@ -102,25 +102,38 @@ Adress/ Leak Sanitizer (ASAN/LSAN)
...
+   gtest --> llvm
+   llvm --> host
+
+Argumentation: Host-based Coverage with Target Test Execution
+-------------------------------------------------------------
+
+Measuring code coverage exclusively on the **Linux host** (via LLVM) is considered sufficient for safety certification, provided that a **two-step verification** is performed to demonstrate equivalence on the target:
+
+#. **Quantitative Verification (Host):** The complete test suite is executed on the Linux host with code coverage enabled. This demonstrates that the test cases are structurally complete (100% C0/C1 coverage) and that no dead code exists.
+#. **Qualitative Verification (Target):** The identical test suite is executed on the **QNX target**, but with code coverage instrumentation turned off. This demonstrates that the software compiles, links, and behaves identically on the real target hardware (correctness of execution, no compiler/linker optimization bugs, no endianness or memory alignment issues).
```
limdor, line 141: What about when the baseline of coverage is different? For example, files that might be QNX only.
I think this argumentation is missing what happens with code that it is QNX only and code that it is Linux only.

docs/contribute/development/cpp/code_analysis.rst
```diff
@@ -102,25 +102,38 @@ Adress/ Leak Sanitizer (ASAN/LSAN)
...
+
+This approach satisfies *ISO 26262-6* for *ASIL_B* for the following reasons:
+
+* **Identical Test Results:** Passing 100% of the tests on both the host and the target demonstrates that the control flow under test is identical.
+* **Mitigation of Target Instrumentation Risks:** Turning off coverage instrumentation on the target is recommended for embedded systems, as active instrumentation alters the timing behavior, memory footprint, and compiler optimizations on the target. Testing the uninstrumented code on QNX ensures that the actual production binary is verified (mitigating "Heisenbugs").
```
limdor: That is true, instrumentation affects behavior, but it could be run, once instrumented to measure coverage and one without instrumentation just to check that the tests pass.

docs/contribute/development/cpp/code_analysis.rst
```diff
@@ -102,25 +102,38 @@ Adress/ Leak Sanitizer (ASAN/LSAN)
...
+* **Equivalence Justification:** The host-based coverage measurement is justified by confirming that no platform-specific code paths (e.g., ``#ifdef QNX`` blocks) are bypassed. Any target-specific hardware abstraction layer (HAL) is verified separately via system-level integration tests.
```
limdor: > The host-based coverage measurement is justified by confirming that no platform-specific code paths (e.g., ``#ifdef QNX`` blocks) are bypassed.

How should a developer do that? How can a developer find such paths? As mentioned above, it is not only the pragmas, but there could be files that are completly excluded from the linux build.

docs/contribute/development/cpp/code_analysis.rst
RolandJentschETAS: Good point — the argumentation was missing the case where the coverage baseline differs between host and target. I've added a new subsection **"Handling Platform-Specific Code"** that splits this into three cases:

1. **Linux-only code** (mocks, emulation, host utilities): classified as QM/test-support and excluded from the ASIL-B production baseline, so it can't skew the metrics.
2. **QNX-only code inside mixed files** (`#ifdef __QNX__` blocks): marked with governed coverage-exclusion markers plus a justification that the region is verified on the target.
3. **QNX-only whole files** compiled only for QNX: excluded from the host structural-coverage baseline (to avoid false-negative 0%), and treated as a **justified deviation with residual risk** rather than an equivalence claim, compensated by requirements-based integration tests on the target.

The subsection also explains that the host/target coverage baselines differ by construction and how each delta is reconciled.

docs/contribute/development/cpp/code_analysis.rst
RolandJentschETAS: Agreed. I've reworded the *Mitigation of Target Instrumentation Risks* bullet to make explicit that the two-step verification **is** exactly that pattern — measure coverage with instrumentation, then re-run without instrumentation to confirm the tests still pass — but split across the two environments: the **host run is instrumented** (for structural coverage) and the **target run is uninstrumented** (to confirm the production binary behaves identically).

The one constraint I called out in the text: `qcc` cannot produce LLVM source-based coverage, so an *instrumented* coverage run on the QNX target itself is not available; the instrumented measurement necessarily happens on the host.

docs/contribute/development/cpp/code_analysis.rst
RolandJentschETAS: You're right that this can't rely on a developer manually grepping for `#ifdef` guards, since whole files can be excluded from the host build. I've added an **"Identifying platform-specific code"** subsection describing a concrete, tool-based process:

- **Build-set comparison:** the set of files compiled for the host is diffed against the set compiled for QNX; files present only in the QNX build are flagged automatically as the "QNX-only files" case.
- **Explicit exclusion manifest:** every host-coverage exclusion (per-file or within-file region marker) is recorded in one reviewable list, so the full delta between the host baseline and the target code base is visible in a single place.
- **CI gate:** the comparison and the manifest are enforced in CI, so a new QNX-only file or conditional block without a documented exclusion and a compensating target verification fails the quality gate.

I also added a note that the exclusion-marker syntax must match the actual toolchain: since coverage uses `llvm-cov`/`llvm-profdata`, `LCOV_EXCL_*` markers are not interpreted natively and must not be assumed to work unless an `lcov` conversion/merge step is explicitly part of the pipeline.

RolandJentschETAS: I'm only not sure, how we bring this to the infrastructure.

limdor: But then to me it sounds that this PR is dropping the need for code coverage on QNX without a clear path on how to  cover the gap in an automatic, reliable way.

aschemmel-tech (changes requested):
Block for discussion in process community

aschemmel-tech (changes requested):
see discussion: https://github.com/orgs/eclipse-score/discussions/2234#discussioncomment-17837708
- also change https://eclipse-score.github.io/score/main/platform_management_plan/software_verification.html#tools
- include checkpoint for missing "QNX only" code coverage in implementation inspection.

docs/contribute/development/cpp/code_analysis.rst
RolandJentschETAS: After discussion with the process colleagues we will add a review checklist point for that.

anmittag:
@RolandJentschETAS 

- qnx is providing its own coverage tool, why not using gcov/lcov in such cases? 
- have you added the information to check on target into the safety manual?

aschemmel-tech:
need to re-discuss in process/safety community together with @masc2023 (after discussion in PL round). Expectation is that the explanation how to (not) cover QNX only code is not sufficient. Do we have this (in which amount)?

RolandJentschETAS:
> @RolandJentschETAS
> 
>     * qnx is providing its own coverage tool, why not using gcov/lcov in such cases?
> 
>     * have you added the information to check on target into the safety manual?

@anmittag : Please see the text of the design decision itself. gcov have same issues, which leads to questionable result in respect of the fulfillment of the safety requirements (like branch coverage). Additional do not want spend additional effort to support and merge two different coverage toolings without any need. A didn't add the check, because the DR is not approved yet.

docs/contribute/development/cpp/code_analysis.rst
```diff
@@ -102,25 +102,91 @@ Adress/ Leak Sanitizer (ASAN/LSAN)
...
-   object "QNX" as qnx
 
    coverage --> gtest
-   gtest --> gcov
-   gcov --> host
-   gcov --> qnx
+   gtest --> llvm
+   llvm --> host
+
+Argumentation: Host-based Coverage with Target Test Execution
```
aschemmel-tech: As agreed in the community in https://github.com/orgs/eclipse-score/discussions/2234#discussioncomment-17999707 this has to be re-written based on the Exida comment https://github.com/orgs/eclipse-score/discussions/407#discussioncomment-18018143. Main points:
- S-CORE shall not argue why a structural coverage on the target is not needed - it is needed to cover undefined behaviour.
- S-CORE will use llvmcov to determine structural coverage on host, but will not determine structural coverage on (reference) target
- S-CORE will have a platform AoU to require determining the structural coverage on target from the user/distributor.

aschemmel-tech (changes requested):
needs rework after process community discussion, see inline comment

docs/contribute/development/cpp/code_analysis.rst
RolandJentschETAS: Fixed ...

LittleHuba: @aschemmel-tech we have target-only safety-relevant code. We can not just tell users to take care of structural coverage on target. This would defeat our goal to become safety elements out of context.

PR #3147, opened 2026-08-03
docs: add breaking changes policy document

fbaeuerle:
Introduce breaking_changes.rst, documenting S-COREs policy for breaking changes.

### DIFF
--- a/docs/contribute/development/breaking_changes.rst
+++ b/docs/contribute/development/breaking_changes.rst
@@ -1,0 +1,118 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2026 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+.. document:: Breaking Changes
+   :id: doc__breaking_changes
+   :status: valid
+   :version: 1
+   :safety: ASIL_B
+   :security: NO
+   :realizes: wp__sw_development_plan[version==1]
+
+.. _breaking_changes:
+
+Breaking Changes
+#################
+
+Scope
+*****
+
+This document only concerns **source-level** breaking changes, i.e. changes that require a user of a
+module (another S-CORE module, a downstream integrator or an external consumer) to adapt their own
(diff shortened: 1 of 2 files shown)


docs/contribute/development/breaking_changes.rst
```diff
@@ -0,0 +1,118 @@
...
+behaving unchanged. Only once the announced grace period (e.g. tied to a number of releases or a
+deadline stated in the deprecation message) has elapsed may the deprecated overload be removed.
+
+Handling Unavoidable Breaking Changes
+**************************************
+
+There may be rare cases where a breaking change cannot be avoided, even if deprecation was considered.
+In such cases:
+
+* The commit introducing the breaking change **must** clearly state that it contains a breaking change,
```
fbaeuerle, line 110: Here I would suggest adding some sort of distinct tag to the git commit below the commit message like "BREAKING CHANGE", like described [here](https://www.conventionalcommits.org/en/v1.0.0/)

docs/contribute/development/breaking_changes.rst
```diff
@@ -0,0 +1,118 @@
...
+**deprecated**.
+
+This means:
+
+#. New behavior, APIs or interfaces are added without altering the observable behavior, signature or
+   contract of what already exists.
+#. The functionality that the new addition is meant to replace is marked as deprecated (e.g. via
+   language-specific deprecation attributes/annotations, documentation notes, and/or compiler or linter
+   warnings), so that users are made aware of the planned removal as early as possible.
+#. Users are given a **grace period** during which both the deprecated and the new functionality are
```
fbaeuerle, line 68: we have to specify the length of the grace period. for example for semver usually deprecated symbols are only removed on major releases.

docs/contribute/development/breaking_changes.rst
```diff
@@ -0,0 +1,118 @@
...
+           return ComputeChecksum(data, ChecksumAlgorithm::kCrc32);
+       }
+   };
+
+The ``[[deprecated]]`` attribute makes the compiler emit a warning at every call site of the old overload.
+This gives users of the module clear, tool-supported notice, while their existing code keeps building and
+behaving unchanged. Only once the announced grace period (e.g. tied to a number of releases or a
+deadline stated in the deprecation message) has elapsed may the deprecated overload be removed.
+
+Handling Unavoidable Breaking Changes
```
LittleHuba, line 104: I would welcome if we add a section for purposefully unstable API:

Sometimes new features are introduced with an API where not enough experience is gained yet. Forcing migration paths via deprecations at that moment will slow down development significantly.

Such APIs should be marked either in code or in the release notes as unstable/experimental.

This is a warning to users, that depending on these APIs is a maintenance risk on their side, since API may break at any moment in any way without notice or migration paths.

PR #3148, opened 2026-08-03
Fix example command in README

nicu1989:
Corrected command from 'bazel tests' to 'bazel test' for running Python tests.


FScholPer (approved):

AlexanderLanin (approved):
no idea why I'm codeowner here, but fine 😆

PR #3166, opened 2026-08-07
Update Documentation Mgt Plan

PandaeDo:
Update to S-CORE 1.0 and also to https://eclipse-score.github.io/score/main/contribute/general/folder.html


masc2023 (changes requested):
in the documentation many needs do not pass the filter

PandaeDo:
> in the documentation many needs do not pass the filter

Yes sure. A lot of the expected documentation is not created yet. My idea was to update to platform architecture 1.0 so that we can see also in the documentation plan what's actually missing.

masc2023 (approved):

PR #3171, opened 2026-08-11
Add module folder in structure

aschemmel-tech:


RolandJentschETAS (approved):

masc2023:
@aschemmel-tech , @RolandJentschETAS , seems some duplicated information with module_templates, should we not consider only to maintain the information once and links to it , if needed?

PR #3182, opened 2026-08-12
Update codeowner for score_tools

masc2023:
SCORE_TOOLS contains the tool evaluation, qualification reports, shall be approved by Safety/Security Manager


MaximilianSoerenPollak (approved):
For me this makes sense, but I would like some more input from others here before we merge this.

anmittag (approved):

PR #3195, opened 2026-08-19
ci(docs): fix docs publishing

AlexanderLanin:


.github/workflows/docs.yml
```diff
@@ -14,30 +14,20 @@
...
-  contents: write
-  pages: write
-  pull-requests: write
-  id-token: write
+  contents: read
 
 on:
-  pull_request_target:
-    types: [opened, reopened, synchronize] # Handles forked PRs
+  pull_request:
```
a-zw, line 20: This means PRs from outside of eclipse-score don't get built? Probably not the intention? Undo?

```suggestion
  pull_request_target:
```

.github/workflows/docs.yml
```diff
@@ -14,30 +14,20 @@
...
   docs-build:
     uses: eclipse-score/cicd-workflows/.github/workflows/docs.yml@8d80e8df150cae21d53cbc8031d0f970648f7a67 # v0.0.3
-    permissions:
-      contents: write
-      pages: write
-      pull-requests: write
-      id-token: write
-    with:
-      bazel-target: "//:docs -- --github_user=${{ github.repository_owner }} --github_repo=${{ github.event.repository.name }}"
-      retention-days: 3
```
a-zw, line 43: docs.yml uploads an artifact for docs-publish.yml, doesn't it? Or does it push to gh-pages? Either way it requires some permissions, doesn't it?

MaximilianSoerenPollak (dismissed):

.github/workflows/docs.yml
a-zw: Misunderstanding. The difference is just permissions. Not triggering.

a-zw (approved):

MaximilianSoerenPollak (approved):
Questions answered in a call.

PR #3194, opened 2026-08-19
Add DR about module folder structure

aschemmel-tech:
DR raised based on the discussion here:
https://github.com/orgs/eclipse-score/discussions/434#discussioncomment-18048797

PR #3196, opened 2026-08-19
add design decision include path

### LINKED ISSUE #3269: Public Header Location and Directory Structure for C++ Component
### Feature Request / Modification Description
author: @RolandJentschETAS 
shepherd: @LittleHuba 

### Expected Changes of work products

- [ ] Requirements
- [ ] Architecture
- [ ] Safety Analysis
- [ ] Security Analysis
- [ ] Detailed Design
- [ ] Implementation and Testing
- [ ] all

### Impact analysis

see DR

### Safety or Security relevance

- [ ] none
- [ ] Safety relevant
- [ ] Security relevant

### Expected required ASIL classification

QM

### Expected Implementation Version (Release)

1.0

RolandJentschETAS:
add design decision include path

### DIFF
--- a/docs/design_decisions/DR-003-arch.md
+++ b/docs/design_decisions/DR-003-arch.md
@@ -1,0 +1,457 @@
+<!--
+Copyright (c) 2026 Contributors to the Eclipse Foundation
+
+See the NOTICE file(s) distributed with this work for additional
+information regarding copyright ownership.
+
+This program and the accompanying materials are made available under the
+terms of the Apache License Version 2.0 which is available at
+https://www.apache.org/licenses/LICENSE-2.0
+
+SPDX-License-Identifier: Apache-2.0
+-->
+(DR-003-Arch)=
+# DR-003-Arch: Public Header Location and Directory Structure for C++ Components
+
+* **Date:** 2026-08-19
+
+```{dec_rec} Public Header Location and Directory Structure for C++ Components
+:id: dec_rec__arch__public_header_layout
+:status: proposed
+:version: 1
+:context: Architecture
+:decision: Separate include/ directory for public API headers, private headers next to sources
+```
+
+---
+
+## Context / Problem
+
+As we modularize our C++ components with Bazel, we need a standardized directory
+layout that defines where public (API) and private (implementation) headers live.
+This decision directly impacts API encapsulation, developer workflow, packaging for
+non-Bazel consumers, and integration with Bazel's strict header visibility model
+(`hdrs` vs. `srcs`).
+
+Bazel is authoritative here: a `cc_library` distinguishes between headers that form
+the compilable public surface of a target (`hdrs`) and headers that are purely
+internal to the compilation of that target (`srcs`). A downstream target may only
+`#include` headers that appear in the `hdrs` of a library it directly depends on.
+The physical directory layout does not by itself enforce this boundary — the `BUILD`
+file does. The question is therefore how to structure directories so that the
+physical layout *reinforces* the logical API boundary that Bazel enforces, rather
+than working against it.
+
+### Current state in S-CORE
+
+Within S-CORE there is mostly **no separate `include/` directory** in use today.
+Components rely completely on Bazel's `hdrs` attribute to designate public headers,
+with headers and sources living together in a flat directory. This works for Bazel,
+but it makes the public contract of a component invisible at the file-system level:
+a developer has to open `BUILD.bazel` to learn which headers are public, and
+extracting a clean public SDK for non-Bazel consumers requires filtering.
+
+### Goals
+
+- Make the public API surface of a component obvious without reading `BUILD.bazel`.
+- Map cleanly onto Bazel's `hdrs`/`srcs` split with minimal boilerplate.
+- Produce clean, collision-free `#include` paths for consumers.
+- Enable simple, folder-level packaging of the public API for non-Bazel consumers.
+- Keep the developer's day-to-day workflow (edit `.h` next to `.cc`) reasonable.
+
+### External references
+
+- **[The Pitchfork Layout (PFL)](https://github.com/vector-of-bool/pitchfork)** —
+  in particular the *separated header placement* convention, where public headers
+  live under [`include/`](https://joholl.github.io/pitchfork-website/#tld.include)
+  and private headers/sources live under `src/`.
+- **[The Canonical Project Structure (P1204R0)](https://open-std.org/JTC1/SC22/WG21/docs/papers/2018/p1204r0.html)** —
+  WG21 paper describing a canonical layout with public headers under a
+  project-named subdirectory.
+
+Both references converge on placing public headers under a dedicated,
+project-named include root.
+
+---
+
+## Options Considered
+
+### Option A: Flat Directory Structure (headers and sources mixed)
+
+All public headers, private headers, and source files live together in the same
+directory (e.g. `libs/my_lib/`). Public vs. private is expressed *only* through the
+Bazel `hdrs`/`srcs` split. This is the current de-facto S-CORE convention.
+
+```
+libs/my_lib/
+├── BUILD.bazel
+├── my_lib.h          # public  -> hdrs
+├── my_lib.cc         # source  -> srcs
+├── internal_helper.h # private -> srcs
+└── internal_helper.cc
+```
+
+#### Advantages
+
+- **Zero Bazel path overhead:** No `strip_include_prefix` / `includes` juggling. The
+  header's on-disk path *is* its include path, so there is one obvious way to write
+  an `#include` and no divergence between what the file tree shows and what the
+  compiler sees.
+- **Easier local navigation:** Declarations sit next to their implementations; no
+  jumping between distant folders in the IDE. Refactoring that splits or renames a
+  unit touches a single directory, and "go to file" / fuzzy-open lands on the pair
+  immediately.
+- **Less nesting:** Avoids redundant paths like `.../my_lib/include/my_lib/`, which
+  keeps both the file tree and relative `#include` statements short.
+- **Lowest friction for small or leaf components:** For a component with one or two
+  public headers, a private helper, and its sources, the flat layout has the best
+  effort-to-value ratio — the `include/` + `src/` split adds structure that such a
+  component does not need. This matches how most of S-CORE is organized today.
+- **Single source of truth for the API boundary:** The public surface is defined in
+  exactly one place — the `hdrs` list in `BUILD.bazel`. There is no second, implicit
+  contract encoded in the directory layout that could drift out of sync with the
+  build file (e.g. a header physically under `include/` but not actually in `hdrs`,
+  or vice versa).
+- **Bazel already enforces the boundary:** A consumer can only `#include` headers a
+  dependency lists in `hdrs`; reaching for a private header in `srcs` fails the
+  build regardless of where the file physically sits. The encapsulation guarantee
+  therefore does not depend on the directory structure at all — the flat layout
+  loses no *enforcement*, only *visibility*.
+- **Cheapest migrations and moves:** Moving a component, extracting a sub-library, or
+  promoting a private header to public is a `BUILD.bazel` edit only — no files move
+  on disk, so diffs stay small and reviews stay focused on the API change itself.
+- **Fewer glob edge cases:** With everything in one directory, `glob` patterns are
+  simple and there is no risk of a misplaced file silently falling outside an
+  `include/**` or `src/**` pattern.
+
+#### Disadvantages
+
+- **Public API not visible from the file tree:** Developers must read `BUILD.bazel`
+  to learn which headers are public.
+- **Accidental API exposure:** A private header may be listed in `hdrs` by mistake,
+  or a consumer may reach for an internal header.
+- **Poor scalability:** As the library grows the folder becomes cluttered.
+- **Packaging difficulty:** Extracting only public headers for external (non-Bazel)
+  delivery requires filtering scripts rather than a folder copy.
+- **Weaker as a published module:** Without a prefix to strip, a consumer sees the
+  producer's repo-relative path, which leaks the internal layout and is not
+  namespaced by component:
+
+  ```cpp
+  // consuming @my_lib as a module, flat layout
+  #include "libs/my_lib/my_lib.h"
+  ```
+
+  Getting the clean `#include "my_lib/my_lib.h"` requires adding `include_prefix`
+  (or naming the flat dir `my_lib/`), i.e. reintroducing the Bazel path machinery
+  Option A set out to avoid.
+
+---
+
+### Option B: Separate `include/` Directory (industry standard)
+
+Public headers are placed in a dedicated `include/` directory, nested under the
+component name (`include/my_lib/*.h`) to prevent include-path collisions. Private
+headers and sources live under `src/`.
+
+```
+libs/my_lib/
+├── BUILD.bazel
+├── include/
+│   └── my_lib/
+│       └── my_lib.h       # public -> hdrs
+└── src/
+    ├── my_lib.cc          # source  -> srcs
+    ├── internal_helper.h  # private -> srcs
+    └── internal_helper.cc
+```
+
+#### Advantages
+
+- **High encapsulation:** The public contract is physically isolated. It is
+  immediately clear to a consumer what may be used.
+- **Clean BUILD files:** `hdrs = glob(["include/**/*.h"])` and
+  `srcs = glob(["src/**"])` without risk of accidentally exposing private files.
+- **Standardized packaging:** Exporting the public API is a single folder copy.
+- **Clear architectural intent:** Naturally enforces good API-design discipline and
+  aligns with the Pitchfork Layout and the Canonical Project Structure.
+- **Collision-free includes:** The `my_lib/` nesting guarantees globally unique
+  include paths, e.g. `#include "my_lib/my_lib.h"`.
+
+#### Disadvantages
+
+- **Bazel path handling:** Requires `strip_include_prefix = "include"` so consumers
+  can write `#include "my_lib/my_lib.h"` rather than `include/my_lib/my_lib.h`.
+- **Folder redundancy:** The `include/my_lib/` nesting feels redundant to some.
+- **Navigation overhead:** Minor cost from switching between `include/` and `src/`
+  during active development.
+
+---
+
+### Option C: Hybrid Directory Structure
+
+Only strictly public API headers are placed in `include/my_lib/*.h`. All private
+headers (`*.h`) and implementation sources (`*.cc`) sit directly in the main
+directory (or a local `src/` next to it).
+
+```
+libs/my_lib/
+├── BUILD.bazel
+├── include/
+│   └── my_lib/
+│       └── my_lib.h       # public -> hdrs
+├── my_lib.cc              # source  -> srcs
+├── internal_helper.h      # private -> srcs
+└── internal_helper.cc
+```
+
+#### Advantages
+
+- **Reduced navigation overhead:** Developers open `include/` only when changing the
+  public contract; internal headers stay next to their `.cc`.
+- **Clear Bazel separation:** `hdrs` matches `include/**`, `srcs` matches the rest.
+- **Clean API packaging:** Copying `include/` still yields a pristine public SDK
+  with no leakage of private helpers.
+
+#### Disadvantages
+
+- **Inconsistent private-header location:** Some headers live in `include/`, some in
+  the root — developers may be unsure where to look.
+- **Strict include conventions required:** Private headers use relative paths
+  (`#include "internal_helper.h"`) while public headers use the nested path
+  (`#include "my_lib/my_lib.h"`), which must be applied consistently.
+
+---
+
+## Evaluation Criteria
+
+| Criterion                         | A: Flat | B: Separate `include/` | C: Hybrid |
+|-----------------------------------|:-------:|:----------------------:|:---------:|
+| Public API visible in file tree   |   --    |          ++            |    +      |
+| API encapsulation / leak safety   |    -    |          ++            |    +      |
+| Clean `#include` paths (no collisions) |  +/-  |          ++            |    +      |
+| Bazel `hdrs`/`srcs` clarity        |    -    |          ++            |    +      |
+| Non-Bazel packaging (folder copy)  |   --    |          ++            |    +      |
+| Bazel boilerplate (low is better)  |   ++    |           -            |   +/-     |
+| Local navigation (edit .h next .cc)|   ++    |           -            |    +      |
+| Scalability of large components    |    -    |          ++            |    +      |
+| Consistency / low cognitive load   |    +    |          ++            |    -      |
+| Alignment with PFL / P1204         |    -    |          ++            |    +      |
+
+---
+
+## Decision Proposal
+
+**Option B: Separate `include/` Directory** — public API headers live under
+`include/<component>/`, private headers and sources live under `src/`.
+
+For components that are small, header-heavy, or where the public/private split is
+trivial, **Option C (Hybrid)** is an accepted variation: keep the `include/<component>/`
+public root, but allow private headers to sit next to their `.cc` in `src/` (or the
+component root). The invariant that must always hold is: **everything a consumer is
+allowed to `#include` lives under `include/<component>/`, and nothing else does.**
+
+### Rationale
+
+1. **The public contract becomes a first-class, visible artifact.** A reviewer or
+   consumer can see the entire API surface by looking at `include/<component>/`
+   without opening `BUILD.bazel`. This is the single biggest weakness of the current
+   flat approach.
+
+2. **It maps directly onto Bazel's `hdrs`/`srcs` model.** The directory boundary and
+   the Bazel visibility boundary coincide, so the file layout reinforces the
+   guarantee Bazel already enforces, instead of leaving it implicit.
+
+3. **Collision-free include paths.** Nesting under the component name
+   (`include/my_lib/`) guarantees that `#include "my_lib/foo.h"` is globally unique
+   across the whole build graph, which matters at S-CORE scale with many modules.
+
+4. **Trivial packaging for non-Bazel consumers.** Delivering the public SDK is a
+   folder-level copy of `include/`, not a filtering script over a mixed directory.
+
+5. **Ecosystem alignment.** The layout matches the Pitchfork Layout (separated
+   header placement) and the WG21 Canonical Project Structure, lowering the barrier
+   for external contributors and tooling.
+
+The main cost — a small amount of Bazel boilerplate (`strip_include_prefix`) and the
+`include/<component>/` nesting — is one-time, mechanical, and easily templated.
+
+---
+
+## Bazel Usage
+
+### Defining a component
+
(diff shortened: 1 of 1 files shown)


anmittag:
TLC 2026-08-20: @qor-lb , @arsibo please review

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+  simple and there is no risk of a misplaced file silently falling outside an
+  `include/**` or `src/**` pattern.
+
+#### Disadvantages
+
+- **Public API not visible from the file tree:** Developers must read `BUILD.bazel`
+  to learn which headers are public.
+- **Accidental API exposure:** A private header may be listed in `hdrs` by mistake,
+  or a consumer may reach for an internal header.
+- **Poor scalability:** As the library grows the folder becomes cluttered.
```
LittleHuba: Disagree on this. You can create subdirectories.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+- **Packaging difficulty:** Extracting only public headers for external (non-Bazel)
+  delivery requires filtering scripts rather than a folder copy.
```
LittleHuba: This will not be solved by doing a pitch-fork. Public headers depend on internal headers in some cases. Especially in a highly templated API as we have it in score_communication.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+├── internal_helper.h # private -> srcs
+└── internal_helper.cc
+```
+
+#### Advantages
+
+- **Zero Bazel path overhead:** No `strip_include_prefix` / `includes` juggling. The
+  header's on-disk path *is* its include path, so there is one obvious way to write
+  an `#include` and no divergence between what the file tree shows and what the
+  compiler sees.
```
LittleHuba: This is actually quite a crucial advantage. Using `strip_include_prefix` messes with external tools. Relying on this option would make packaging tremendously hard since you would have to rewrite source to actually achieve a compilation with any tool that is not aware of bazel.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+- **Easier local navigation:** Declarations sit next to their implementations; no
+  jumping between distant folders in the IDE. Refactoring that splits or renames a
+  unit touches a single directory, and "go to file" / fuzzy-open lands on the pair
+  immediately.
```
LittleHuba: This is not just about navigation. This is also about maintainability. By vicinity of code in the source tree you also know likely affected code.

aschemmel-tech (changes requested):
Option B conflicts with DR-003-proc (as in https://github.com/eclipse-score/score/pull/3194).

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+└── src/
+    ├── my_lib.cc          # source  -> srcs
+    ├── internal_helper.h  # private -> srcs
+    └── internal_helper.cc
+```
+
+#### Advantages
+
+- **High encapsulation:** The public contract is physically isolated. It is
+  immediately clear to a consumer what may be used.
```
LittleHuba: Not really true. E.g. we include private headers in public headers for templated APIs. There we also leak private symbols which could be used by users. This is a false promise.
See https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#src-dir

Also consider that Bazel prohibits you refering to files in other packages. So your Bazel rules would become very awkward if you would not put all headers into `includes` that are referenced in at least one public header.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+- **Clean BUILD files:** `hdrs = glob(["include/**/*.h"])` and
+  `srcs = glob(["src/**"])` without risk of accidentally exposing private files.
```
LittleHuba: How can you raise globbing here as a advantage when you state it in option A as an advantage that it is avoided?

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+- **Standardized packaging:** Exporting the public API is a single folder copy.
```
LittleHuba: Wrong as soon as you must include private headers into public headers.
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#src-dir

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+- **Clear architectural intent:** Naturally enforces good API-design discipline and
+  aligns with the Pitchfork Layout and the Canonical Project Structure.
```
LittleHuba: This sentence is not logical to me. Pitchfork layout and canonical project structure are opposite ends of the spectrum. One stands for spreading headers and sources over different source tree branches while the other explicitly asks to keep them together. Aligning with both at the same time is IMO not possible...

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+- **Collision-free includes:** The `my_lib/` nesting guarantees globally unique
+  include paths, e.g. `#include "my_lib/my_lib.h"`.
```
LittleHuba: I don't see how that is an advantage of pitchforking over canonical project structure.
In both you can have collisions accross bazel modules.

Putting the component name in the path is one way to fix that. But you can do that with either strategy.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+
+#### Disadvantages
+
+- **Bazel path handling:** Requires `strip_include_prefix = "include"` so consumers
+  can write `#include "my_lib/my_lib.h"` rather than `include/my_lib/my_lib.h`.
+- **Folder redundancy:** The `include/my_lib/` nesting feels redundant to some.
+- **Navigation overhead:** Minor cost from switching between `include/` and `src/`
+  during active development.
```
LittleHuba: Minor cost in time but major cost in maintenance.

docs/design_decisions/DR-003-arch.md
LittleHuba: This type of globbing leads to very large bazel targets. That way you loose the caching advantage of Bazel. This will slow down builds tremendously.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+- **Clean API packaging:** Copying `include/` still yields a pristine public SDK
+  with no leakage of private helpers.
+
+#### Disadvantages
+
+- **Inconsistent private-header location:** Some headers live in `include/`, some in
+  the root — developers may be unsure where to look.
+- **Strict include conventions required:** Private headers use relative paths
+  (`#include "internal_helper.h"`) while public headers use the nested path
+  (`#include "my_lib/my_lib.h"`), which must be applied consistently.
```
LittleHuba: Option should be removed since it is absolutely infeasible with templated APIs and how Bazel works.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+
+---
+
+## Evaluation Criteria
+
+| Criterion                         | A: Flat | B: Separate `include/` | C: Hybrid |
+|-----------------------------------|:-------:|:----------------------:|:---------:|
+| Public API visible in file tree   |   --    |          ++            |    +      |
```
LittleHuba: Wrong for templated APIs. No clear differentiation feasible based on files.
See https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#src-dir

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+| API encapsulation / leak safety   |    -    |          ++            |    +      |
```
LittleHuba: Encapsulation / leak safety is enforced by Bazel in either option equally well (see score_communication with the `impl` pattern).

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+| Clean `#include` paths (no collisions) |  +/-  |          ++            |    +      |
```
LittleHuba: This topic is not actually connected. Collisions can be achieved/avoided with all options equally well.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+| Bazel `hdrs`/`srcs` clarity        |    -    |          ++            |    +      |
```
LittleHuba: Wrong. Bazel has more clarity with option A. Also see https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#src-dir

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+├── my_lib.cc              # source  -> srcs
+├── internal_helper.h      # private -> srcs
+└── internal_helper.cc
+```
+
+#### Advantages
+
+- **Reduced navigation overhead:** Developers open `include/` only when changing the
+  public contract; internal headers stay next to their `.cc`.
+- **Clear Bazel separation:** `hdrs` matches `include/**`, `srcs` matches the rest.
```
LittleHuba: Also Bazel packaging forces you to create `cc_library` targets for only headers and only sources which then must be put into a dependency setup via `deps` and `implementation_deps`.
Alternatively you have to work with lots of `export_files` directives.

This makes rule setup overly complex and hard to maintain.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+- **Clean API packaging:** Copying `include/` still yields a pristine public SDK
+  with no leakage of private helpers.
```
LittleHuba: Wrong for templated APIs.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+| Criterion                         | A: Flat | B: Separate `include/` | C: Hybrid |
+|-----------------------------------|:-------:|:----------------------:|:---------:|
+| Public API visible in file tree   |   --    |          ++            |    +      |
+| API encapsulation / leak safety   |    -    |          ++            |    +      |
+| Clean `#include` paths (no collisions) |  +/-  |          ++            |    +      |
+| Bazel `hdrs`/`srcs` clarity        |    -    |          ++            |    +      |
+| Non-Bazel packaging (folder copy)  |   --    |          ++            |    +      |
+| Bazel boilerplate (low is better)  |   ++    |           -            |   +/-     |
+| Local navigation (edit .h next .cc)|   ++    |           -            |    +      |
+| Scalability of large components    |    -    |          ++            |    +      |
```
LittleHuba: I don't know how you get to this.
I also don't see how option B would be so much better here.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+| Consistency / low cognitive load   |    +    |          ++            |    -      |
```
LittleHuba: Disagree. See https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#src-dir

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+| Alignment with PFL / P1204         |    -    |          ++            |    +      |
```
LittleHuba: Sorry, but that is literally saying option B aligns with option B. How should anything compete here?

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+cc_library(
+    name = "my_lib",
+    srcs = glob([
+        "src/**/*.cc",
+        "src/**/*.h",      # private headers are NOT part of the public API
+    ]),
+    hdrs = glob(["include/**/*.h"]),
+    strip_include_prefix = "include",
+    visibility = ["//visibility:public"],
+)
```
LittleHuba, line 334: Sorry, but this really just throws most benefits of Bazel over board. I highly recommend that you go through the basic Bazel documentation and read up on why Bazel recommends small targets.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+
+4. **Trivial packaging for non-Bazel consumers.** Delivering the public SDK is a
+   folder-level copy of `include/`, not a filtering script over a mixed directory.
+
+5. **Ecosystem alignment.** The layout matches the Pitchfork Layout (separated
+   header placement) and the WG21 Canonical Project Structure, lowering the barrier
+   for external contributors and tooling.
+
+The main cost — a small amount of Bazel boilerplate (`strip_include_prefix`) and the
+`include/<component>/` nesting — is one-time, mechanical, and easily templated.
```
LittleHuba: This rationale is wrong in every point:

1. The public contract is not just the headers. And once more, this fully breaks apart for templated APIs.
2. This is fully wrong. See https://bazel.build/tutorials/cpp-use-cases for this. They only show it for **legacy** code and recommend canonical project structure for new projects.
3. Collisions in include paths can be created/avoided with both approaches. Mixing these two topics just adds confusion.
4. Pitchforking actually makes packaging harder, since `strip_include_prefix` will require source level adaptations to actually cope with the rewrites in other build systems.
5. See https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#src-dir
WG21 is clearly not recommending use of the pitch fork layout. So stating that the chosen option matches with both is impossible.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+
+## Evaluation Criteria
+
+| Criterion                         | A: Flat | B: Separate `include/` | C: Hybrid |
+|-----------------------------------|:-------:|:----------------------:|:---------:|
+| Public API visible in file tree   |   --    |          ++            |    +      |
+| API encapsulation / leak safety   |    -    |          ++            |    +      |
+| Clean `#include` paths (no collisions) |  +/-  |          ++            |    +      |
+| Bazel `hdrs`/`srcs` clarity        |    -    |          ++            |    +      |
+| Non-Bazel packaging (folder copy)  |   --    |          ++            |    +      |
```
LittleHuba: Wrong, this is actually harder because of `strip_include_prefix`.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+  #include "libs/my_lib/my_lib.h"
+  ```
+
+  Getting the clean `#include "my_lib/my_lib.h"` requires adding `include_prefix`
+  (or naming the flat dir `my_lib/`), i.e. reintroducing the Bazel path machinery
+  Option A set out to avoid.
+
+---
+
+### Option B: Separate `include/` Directory (industry standard)
```
LittleHuba: Please provide proof for this claim of "industry standard".
Looking at major C++ projects out there, I see both styles.
This feels very biased...

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+`include/<component>/` nesting — is one-time, mechanical, and easily templated.
+
+---
+
+## Bazel Usage
+
+### Defining a component
+
+Public headers go into `hdrs` and are exposed through `strip_include_prefix` so that
+the on-disk `include/` prefix is removed from the include path consumers see:
```
LittleHuba: Doing this, completely breaks other tools because headers are not where source files expect them.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+consumer's source from the producer's on-disk layout and is the key property that
+makes the module boundary clean.
+
+### Consumer view across the three layouts
+
+How consuming `@my_lib` as a module looks from the outside, for each option:
+
+| Option | Producer `BUILD.bazel` (public headers) | Consumer `#include` |
+|--------|------------------------------------------|---------------------|
+| **A: Flat** | `hdrs = glob(["*.h"])` — no prefix handling | `#include "libs/my_lib/my_lib.h"` — leaks repo layout, not namespaced |
```
LittleHuba: Where does the `libs` come from? We do not have this today.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+  requires headers to be included by their **repo-root-relative path**
+  (`#include "score/mw/foo/bar.h"`), with no dedicated `include/` directory. The
+  [Bazel `cc_library` reference](https://bazel.build/reference/be/c-cpp#cc_library)
+  encodes the same public/private model via `hdrs`/`srcs`. This is essentially
+  **Option A**, but with repo-root paths rather than bare filenames.
+- **General open-source convention (dedicated `include/`):** The
+  [Pitchfork Layout (PFL)](https://github.com/vector-of-bool/pitchfork) and the
+  [WG21 Canonical Project Structure (P1204R0)](https://open-std.org/JTC1/SC22/WG21/docs/papers/2018/p1204r0.html)
+  both place public headers under a project-named `include/<lib>/` root — matching
+  **Options B/C**.
```
LittleHuba: This is just wrong.

1. There is no general open-source convention. There are many project examples for both.
2. Canonical Project Structure explicitly mentions pitch forking as a bad practice that should not be followed - see https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#src-dir

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+The **Bazel module name does not protect against this**: `@module_a` / `@module_b` do
+not appear in the C++ include path by default; the path is determined solely by the
+package location and by `strip_include_prefix` / `include_prefix` / `includes`.
+
+Consequences per option:
+
+- **Option A (flat, repo-root includes):** collision-safe *as long as* headers are
+  included by their full repo-root path. The danger is resetting the prefix
+  (`strip_include_prefix = "."` as in `utils/base64`, or `includes = ["."]`), which
+  collapses the path to `#include "error.h"` and re-introduces the ambiguity globally.
```
LittleHuba: This is why we are strict about full include paths in our source in score_communication.
The same logic applies for option B. It is just enforced because without providing the full path you simply don't find the header. The result in your source files is the same.

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,534 @@
...
+
+### Follow-Up Actions
+
+- Provide a component template / scaffolding (directory skeleton + `BUILD.bazel`) that
+  encodes the `include/<component>/` + `src/` layout and `strip_include_prefix`.
+- Document the convention in the S-CORE contribution guidelines and C++ coding
+  guidelines, including the private-vs-public include-path rules.
+- Define a migration path for existing flat components (opportunistic, per module).
+- Consider a lightweight CI/lint check that flags private headers appearing in `hdrs`
+  or public headers being included via non-canonical paths.
```
LittleHuba: As stated previously, this discussion is not related to the main topic of this DR. It should be handled separately to avoid confusion.

LittleHuba (changes requested):
I find this proposal heavily skewed towards a pitchfork layout (option B) (even clearly stating pitchfork layout in the evaluation criteria. Based on that, I find it hard to base a discussion on this, since it feels like the arguments were written based on the decision instead of the other way round.

From my comments you can probably guess that I'm clearly leaning towards option A. And this review felt like a uphill battle from the start. Looking just at the advantages/disadvantages of options A and B, the evaluation should not be leaning so heavily towards option B.

Two sources for reference why I heavily prefer option A:

1. https://bazel.build/tutorials/cpp-use-cases where the pitch fork layout is clearly only stated as a legacy adoption use case. This means that Bazel is not built for this project structure and that you will feel lots of friction trying to use it.
2. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#src-dir has a full section just talking about why pitch fork layout is bad and should not be used

Side note: I appreciate the accepted alternative that will safe communication and baselibs from having to rewrite 90% of their code. I still do not feel comfortable agreeing with the final decision because of the consequences for the overall project.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Agreed and changed, A flat package can be organized with subdirectories and per-package targets, so "clutter" was overstated. This disadvantage has been removed; Option A is now the recommended layout and its write-up no longer claims poor scalability.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Agreed. It is now called out as an advantage and in the consumer-view table ("works with non-Bazel tools: yes").

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Agreed. Option A now lists the maintainability/vicinity benefit ("the source tree makes likely-affected code visible by vicinity — a maintainability property, not just a navigation convenience"), and it's a rationale point.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Also Agreed. This is now treated as the decisive argument. Option A solves it with impl/ + restricted visibility, which holds even for templated public headers.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Sorry that was contradictory. Both glob-related claims were removed.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Agreed — this is now explicitly documented as an Option B/C disadvantage and an Option A advantage (include path == on-disk location), including in the consumer-view table.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Fixed — all examples now use real S-CORE paths (score/mw/my_component/…) instead of the invented libs/my_lib/.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: The "general open-source convention (dedicated include/)" claim is removed; the guidance now notes both styles exist.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Agreed. Option A adopts exactly that rule — full project-prefixed repo-root include paths — and the DR notes that the resulting #include in source is effectively the same discipline communication already enforces, without needing strip_include_prefix.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Agreed. Globbing leads to a problem, because any edit to any file invalidates the whole target and forces every reverse dependency to rebuild. So thats not a good idea.

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: Agreed. Header-name collisions are orthogonal to the public-header layout decision — they're governed by the include-path prefix, which every option can get right or wrong independently. Keeping a full analysis here blurs the actual decision. It's not part of the discussion of the option discussion anymore. But an general discussion. Shall I remove that from the PR or move it somewhere else ?

LittleHuba: IMO we need a separate DR for the topic of include path collision and symbol collision. AFAIK there were already some initial talks in some meetings in S-CORE.

aschemmel-tech (dismissed):
Thanks for the rework, option A selected matches with DR-003-proc

docs/design_decisions/DR-003-arch.md
RolandJentschETAS: I have remove that chapter. Can be introduced like recommended into an own design decision.
@LittleHuba : Can you please check now again ?

docs/design_decisions/DR-003-arch.md
```diff
@@ -0,0 +1,522 @@
...
+- No migration is required for the existing `communication`/`baselibs` code, which
+  already follows this layout; and the decision aligns with the Canonical Project
+  Structure and Bazel guidance.
+
+### Negative / Costs
+
+- The public surface is discovered via the `impl/` convention and Bazel visibility
+  rather than a single `include/` folder, so the convention must be applied
+  consistently.
+- Contributors must consistently use full project-prefixed include paths and must not
```
FScholPer, line 511: We should also encounter how we check new sources and monitor the structure

LittleHuba: There was some work in clang-tidy that went in this direction but was never finished:
https://github.com/llvm/llvm-project/issues/63226

docs/design_decisions/DR-003-arch.md
```diff
@@ -45,358 +45,381 @@ than working against it.
...
 
-### Option A: Flat Directory Structure (headers and sources mixed)
+### Option A: Bazel-native flat layout (headers next to sources) — recommended
 
-All public headers, private headers, and source files live together in the same
-directory (e.g. `libs/my_lib/`). Public vs. private is expressed *only* through the
-Bazel `hdrs`/`srcs` split. This is the current de-facto S-CORE convention.
+Headers and sources live **next to each other** in the component's package, and the
+public/private boundary is drawn by Bazel: public headers go in `hdrs`, everything
+non-public is isolated in an `impl/` subpackage whose `visibility` is restricted to
```
LittleHuba, line 93: Not quite sure how prescribing this impl package makes things better/worse.
While we have this pattern in mw::com, it is honestly a lie to some extent (see https://github.com/eclipse-score/communication/blob/main/score/mw/com/types.h).

For mw::com this is something we will probably never change, but other libraries might not want that pattern.

Before we make users deduce public API from directory structures we should actually provide clear documentation with the public interface. In communication we work with breathe, Sphinx and these tags to mark public API: https://github.com/eclipse-score/communication/blob/main/score/mw/com/types.h#L37

TL;DR: I vote for removing the impl package guideline in favor of describing the public API via tagging and a generated documentation, while leaving projects to come up with a good source structure based on their architecture.

RolandJentschETAS: @LittleHuba : I can change it in a way, that this is optional to isolate the private headers/sources in an impl folder ? Would that be better ? Or complete remove it ?

qor-lb:
As discussed in [Weekly 2026-09-21](https://github.com/orgs/eclipse-score/discussions/2457#discussioncomment-18536335) this was onboarded to the FEP process. #3269 tracks also the implementation once the DR is agreed.

@LittleHuba let me know once it is ready for the FCP and then I will do the announcement and facilitate the alignment.

anmittag (approved):

LittleHuba (approved):

LittleHuba:
@qor-lb this PR is ready now for the FCP from my side.

PR #3211, opened 2026-08-24
Add software development plan elements section

anmittag:
Added a section outlining software development plan elements.

### DIFF
--- a/docs/platform_management_plan/software_development.rst
+++ b/docs/platform_management_plan/software_development.rst
@@ -209,3 +209,13 @@ The qualified `Ferrocene <https://github.com/ferrocene>`__ compiler is planned t
 
 The main build environment of the project is based on `Bazel <https://bazel.build>`__. It it used to build software
 components, documentation, and automated tests.
+
+Software Development Plan Elements
+----------------------------------
+
+.. needtable::
+   :style: table
+   :columns: title;id;safety;security;status
+   :colwidths: 25,45,10,10,10
+   :sort: docname
+   :filter: "wp__sw_development_plan" in realizes


anmittag:
changes are visible here https://eclipse-score.github.io/score/pr-3211/platform_management_plan/software_development.html

docs/platform_management_plan/software_development.rst
```diff
@@ -209,3 +209,13 @@ The qualified `Ferrocene <https://github.com/ferrocene>`__ compiler is planned t
 
 The main build environment of the project is based on `Bazel <https://bazel.build>`__. It it used to build software
 components, documentation, and automated tests.
+
+Software Development Plan Elements
+----------------------------------
```
PandaeDo, line 214: I would prefer to add the additional coding guidelines to the chapter with the coding guidelines. Same for additional references. I think this would make it more readable than a list at the end of the document with additional information. Appreciate the idea of making it complete!

PandaeDo (changes requested):

docs/platform_management_plan/software_development.rst
anmittag: @PandaeDo : Adding additional Chapters is fine, but how to ensure all Elements relevant for the Software Development plan is "added". Then I suggest to add the information as you requested AND having a Summary of all relevant SWDevPlan Elements at the end. Agree? Anyhow coding guideleines are already added Line 107ff.

PandaeDo: I think that's the main question. It's a responsibility of the maintenance of the plan(s). I think that's an improvement to check against them. So we'll get informed if there are changes. The list will be always updated. Even when you define a document that was defined to realise the plan. So I would agree on update the plan according to additional aspects and add the list at the end.

anmittag:
part of #3224 => can be closed

